PR #28, opened 2024-08-06
Support 1-d array data in profile exporter

nv-hwoo:
1. Support 1-d array data in profile exporter so that when the input/output involves more than one data value (or array of values), then it captures all the data, rather than just the first element. This is an incremental step towards supporting full N-dimensional tensor in profile exporter. Currently, we don't need to support N-dimensional tensors because we are mostly interested in the text string or token ids (which is just a 1-d array). Additionally, N-dimensional tensor support requires more changes to PA since response outputs doesn't store shape information of the tensors returned from the model so we need to retrieve that information somewhere.
2. Support other data types such as uint, fp, and etc.

Example:

When given trtllm engine the following input:
```
{
    "token_ids": [1, 2, 3, 4, 5, 6, 7, 8],
    "input_length": [8],
    "request_output_len": [10],
    ...
}
```
what we got from PA _before_ the change is
```
{
    "request_inputs": {
        "token_ids": 1,  // missing remaining values
        "input_length": 8,
        "request_output_len": 10,
    },
    ...
}
```
_After_ the fix, we get the full array of token ids (same for any other fields in the profile export json file):
```
{
    "request_inputs": {
        "token_ids": [1, 2, 3, 4, 5, 6, 7, 8],
        "input_length": 8,
        "request_output_len": 10,
    },
    ...
}
```

### DIFF
--- a/genai-perf/genai_perf/parser.py
+++ b/genai-perf/genai_perf/parser.py
@@ -582,25 +582,6 @@ def _add_endpoint_args(parser):
     )
 
     endpoint_group.add_argument(
-        "-m",
-        "--model",
-        nargs="+",
-        default=[],
-        help=f"The name of the model(s) to benchmark.",
-    )
-    endpoint_group.add_argument(
-        "--model-selection-strategy",
-        type=str,
-        choices=utils.get_enum_names(ModelSelectionStrategy),
-        default="round_robin",
-        required=False,
-        help=f"When multiple model are specified, this is how a specific model "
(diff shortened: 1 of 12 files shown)


matthewkotila (approved):

genai-perf/genai_perf/parser.py
matthewkotila, line 1: @nvda-mesharma how come the CI triggering is not part of the checks for this PR?

This PR is based on top of your merged #7 PR which adds CI triggering.

PR #30, opened 2024-08-06
Fix handling of unfinished responses from PA

debermudez:
Previously, we were seeing SSE event responses coming back from the model that were not complete.
Complete responses would span multiple responses.
Parsing these responses would fail.
Now we check the responses and merge when necessary to enable correct parsing.

- Added fix
- Add unit test with an example offending response set.

Example failing response
```
{
                                "response": 'data: {"id":"8ae","object":"chat.completion.chunk","created":172,"choices":[{"index":0,"text":" seems","logprobs":null,"finish_reason":null,"seed":null,"delta":{"token_id":5084,"role":"assistant","content":" seems","tool_calls":null}}],"model":"meta-llama/Llama-3-8b-chat-hf","usage":null}\n\ndata: {"id":"8ae835f2ecbb67f3-SJC","object":"chat.completion.chunk","created":1722875835,"choices":[{"index":0,"text":" like","logprobs":null,"finish_reason":null,"seed":null,"delta":{"token_id":1093,"role":"assistant","content":" like","tool_calls":null}}],"model":"meta-llama/Llama-3-8b-chat-hf","usage":null}\n\ndata: {"id":"8ae835f2ecbb67f3-SJC","object":"chat.completion.chunk","created":1722875835,"choices":[{"index":0,"text":" you","logprobs":null,"finish_reason":null,"seed":null,"delta":{"token_id":499,"role":"assistant","content":" you","tool_calls":null}}],"model":"meta-llama/Llama-3-8b-chat-hf","usage":null}\n\ndata: {"id":"8ae835f2ecbb67f3-SJC","object":"chat.completion.chunk","created":1722875835,"choices":[{"index":0,"text":"\'re","logprobs":null,"finish_reason":null,"seed":null,"delta":{"token_id":2351,"role":"assistant","content":"\'re","tool_calls":null}}],"model":"meta-llama/Llama-3-8b-chat-hf","usage":null}\n\ndata: {"id":"8ae835f2ecbb67f3-SJC","object":"chat.completion.chunk","created":1722875835,"choices":[{"index":0,"text"'
                            },
                            {
                                "response": ':" writing","logprobs":null,"finish_reason":null,"seed":null,"delta":{"token_id":4477,"role":"assistant","content":" writing","tool_calls":null}}],"model":"meta-llama/Llama-3-8b-chat-hf","usage":null}\n\ndata: {"id":"8ae835f2ecbb67f3-SJC","object":"chat.completion.chunk","created":1722875835,"choices":[{"index":0,"text":" a","logprobs":null,"finish_reason":null,"seed":null,"delta":{"token_id":264,"role":"assistant","content":" a","tool_calls":null}}],"model":"meta-llama/Llama-3-8b-chat-hf","usage":null}\n\ndata: {"id":"8ae835f2ecbb67f3-SJC","object":"chat.completion.chunk","created":1722875835,"choices":[{"index":0,"text":" letter","logprobs":null,"finish_reason":null,"seed":null,"delta":{"token_id":6661,"role":"assistant","content":" letter","tool_calls":null}}],"model":"meta-llama/Llama-3-8b-chat-hf","usage":null}\n\ndata: {"id":"8ae835f2ecbb67f3-SJC","object":"chat.completion.chunk","created":1722875835,"choices":[{"index":0,"text":" or","logprobs":null,"finish_reason":null,"seed":null,"delta":{"token_id":477,"role":"assistant","content":" or","tool_calls":null}}],"model":"meta-llama/Llama-3-8b-chat-hf","usage":null}\n\ndata: {"id":"8ae835f2ecbb67f3-SJC","object":"chat.completion.chunk","created":1722875835,"choices":[{"index":0,"text":" a","logprobs":null,"finish_reason":null,"seed":null,"delta":{"token_id":264,"role":"assistant","content":" a","tool_'
                            },
                            {
                                "response": 'calls":null}}],"model":"meta-llama/Llama-3-8b-chat-hf","usage":null}\n\ndata: {"id":"8ae835f2ecbb67f3-SJC","object":"chat.completion.chunk","created":1722875835,"choices":[{"index":0,"text":" passage","logprobs":null,"finish_reason":null,"seed":null,"delta":{"token_id":21765,"role":"assistant","content":" passage","tool_calls":null}}],"model":"meta-llama/Llama-3-8b-chat-hf","usage":null}\n\ndata: {"id":"8ae835f2ecbb67f3-SJC","object":"chat.completion.chunk","created":1722875835,"choices":[{"index":0,"text":" with","logprobs":null,"finish_reason":null,"seed":null,"delta":{"token_id":449,"role":"assistant","content":" with","tool_calls":null}}],"model":"meta-llama/Llama-3-8b-chat-hf","usage":null}\n\n'
                            },

```

Closes https://github.com/triton-inference-server/perf_analyzer/pull/27

### DIFF
--- a/genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
+++ b/genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
@@ -174,6 +174,14 @@ class LLMProfileDataParser(ProfileDataParser):
         if self._service_kind == "openai":
             # PA sometimes receives multiple SSE responses at once (as a single
             # response). Handle these responses by merging into a single response.
+            for i in range(len(res_outputs) - 1, -1, -1):
+                response = res_outputs[i]["response"]
+                if not response.startswith("data: "):
+                    first_data = response.find("data: ")
+                    res_outputs[i - 1]["response"] = (
+                        res_outputs[i - 1]["response"] + response[0:first_data].strip()
+                    )
+                    res_outputs[i]["response"] = response[first_data:].strip()
             for i in range(len(res_outputs)):
                 response = res_outputs[i]["response"]
                 responses = response.strip().split("\n\n")
@@ -193,7 +201,9 @@ class LLMProfileDataParser(ProfileDataParser):
             # Remove responses without any content
             indices_to_remove = []
             for idx, out in enumerate(res_outputs):
-                if self._is_openai_empty_response(out["response"]):
+                if not out["response"] or self._is_openai_empty_response(
+                    out["response"]
+                ):
                     indices_to_remove.append(idx)
             indices_to_remove.sort(reverse=True)
             for index in indices_to_remove:
@@ -268,7 +278,9 @@ class LLMProfileDataParser(ProfileDataParser):
         """Extracts text/content of the OpenAI response object."""
         response = remove_sse_prefix(response)
 
-        if response == "[DONE]":
+        if response == "[DONE]" or not (
+            response.strip().startswith("{") and response.strip().endswith("}")
+        ):
             return ""
 
         data = load_json_str(response)
--- a/genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
+++ b/genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
@@ -172,16 +172,25 @@ class LLMProfileDataParser(ProfileDataParser):
     ) -> None:
         """Helper function to preprocess responses of a request."""
         if self._service_kind == "openai":
+            # Sometimes streamed chunks are returned in a splintered fashion.
+            # This forces a merge with the previous chunk if error detected.
+            if len(res_outputs) > 1:
+                for i in reversed(range(len(res_outputs))):
+                    response = res_outputs[i]["response"]
+                    if not response.startswith("data: "):
+                        first_data = response.find("data: ")
+                        if first_data == -1:
+                            res_outputs[i - 1]["response"] = (
+                                res_outputs[i - 1]["response"] + response.strip()
+                            )
+                            res_outputs[i]["response"] = ""
+                        else:                        
+                            res_outputs[i - 1]["response"] = (
+                                res_outputs[i - 1]["response"] + response[0:first_data].strip()
+                            )
+                            res_outputs[i]["response"] = response[first_data:].strip()
             # PA sometimes receives multiple SSE responses at once (as a single
             # response). Handle these responses by merging into a single response.
-            for i in range(len(res_outputs) - 1, -1, -1):
-                response = res_outputs[i]["response"]
-                if not response.startswith("data: "):
-                    first_data = response.find("data: ")
-                    res_outputs[i - 1]["response"] = (
-                        res_outputs[i - 1]["response"] + response[0:first_data].strip()
-                    )
-                    res_outputs[i]["response"] = response[first_data:].strip()
             for i in range(len(res_outputs)):
                 response = res_outputs[i]["response"]
                 responses = response.strip().split("\n\n")
@@ -278,9 +287,7 @@ class LLMProfileDataParser(ProfileDataParser):
         """Extracts text/content of the OpenAI response object."""
         response = remove_sse_prefix(response)
 
-        if response == "[DONE]" or not (
-            response.strip().startswith("{") and response.strip().endswith("}")
-        ):
+        if response == "[DONE]":
             return ""
 
         data = load_json_str(response)
--- a/genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
+++ b/genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
@@ -184,9 +184,10 @@ class LLMProfileDataParser(ProfileDataParser):
                                 res_outputs[i - 1]["response"] + response.strip()
                             )
                             res_outputs[i]["response"] = ""
-                        else:                        
+                        else:
                             res_outputs[i - 1]["response"] = (
-                                res_outputs[i - 1]["response"] + response[0:first_data].strip()
+                                res_outputs[i - 1]["response"]
+                                + response[0:first_data].strip()
                             )
                             res_outputs[i]["response"] = response[first_data:].strip()
             # PA sometimes receives multiple SSE responses at once (as a single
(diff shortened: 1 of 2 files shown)


genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
matthewkotila, line 1: It looks like this PR builds on #27--can we close #27?

debermudez: we will.
i actually think there is a way to automate that with a comment here. I just dont know the exact syntax.

matthewkotila: You can add this to this PR's description:

> Closes #27

https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

debermudez: you are the best dude. 
thank you

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
```diff
@@ -174,6 +174,14 @@ def _preprocess_response(
...
             # PA sometimes receives multiple SSE responses at once (as a single
             # response). Handle these responses by merging into a single response.
+            for i in range(len(res_outputs) - 1, -1, -1):
+                response = res_outputs[i]["response"]
+                if not response.startswith("data: "):
+                    first_data = response.find("data: ")
+                    res_outputs[i - 1]["response"] = (
+                        res_outputs[i - 1]["response"] + response[0:first_data].strip()
+                    )
+                    res_outputs[i]["response"] = response[first_data:].strip()
```
nv-hwoo, line 184: let me know if I understood the code correctly:
1. check if the i-th response starts with `data: ` prefix
2. if not, then find the prefix _within_ the i-th response
3. concatenate the (i-1)-th response and i-th response (up to the prefix)
4. set the i-th response to the remaining content.

then my question is: what happens if `data: ` prefix does _not_ exist in i-th response?

for example, if the res output looked like below (simplified)
```python
res_output = [ 'data: {"x": 10', '"y": 20}' ]
```

cc @IzzyPutterman

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
```diff
@@ -268,7 +278,9 @@ def _extract_openai_text_output(self, response: str) -> str:
         """Extracts text/content of the OpenAI response object."""
         response = remove_sse_prefix(response)
 
-        if response == "[DONE]":
+        if response == "[DONE]" or not (
+            response.strip().startswith("{") and response.strip().endswith("}")
```
nv-hwoo, line 282: I don't know how I feel about this line. This is basically just ignoring any responses that doesn't start or end with proper json format, but given that we already have a lot of custom checks in `_preprocess_response` method, I would rather let the program complain instead of (potentially) silently ignoring it when something's wrong with the response.

IzzyPutterman: I had this occur when running against TGI, where everything was formatted correctly excpet for the last chunk with was 1-2 random characters.  I think probably some garbage due to the fact that TGI doesnt send a DONE, (which is now properly killed in GenAI Perf, but seemed to be unhappy)

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
IzzyPutterman: If it does not exist, then that chunk will be empty and it will be taken out in line 204 below.

nv-hwoo: If it doesn't exist, it would still go into the if block right? So then using my example, 
1. it will check if `'"y": 20}'` starts with the prefix, and since it doesn't it will go into the block
2. check if prefix is within the string and again it doesn't so `first_data` is going to be `-1`
3. then `res_output[0] = res_output[0] + res_output[1][0:-1]` which is concatenation of `'data: {"x": 10,'` and `'"y": 20'`, or `'data: {"x": 10,"y": 20'`, which is missing that last character `}` (because slicing is from `0` to `-1`).

Am I understanding this correctly?

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
nv-hwoo: Ah I see so this is to workaround some issues with TGI. In that case, I think it would be better if this PR only handled the fix for sse related error, and maybe create like a follow up ticket to investigate/handle this as a separate work. @debermudez what do you think?

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
IzzyPutterman: Must be an edge case which isnt hit somehow.  Perhaps responses that are split have always containerd data: in them.

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
```diff
@@ -174,6 +174,14 @@ def _preprocess_response(
         if self._service_kind == "openai":
             # PA sometimes receives multiple SSE responses at once (as a single
             # response). Handle these responses by merging into a single response.
+            for i in range(len(res_outputs) - 1, -1, -1):
```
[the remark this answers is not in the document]
IzzyPutterman: Maybe also wrap in an `if len(res_outputs) > 1`

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
```diff
@@ -174,6 +174,14 @@ def _preprocess_response(
         if self._service_kind == "openai":
             # PA sometimes receives multiple SSE responses at once (as a single
             # response). Handle these responses by merging into a single response.
```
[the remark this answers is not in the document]
debermudez: I think this got updated. Resolving.

genai-perf/tests/test_llm_profile_data_parser.py
```diff
@@ -512,6 +515,16 @@ def test_empty_response(self, mock_read_write: pytest.MonkeyPatch) -> None:
             tokenizer=tokenizer,
         )
 
+    def test_unfinished_responses(self, mock_read_write: pytest.MonkeyPatch) -> None:
```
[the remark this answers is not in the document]
debermudez: Updated

nv-hwoo (approved):
LGTM 👍

PR #31, opened 2024-08-07
Codeql updates

nvda-mesharma:


dyastremsky (approved):
Great work! Thanks for getting this done so quickly.

dyastremsky:
@fpetrini15 @nvda-mesharma @ganeshku1 

Do we just want to run CodeQL on Python? That's what we have historically done. With the PA repo being on its own, it could be a good opportunity to run it on C++ but I notice the time to run CodeQL is now 48 minutes, which is a long runtime for someone to get feedback on their code after pushing a commit.

Also, it took a few minutes of scrolling down the build commands to find the build failure (not the best UI for searching the logs for build failures), but it looks like it failed due to needing rapidjson. If we do add the C++ build, it might require a couple of fixes to get it building.

PR #32, opened 2024-08-07
Goodput initial implementation

AndyDai-nv:
Migrate goodput dev branch to this new repo.


genai-perf/docs/goodput_tutorial.md
```diff
@@ -0,0 +1,102 @@
...
+<details>
+<summary>See instructions</summary>
+
+Run Triton Inference Server with vLLM backend container:
+
+```bash
+export RELEASE="yy.mm" # e.g. export RELEASE="24.06"
+
+
+docker run -it --net=host --gpus=1 --shm-size=2g --ulimit memlock=-1 --ulimit stack=67108864 nvcr.io/nvidia/tritonserver:${RELEASE}-vllm-python-py3
```
dyastremsky, line 42: Could we use vLLM or another approach that does not require several minutes to run? See the tutorials we have that use the OpenAI container as an example.

genai-perf/genai_perf/metrics/llm_metrics.py
```diff
@@ -105,4 +109,6 @@ def system_metrics(self) -> List[MetricMetadata]:
         # base metrics first and then task specific metrics. Uncomment the below
         # line to enable this order:
         # return base_metrics + self.LLM_SYSTEM_METRICS
+        # Rightnow the goodput will be printed out before throughput if there is
```
dyastremsky, line 112: This comment seems like it could be removed.

genai-perf/genai_perf/metrics/statistics.py
```diff
@@ -131,6 +131,8 @@ def _add_units(self, key) -> None:
             self._stats_dict[key]["unit"] = "ms"
         elif key == "request_throughput":
             self._stats_dict[key]["unit"] = "requests/sec"
+        elif key == "request_goodput":
```
dyastremsky, line 134: You can combine this with the previous if condition for conciseness.

genai-perf/genai_perf/parser.py
```diff
@@ -244,6 +244,24 @@ def _check_load_manager_args(args: argparse.Namespace) -> argparse.Namespace:
         args.concurrency = 1
     return args
 
+def _check_goodput_args(args):
```
dyastremsky, line 247: Please look at how extra inputs works and see if you can use the same code (or even refactor out a function). Goodput constraints should be turned into a dictionary where the users can put in any value of --goodput-constraint <key:value> --goodput-constraint <key:value> and so on, where order does not matter. Those keys should match any statistic available for that model. The value should be able to be a float or an int. This is the biggest change I'd like to see.

If the model has a statistic available that matches that key, then it should be used. If not, then goodput can be set to N/A and we can print out a message stating that goodput could not be calculated because <key> does not exist as a statistic for the model.

genai-perf/tests/test_llm_metrics.py
```diff
@@ -71,11 +71,13 @@ def test_llm_metric_system_metrics(self) -> None:
         )
 
         sys_metrics = m.system_metrics
-        assert len(sys_metrics) == 2
+        assert len(sys_metrics) == 3
         assert sys_metrics[0].name == "output_token_throughput"
         assert sys_metrics[0].unit == "per sec"
-        assert sys_metrics[1].name == "request_throughput"
+        assert sys_metrics[1].name == "request_goodput"
```
dyastremsky, line 77: You want to write tests specifically for goodput. My recommendation on the unit testing side is at least these three:
1. What happens when 1 goodput constraint is added?
2. What happens when 3 goodput constraints are added?
3. What happens when 2 goodput constraints are added and one does not exist in the model?

dyastremsky (commented):
Great start, Andy! Left a few comments. If you have any questions, let me know.

genai-perf/tests/test_llm_metrics.py
AndyDai-nv: Thanks for the comment. This is not a start for unit testing work.
Unit tests work for goodput is another todo ticket.
This one is just one of previous unit tests, if I don't make these changes, it does not pass locally.

dyastremsky (commented):
Excellent work, Andy! This POC looks good. I added some small comments as I reviewed to update once the core development is done.

What are the next steps? We can also chat offline. I think it'd be great to get unit testing up for this.

Once you finish doing this work for LLMs, we can look at creating a class for non-LLMs (specifically embeddings/rankings). With the way you coded this, I don't see anything special about the llm_goodput_reporter that couldn't be used for non-LLM models. You may want to try using it with embeddings/rankings and seeing if it works.

genai-perf/genai_perf/goodput_reporter/llm_goodput_reporter.py
```diff
@@ -0,0 +1,86 @@
...
+        target_metric_values = list(self._valid_slos.values())
+        requests_metric_values = self._combined_requests_metric_values
+        good_req_count = 0
+
+        for request_metric_values in requests_metric_values:
+            if all(val < slo 
+                   for val, slo in zip(request_metric_values, target_metric_values)
+            ):
+                good_req_count += 1
+        self._good_req_count = good_req_count
```
nv-hwoo, line 81: I think we should avoid setting return values that will be used later as an attribute _within_ the method. This is very confusing from the reader's perspective because the "count" is hiding inside the method, and it's being used under `compute_goodput`. 

Let's return the counts 
```python
def count_good_reqs(self) -> int:
    ...
    return good_req_count
```

and same goes for other methods as well.

genai-perf/genai_perf/goodput_reporter/llm_goodput_reporter.py
```diff
@@ -0,0 +1,86 @@
...
+    def count_good_reqs(self) -> None:
+        """Count the number of good requests according to SLOs."""
+        target_metric_values = list(self._valid_slos.values())
+        requests_metric_values = self._combined_requests_metric_values
+        good_req_count = 0
+
+        for request_metric_values in requests_metric_values:
+            if all(val < slo 
+                   for val, slo in zip(request_metric_values, target_metric_values)
+            ):
```
nv-hwoo, line 79: Could we unroll this? It hurts readability and I don't think there's much gain in fitting this all into the if conditional block.

genai-perf/genai_perf/goodput_reporter/llm_goodput_reporter.py
```diff
@@ -0,0 +1,86 @@
...
+        invalid_slos = []
+        self._valid_slos = {}
+        valid_names = [metric.name for metric in self._metric.request_metrics]
+
+        for slo_name, slo_value in self._goodput_constraints.items(): 
+            if self._metric.get_base_name(slo_name) not in valid_names:
+                invalid_slos.append(slo_name)
+            else:
+                self._valid_slos[slo_name] = slo_value * self.MS_TO_NS_CONVERSION
+        if invalid_slos:
```
nv-hwoo, line 58:
```suggestion
                self._valid_slos[slo_name] = slo_value * self.MS_TO_NS_CONVERSION
                
        if invalid_slos:
```

genai-perf/docs/goodput_tutorial.md
```diff
@@ -0,0 +1,103 @@
...
+┃                Statistic ┃    avg ┃    min ┃    max ┃    p99 ┃    p90 ┃    p75 ┃
+┡━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━┩
+│ Time to first token (ms) │   5.60 │   4.00 │  13.03 │  12.69 │   7.02 │   5.04 │
+│ Inter token latency (ms) │   1.63 │   1.38 │   2.28 │   2.23 │   1.82 │   1.63 │
+│     Request latency (ms) │ 184.84 │ 173.10 │ 237.81 │ 237.75 │ 228.81 │ 180.72 │
+│   Output sequence length │ 111.43 │ 103.00 │ 126.00 │ 124.55 │ 115.10 │ 114.00 │
+│    Input sequence length │ 200.00 │ 200.00 │ 200.00 │ 200.00 │ 200.00 │ 200.00 │
+└──────────────────────────┴────────┴────────┴────────┴────────┴────────┴────────┘
+Output token throughput (per sec): 1205.26
+Request goodput (per sec): 7.21
```
debermudez, line 100: We are outputting the raw value here.
Do we also want to include a percentage? To give the user of a quick sense of the quality of the run?

genai-perf/genai_perf/goodput_reporter/goodput_reporter.py
```diff
@@ -0,0 +1,91 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+from abc import ABC, abstractmethod
+from typing import Dict, List
+
+from genai_perf.metrics import Metrics
+
+class GoodputReporter(ABC):
```
debermudez, line 35: When I see reporter, I think it is outputting data to a screen or file.
Perhaps GoodputCalculator?

genai-perf/genai_perf/goodput_reporter/goodput_reporter.py
```diff
@@ -0,0 +1,91 @@
...
+        self.set_valid_slos()
+        self.combine_requests_metric_values()
+        self.count_good_reqs()
+        self.compute_goodput()
+    
+    @abstractmethod
+    def set_valid_slos(self) -> None:
+        """Check user's Service Level Objectives (SLOs) inputs. 
+        Set the valid ones while logging the invalid ones. 
+        To be implemented by subclasses.
```
debermudez, line 63: This is known from the decorator.

genai-perf/genai_perf/goodput_reporter/goodput_reporter.py
```diff
@@ -0,0 +1,91 @@
...
+        """Combine metric values at per request level.
+        Only the metrics from valid SLOs.  
+        To be implemented by subclasses.
+        """
+        pass
+    
+    @abstractmethod
+    def count_good_reqs(self) -> None:
+        """Count the number of good requests according to SLOs. 
+        To be implemented by subclasses.
```
debermudez, line 78: same as above

genai-perf/genai_perf/goodput_reporter/goodput_reporter.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        pass
+    
+    @abstractmethod
+    def compute_goodput(self) -> None:
+        """Compute the goodput. To be implemented by subclasses."""
```
debermudez, line 84: same as above

genai-perf/genai_perf/goodput_reporter/llm_goodput_reporter.py
```diff
@@ -0,0 +1,86 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from typing import Dict
+from genai_perf.goodput_reporter.goodput_reporter import GoodputReporter
+from genai_perf.metrics.llm_metrics import LLMMetrics
+
+
+class LLMGoodputReporter(GoodputReporter):
+    """A subclass to report goodput for language models."""
```
debermudez, line 35: We know its a subclass from the line above.

genai-perf/genai_perf/goodput_reporter/llm_goodput_reporter.py
```diff
@@ -0,0 +1,86 @@
...
+        self._valid_slos = {}
+        valid_names = [metric.name for metric in self._metric.request_metrics]
+
+        for slo_name, slo_value in self._goodput_constraints.items(): 
+            if self._metric.get_base_name(slo_name) not in valid_names:
+                invalid_slos.append(slo_name)
+            else:
+                self._valid_slos[slo_name] = slo_value * self.MS_TO_NS_CONVERSION
+        if invalid_slos:
+            raise ValueError(f"Invalid SLOs found: {', '.join(invalid_slos)}, "
```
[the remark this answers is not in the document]
debermudez: agree with the above

genai-perf/genai_perf/goodput_reporter/llm_goodput_reporter.py
debermudez: +1. Readability and Maintainability are the top priority. We optimize only if we find a bottleneck impacting performance.

genai-perf/genai_perf/goodput_reporter/llm_goodput_reporter.py
debermudez: This is what is called a side effect.
In general, we want to avoid those if they are not necessary. 
I agree with @nv-hwoo here.

debermudez (commented):
Lots of good work in this PR.
Lets some comments but great job.

genai-perf/docs/goodput_tutorial.md
```diff
@@ -0,0 +1,439 @@
+<!--
```
AndyDai-nv, line 1: The tutorial needs to be further refined, the main purpose here is to show goodput functionality for LLM, Embeddings and Rankings under three conditions: without goodput-constraints, with valid goodput constraints, and with invalid goodput constraints.

genai-perf/genai_perf/export_data/console_exporter.py
```diff
@@ -66,6 +66,15 @@ def export(self) -> None:
         # System metrics are printed after the table
         for metric in self._metrics.system_metrics:
             line = metric.name.replace("_", " ").capitalize()
+            if metric.name == "request_goodput":
+                if not self._args.goodput:
+                    continue
+                value = self._stats[metric.name]["avg"]
+                if value is None:
+                    value = "N/A"
```
dyastremsky, line 74: What if we make the value 0 in these cases? Then the code could be refactored to avoid this unique logic handling and continue. It would also make parsing a bit easier, because all of the parsed values would be ints or floats.

For example:
```
for metric in self._metrics.system_metrics:
           line = metric.name.replace("_", " ").capitalize()
           if metric.name == "request_goodput":
                if not self._args.goodput:
                    continue
           line += f" ({metric.unit}): {value}"
                    print(line)
                    continue
```

Then, you could set the value to 0 somewhere else if there's an error. You should log that error in the same place, if you're not already.

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+
+    def __init__(self,
+                goodput_constraints: Dict[str, float],
+                metric: LLMMetrics,
+                benchmark_duration: float,
+    ) -> None:
+        super().__init__(goodput_constraints, metric, benchmark_duration)
+        # (TMA-1975 related) The order is hardcoded as below due to the hardcoded order
+        # in LLMMetirc class. We would eventually want to impose some consistent order
```
dyastremsky, line 48: Nit: typo

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def __init__(self,
+                goodput_constraints: Dict[str, float],
+                metric: LLMMetrics,
+                benchmark_duration: float,
+    ) -> None:
+        super().__init__(goodput_constraints, metric, benchmark_duration)
+        # (TMA-1975 related) The order is hardcoded as below due to the hardcoded order
+        # in LLMMetirc class. We would eventually want to impose some consistent order 
+        # for time-related metrics and throughput related metrics.
```
dyastremsky, line 49: Nit: "metrics related to time and throughput"

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
```diff
@@ -0,0 +1,169 @@
...
+                invalid_slos.append(slo_name)
+        if self._valid_time_related_slos:
+            self._has_time_target = True
+        if self._valid_throughput_related_slos:
+            self._has_throughput_target = True 
+        if invalid_slos:
+            valid_slos_list = ', '.join(self._valid_metric_names)
+            logger.info(f"Invalid SLOs found: {', '.join(invalid_slos)}. "
+                        f"The goodput will be N/A. "
+                        f"Valid SLOs are: {valid_slos_list} in plural forms.")
```
dyastremsky, line 92: Is there a way to use singular forms? Like using get_base_name?

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
```diff
@@ -0,0 +1,183 @@
...
+        self._slo_base_names["time_to_first_token"] = "time_to_first_tokens"
+        self._slo_base_names["inter_token_latency"] = "inter_token_latencies"
+        self._slo_base_names["output_token_throughput_per_request"] = (
+            "output_token_throughputs_per_request"
+        )
+
+    def _set_valid_slos(self) -> None:
+        """
+        Check users' Service Level Objectives (SLOs) inputs.
+        Set the valid ones while logging the invalid ones.s
```
nv-hwoo, line 73: typo
```suggestion
        Set the valid ones while logging the invalid ones.
```

genai-perf/genai_perf/metrics/llm_metrics.py
```diff
@@ -26,22 +26,32 @@
...
 class LLMMetrics(Metrics):
     """A simple dataclass that holds core LLM performance metrics."""
 
-    LLM_REQUEST_METRICS = [
+    LLM_REQUEST_TIME_METRICS = [
         MetricMetadata("time_to_first_token", "ms"),
         MetricMetadata("inter_token_latency", "ms"),
+    ]
+
+    LLM_REQUEST_THROUGHPUT_METRICS = [
```
nv-hwoo, line 42: I'm a bit confused. Why do we need to distinguish time related metrics and throughput related metrics? Looking at the code, the only part that needs to treat them differently is when we convert time metric units:
```python
if slo_name in self._valid_time_related_names:
    self._valid_time_related_slos[slo_name] = (
        slo_value * self.MS_TO_NS_CONVERSION
    )
elif slo_name in self._valid_throughput_related_names:
    self._valid_throughput_related_slos[slo_name] = slo_value
```

AndyDai-nv: Because for time related metrics, we want our request values to be smaller than SLOs target, while for the throughput related metrics, we want our request values to be bigger than SLOs target.

genai-perf/genai_perf/parser.py
```diff
@@ -651,6 +685,17 @@ def _add_other_args(parser):
...
     )
 
+def _add_goodput_args(parser):
+    goodput_group = parser.add_argument_group("Goodput")
+
+    goodput_group.add_argument(
+        "--goodput",
+        "-g",
+        nargs='+',
+        required=False,
```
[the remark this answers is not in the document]
AndyDai-nv: That's true, I am just following the pattern in other args. Do we want to get rid of these?

genai-perf/genai_perf/parser.py
```diff
@@ -733,6 +778,7 @@ def _parse_profile_args(subparsers) -> argparse.ArgumentParser:
     _add_profile_args(profile)
     _add_output_args(profile)
     _add_other_args(profile)
+    _add_goodput_args(profile)
```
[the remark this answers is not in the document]
AndyDai-nv: Maybe more options for users to see exactly what metrics contribute most to the low goodput? Something like this.

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
```diff
@@ -0,0 +1,183 @@
...
+            )
+            self._goodput = self.INVALID_GOODPUT
+
+    def _combine_requests_metric_values(self) -> None:
+        """
+        Combine values from the metrics that match with the valid SLOs at a
+        per request level.
+        """
+        if self.goodput == self.INVALID_GOODPUT:
+            return
```
debermudez, line 106: what happens if these methods return early? do the rest of the code account for that?

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
```diff
@@ -0,0 +1,183 @@
...
+        """Count the number of good requests according to SLOs."""
+        if self.goodput == self.INVALID_GOODPUT:
+            return None
+        target_time_metric_values = []
+        target_throughput_metric_values = []
+        if self._has_time_target:
+            num_of_requests = len(self._combined_requests_time_metric_values)
+            target_time_metric_values = list(self._valid_time_related_slos.values())
+        if self._has_throughput_target:
+            num_of_requests = len(self._combined_requests_throughput_metric_values)
```
debermudez, line 143: if you have a throughput and time target, num_of_requests will be overwritten here.

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
```diff
@@ -0,0 +1,183 @@
...
+        Set the valid ones while logging the invalid ones.s
+        """
+        invalid_slos = []
+        self._valid_time_related_slos = {}
+        self._valid_throughput_related_slos = {}
+        for slo_name, slo_value in self._goodput_constraints.items():
+            if slo_name in self._valid_metric_names:
+                if slo_name in self._valid_time_related_names:
+                    self._valid_time_related_slos[slo_name] = (
+                        slo_value * self.MS_TO_NS_CONVERSION
```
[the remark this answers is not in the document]
AndyDai-nv: The data stored in metrics is in ns unit, it is for later comparison during counting good requests.

genai-perf/genai_perf/goodput_calculator/goodput_calculator.py
```diff
@@ -0,0 +1,104 @@
...
+    @abstractmethod
+    def _compute_goodput(self, good_count) -> None:
+        """Compute the goodput."""
+        pass
+
+    @property
+    def goodput(self) -> Optional[List[float]]:
+        return self._goodput
+
+    def get_slo_base_name(self, metric_name: str) -> str:
```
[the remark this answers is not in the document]
AndyDai-nv: Got it. How about just get_slo_name? We need the plural form as key to retrieve data. Adding this conversion from singular to plural is to enable users to use singular forms as inputs.

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
AndyDai-nv: Yes, only when there are invalid constraints in users' inputs, goodput will be INVALID_GOODPUT, the rest of code doesn't need to execute.

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
AndyDai-nv: Yeah, in that case the num_of_requests is still the same number which equals to the total number of requests.

debermudez: that is always valid?

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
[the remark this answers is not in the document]
debermudez: i see.
Do we require that users provide the SLOs in ms?

genai-perf/genai_perf/goodput_calculator/goodput_calculator.py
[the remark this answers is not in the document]
debermudez: get_slo_name works

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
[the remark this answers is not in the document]
AndyDai-nv: For time related metrics, yes. It is in help messages of args.

genai-perf/docs/goodput_tutorial.md
```diff
@@ -0,0 +1,558 @@
...
+GenAI-Perf is continually evolving to meet the demands of next-generation GenAI workflows. We are excited to introduce our latest feature: goodput support for measuring Service Level Objectives.
+
+This new feature empowers you to more precisely evaluate the efficiency and effectiveness of your AI models and services. By measuring goodput, GenAI-Perf delivers deeper insights into how well your services adhere to your defined SLOs, enabling you to optimize both performance and cost efficiency while prioritizing the user experience.
+
+# Tutorials
+
+This is a tutorial on how to use our goodput feature by examples.
+## Examples
+
+- [LLM Examples](#LLM)
```
dyastremsky, line 40: You don't need examples for every existing API. You can pick one or two and use them as examples to show how goodput works. By providing every API, we create the need to maintain them long-term and change them as the APIs update.

genai-perf/docs/goodput_tutorial.md
```diff
@@ -0,0 +1,558 @@
...
+
+Run Triton Inference Server with vLLM backend container:
+
+```bash
+export RELEASE="yy.mm" # e.g. export RELEASE="24.06"
+
+docker run -it --net=host --gpus=1 --shm-size=2g --ulimit memlock=-1 --ulimit stack=67108864 nvcr.io/nvidia/tritonserver:${RELEASE}-vllm-python-py3
+
+# Install Triton CLI (~5 min):
+pip install "git+https://github.com/triton-inference-server/triton_cli@0.0.8"
```
dyastremsky, line 60: Please use the vLLM container so that we don't need these long-running commands. Also, please go back through all of the feedback the team provided on this PR and respond. If you addressed the issue, then respond so that the person who made the comment can close it.

genai-perf/docs/goodput_tutorial.md
```diff
@@ -0,0 +1,558 @@
...
+│     Request latency (ms) │ 176.02 │ 165.97 │ 197.03 │ 196.99 │ 181.80 │ 177.86 │
+│   Output sequence length │ 113.34 │ 107.00 │ 127.00 │ 125.14 │ 118.00 │ 116.00 │
+│    Input sequence length │ 200.00 │ 200.00 │ 200.00 │ 200.00 │ 200.00 │ 200.00 │
+└──────────────────────────┴────────┴────────┴────────┴────────┴────────┴────────┘
+Output token throughput (per sec): 1287.08
+Request throughput (per sec): 11.36
+Request goodput (per sec): 10.65
+```
+
+### Run GenAI-Perf with invalid goodput constraints
```
dyastremsky, line 110: Please remove this. We don't need to show the invalid case.

genai-perf/docs/goodput_tutorial.md
```diff
@@ -0,0 +1,558 @@
...
+│     Request latency (ms) │ 176.93 │ 170.95 │ 197.70 │ 197.68 │ 177.86 │ 176.21 │
+│   Output sequence length │ 111.47 │ 103.00 │ 123.00 │ 122.38 │ 118.80 │ 114.00 │
+│    Input sequence length │ 200.00 │ 200.00 │ 200.00 │ 200.00 │ 200.00 │ 200.00 │
+└──────────────────────────┴────────┴────────┴────────┴────────┴────────┴────────┘
+Output token throughput (per sec): 1259.65
+Request throughput (per sec): 11.30
+Request goodput (per sec): -1.00
+```
+
+### Run GenAI-Perf with no goodput constraints
```
dyastremsky, line 152: We don't need to show this either.

genai-perf/docs/goodput_tutorial.md
```diff
@@ -0,0 +1,558 @@
...
+                           Embeddings Metrics
+┏━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━┳━━━━━━━━┳━━━━━━━┳━━━━━━━┳━━━━━━━┓
+┃            Statistic ┃   avg ┃   min ┃    max ┃   p99 ┃   p90 ┃   p75 ┃
+┡━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━━╇━━━━━━━━╇━━━━━━━╇━━━━━━━╇━━━━━━━┩
+│ Request latency (ms) │ 23.51 │ 21.84 │ 200.79 │ 23.17 │ 22.70 │ 22.47 │
+└──────────────────────┴───────┴───────┴────────┴───────┴───────┴───────┘
+Request throughput (per sec): 42.39
+```
+
+### Run GenAI-Perf with valid goodput constraints
```
dyastremsky, line 245: Every case shown should be valid, so "valid" would be redundant here.

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
AndyDai-nv: A little bit hard to answer this. : )
Yes, because for every completed request, we assume PA populates the information we need in the json and data parser would compute those metrics for the particular request. 
No, if there is something broken with PA. In that case, I believe GenAI-Perf won't work anyways.

genai-perf/genai_perf/goodput_calculator/goodput_calculator.py
```diff
@@ -0,0 +1,104 @@
...
+    def _combine_requests_metric_values(self) -> None:
+        """
+        Combine values from the metrics that match with the valid SLOs at a
+        per request level.
+        """
+        pass
+
+    @abstractmethod
+    def _count_good_reqs(self) -> Optional[int]:
+        """Count the number of good requests according to SLOs."""
```
dyastremsky, line 87: Please go through the code and remove the use of "SLOs". Someone then has to figure out what SLOs are if they're not already familiar. You can use a different term like "constraints" that are more universally understood.

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
```diff
@@ -0,0 +1,183 @@
...
+                    )
+                elif slo_name in self._valid_throughput_related_names:
+                    self._valid_throughput_related_slos[slo_name] = slo_value
```
dyastremsky, line 85: There should be an else here to catch errors. If/elif/else, not if/elif.

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
```diff
@@ -0,0 +1,183 @@
...
+                invalid_slos.append(slo_name)
+        if self._valid_time_related_slos:
+            self._has_time_target = True
+        if self._valid_throughput_related_slos:
+            self._has_throughput_target = True
+        if invalid_slos:
+            valid_slos_list = ", ".join(self._valid_metric_names)
+            logger.info(
+                f"Invalid SLOs found: {', '.join(invalid_slos)}. "
+                f"The goodput will be -1. Valid SLOs are: {valid_slos_list}."
```
dyastremsky, line 96: You can remove the statement that "The goodput will be -1".

genai-perf/genai_perf/goodput_calculator/goodput_calculator.py
```diff
@@ -0,0 +1,104 @@
...
+    def __init__(
+        self,
+        goodput_constraints: Dict[str, float],
+        metric: Metrics,
+        benchmark_duration: float,
+    ) -> None:
+        self._goodput_constraints = goodput_constraints
+        self._benchmark_duration = benchmark_duration
+        self._metric = metric
+        self._goodput: Optional[List[float]] = None
```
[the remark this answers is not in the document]
AndyDai-nv: How about adding comments to explain goodput?
Or change the name to _good_throughput? Or any better ideas?

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
```diff
@@ -0,0 +1,183 @@
...
+                )
+            if self._has_throughput_target:
+                request_throughput_metric_values = list(
+                    self._combined_requests_throughput_metric_values[idx]
+                )
+            for val, slo in zip(request_time_metric_values, target_time_metric_values):
+                if val > slo:
+                    is_good_request = False
+                    break
+            else:
```
dyastremsky, line 165: I think the indentation is off, so I'm surprised the code works. Where does this else clause belong to?

genai-perf/genai_perf/parser.py
```diff
@@ -286,6 +301,22 @@ def _set_artifact_paths(args: argparse.Namespace) -> argparse.Namespace:
...
 
+def parse_goodput(values):
+    constraints = {}
+    try:
+        for item in values:
+            target_metric, target_val = item.split(":")
+            constraints[target_metric] = float(target_val)
+    except ValueError:
+        raise argparse.ArgumentTypeError(
+            f"Invalid format for goodput constraints: {values}. "
```
dyastremsky, line 312: ``
```suggestion
            f"Invalid format found for goodput constraints. "
```

genai-perf/genai_perf/parser.py
```diff
@@ -286,6 +301,22 @@ def _set_artifact_paths(args: argparse.Namespace) -> argparse.Namespace:
...
+            f"The expected format is 'key:value' pairs, where the key should be a "
```
dyastremsky, line 313:
```suggestion
            f"The expected format is 'key:value' pairs. The key should be a "
```

genai-perf/genai_perf/parser.py
```diff
@@ -286,6 +301,22 @@ def _set_artifact_paths(args: argparse.Namespace) -> argparse.Namespace:
...
+            f"valid service level objective name and the value should be a number "
```
dyastremsky, line 314:
```suggestion
            f"service level objective name (e.g. request_latency). The value should be a number "
```

genai-perf/genai_perf/parser.py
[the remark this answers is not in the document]
dyastremsky: I see. Some args include the optional required and some do not. We can leave this one.

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
AndyDai-nv: It is a relatively rare usage. The `else` clause belongs to previous `for` loop in order to avoid unnecessary loops.
https://book.pythontips.com/en/latest/for_-_else.html#:~:text=for%20loops%20also%20have%20an%20else%20clause%20which%20most%20of%20us%20are%20unfamiliar%20with.%20The%20else%20clause%20executes%20after%20the%20loop%20completes%20normally.

genai-perf/tests/test_llm_goodput_calculator.py
```diff
@@ -0,0 +1,150 @@
...
+        assert gc.get_slo_base_name("request_latency") == "request_latencies"
+        assert gc.get_slo_base_name("time_to_first_token") == "time_to_first_tokens"
+        assert gc.get_slo_base_name("inter_token_latency") == "inter_token_latencies"
+        assert gc.get_slo_base_name("output_token_throughput_per_request") == (
+            "output_token_throughputs_per_request"
+        )
+        with pytest.raises(KeyError):
+            gc.get_slo_base_name("hello1234")
+
+    def test_llmgoodputcalculator_compute(self) -> None:
```
dyastremsky, line 80:
```suggestion
    def test_l_compute(self) -> None:
```

genai-perf/tests/test_llm_goodput_calculator.py
```diff
@@ -0,0 +1,150 @@
...
+        """
+        Test compute method
```
dyastremsky, line 82:
```suggestion
```
This comment can be trimmed down, since this line is redundant. This information is already in the function name, which is the ideal approach: self-documenting code.

genai-perf/tests/test_llm_goodput_calculator.py
```diff
@@ -0,0 +1,150 @@
...
+
+        Goodput constraints for experiment 1 and 2:
+        * time_to_first_token: 2.5e-6 ms
+        * inter_token_latency: 2.5e-6 ms
+        * output_token_throughput_per_request: 0.5e9 s
+
+        Benchmark durationfor experiment 1 and 2: 10 s
```
dyastremsky, line 89:
```suggestion
        Benchmark duration for experiment 1 and 2: 10 s
```

genai-perf/tests/test_llm_metrics.py
```diff
@@ -57,6 +58,46 @@ def test_llm_metric_request_metrics(self) -> None:
         assert req_metrics[5].name == "input_sequence_length"
         assert req_metrics[5].unit == "tokens"
 
+    def test_llm_metric_request_time_metrics(self) -> None:
```
dyastremsky, line 61:
```suggestion
    def test_request_time_metrics(self) -> None:
```

dyastremsky (commented):
Wonderful work, Andy! This is looking good.

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
dyastremsky: Interesting. Is there another way to do this? Something that's rare and unknown is not good for code readability. If there is *no* other way, then we could add formatting to separate this for/else into its own block and add a comment explaining that's happening. However, in this case, we're talking about the code execution path being unclear or misunderstood, so that's dangerous.

One way to do this could be to check if is_good_request is False.

genai-perf/docs/goodput_tutorial.md
AndyDai-nv: Updated

genai-perf/genai_perf/metrics/llm_metrics.py
AndyDai-nv: Updated

genai-perf/genai_perf/metrics/statistics.py
AndyDai-nv: Updated

genai-perf/tests/test_llm_metrics.py
AndyDai-nv: Updated

genai-perf/genai_perf/goodput_reporter/llm_goodput_reporter.py
```diff
@@ -0,0 +1,86 @@
...
+    def set_valid_slos(self) -> None:
+        """Check user's Service Level Objectives (SLOs) inputs. 
+        Set the valid ones or raise error for the invalid ones. 
+        """
+        invalid_slos = []
+        self._valid_slos = {}
+        valid_names = [metric.name for metric in self._metric.request_metrics]
+
+        for slo_name, slo_value in self._goodput_constraints.items(): 
+            if self._metric.get_base_name(slo_name) not in valid_names:
```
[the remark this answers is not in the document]
AndyDai-nv: Updated

genai-perf/genai_perf/goodput_reporter/llm_goodput_reporter.py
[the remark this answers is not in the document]
AndyDai-nv: Updated. Changed to -1 according to @dyastremsky's suggestion

genai-perf/genai_perf/goodput_reporter/llm_goodput_reporter.py
```diff
@@ -0,0 +1,86 @@
...
+            if all(val < slo 
+                   for val, slo in zip(request_metric_values, target_metric_values)
+            ):
+                good_req_count += 1
+        self._good_req_count = good_req_count
+    
+    def compute_goodput(self) -> None:
+        """Compute the goodput."""
+        self._goodput = [self._good_req_count / self._benchmark_duration]
+
```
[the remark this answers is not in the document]
AndyDai-nv: Updated

genai-perf/genai_perf/parser.py
```diff
@@ -285,6 +307,18 @@ def _set_artifact_paths(args: argparse.Namespace) -> argparse.Namespace:
...
+def parse_goodput(values):
+    constraints = {}
+    try:
+        for item in values:
+            target_metric, target_val = item.split(':')
+            constraints[target_metric] = float(target_val)
+    except ValueError:
+        raise argparse.ArgumentTypeError(
+            f"Invalid goodput constraints format: {values}. "
+            "Expected format is 'ttft:x itl:y', where x and y are numbers in milliseconds."
```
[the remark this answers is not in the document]
AndyDai-nv: Updated

genai-perf/genai_perf/parser.py
```diff
@@ -651,6 +685,17 @@ def _add_other_args(parser):
...
 
+def _add_goodput_args(parser):
+    goodput_group = parser.add_argument_group("Goodput")
+
+    goodput_group.add_argument(
+        "--goodput",
+        "-g",
+        nargs='+',
+        required=False,
+        help="The goodput constraints are in the format of 'ttft:x itl:y', "
```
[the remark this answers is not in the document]
AndyDai-nv: Updated

genai-perf/docs/goodput_tutorial.md
```diff
@@ -0,0 +1,102 @@
...
+┃                Statistic ┃    avg ┃    min ┃    max ┃    p99 ┃    p90 ┃    p75 ┃
+┡━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━┩
+│ Time to first token (ms) │   9.30 │   5.78 │  20.62 │  20.61 │  18.51 │  10.00 │
+│ Inter token latency (ms) │   2.22 │   1.74 │   3.01 │   2.96 │   2.78 │   2.28 │
+│     Request latency (ms) │ 256.65 │ 216.91 │ 345.47 │ 345.46 │ 303.17 │ 267.83 │
+│   Output sequence length │ 112.41 │ 106.00 │ 125.00 │ 123.71 │ 117.70 │ 115.00 │
+│    Input sequence length │ 200.00 │ 200.00 │ 200.00 │ 200.00 │ 200.00 │ 200.00 │
+└──────────────────────────┴────────┴────────┴────────┴────────┴────────┴────────┘
+Output token throughput (per sec): 1751.15
+Request goodput (per sec): 2.83
```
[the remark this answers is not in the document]
AndyDai-nv: Updated

genai-perf/genai_perf/goodput_reporter/llm_goodput_reporter.py
```diff
@@ -0,0 +1,86 @@
...
+        if invalid_slos:
+            raise ValueError(f"Invalid SLOs found: {', '.join(invalid_slos)}, "
+                            "Make sure these are supported request metrics.")
+
+    def combine_requests_metric_values(self) -> None:
+        """Combine metric values at per request level.
+        Only the metrics from valid SLOs.
+        """
+        metric_data = self._metric.data 
+        requests_metric_values = [metric_data[key] for key in self._valid_slos]
```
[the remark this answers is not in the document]
AndyDai-nv: Updated

genai-perf/genai_perf/goodput_reporter/goodput_reporter.py
```diff
@@ -0,0 +1,91 @@
...
+        self._metric = metric
+        self._goodput = None
+    
+    def report(self) -> None:
+        """Template method to report goodput. Subclasses should not modify this method.
+        """
+        self.set_valid_slos()
+        self.combine_requests_metric_values()
+        self.count_good_reqs()
+        self.compute_goodput()
```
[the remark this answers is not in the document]
AndyDai-nv: Updated

genai-perf/genai_perf/goodput_reporter/goodput_reporter.py
```diff
@@ -0,0 +1,91 @@
...
+            benchmark_duration: float,
+    ) -> None:
+        self._goodput_constraints = goodput_constraints
+        self._benchmark_duration = benchmark_duration
+        self._metric = metric
+        self._goodput = None
+    
+    def report(self) -> None:
+        """Template method to report goodput. Subclasses should not modify this method.
+        """
```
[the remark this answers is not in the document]
AndyDai-nv: Updated

genai-perf/genai_perf/goodput_reporter/goodput_reporter.py
```diff
@@ -0,0 +1,91 @@
...
+        """Check user's Service Level Objectives (SLOs) inputs. 
+        Set the valid ones while logging the invalid ones. 
+        To be implemented by subclasses.
+        """
+        pass
+    
+    @abstractmethod
+    def combine_requests_metric_values(self) -> None:
+        """Combine metric values at per request level.
+        Only the metrics from valid SLOs.
```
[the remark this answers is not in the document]
AndyDai-nv: Updated

genai-perf/genai_perf/goodput_reporter/llm_goodput_reporter.py
AndyDai-nv: Updated

genai-perf/genai_perf/goodput_reporter/llm_goodput_reporter.py
AndyDai-nv: Updated

genai-perf/genai_perf/goodput_reporter/llm_goodput_reporter.py
AndyDai-nv: Updated

genai-perf/genai_perf/parser.py
```diff
@@ -244,6 +244,28 @@ def _check_load_manager_args(args: argparse.Namespace) -> argparse.Namespace:
         args.concurrency = 1
     return args
 
+def _check_goodput_args(args):
+    """
+    Parse and check goodput args
+    """
+    '''
+    if args.goodput:
```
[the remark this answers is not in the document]
AndyDai-nv: Updated

genai-perf/docs/goodput_tutorial.md
```diff
@@ -0,0 +1,103 @@
...
+### Run GenAI-Perf
+
+Run GenAI-Perf from Triton Inference Server SDK container:
+
+```bash
+export RELEASE="yy.mm" # e.g. export RELEASE="24.06"
+
+docker run -it --net=host --gpus=1 nvcr.io/nvidia/tritonserver:${RELEASE}-py3-sdk
+
+# Run GenAI-Perf in the container:
```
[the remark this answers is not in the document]
AndyDai-nv: Updated

genai-perf/genai_perf/export_data/console_exporter.py
AndyDai-nv: Updated

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
AndyDai-nv: Updated

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
AndyDai-nv: Updated

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
AndyDai-nv: Updated

genai-perf/docs/goodput_tutorial.md
AndyDai-nv: Updated

genai-perf/genai_perf/goodput_calculator/goodput_calculator.py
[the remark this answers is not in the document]
AndyDai-nv: Updated. Add a comment to explain.

genai-perf/docs/goodput_tutorial.md
AndyDai-nv: Updated

genai-perf/docs/goodput_tutorial.md
AndyDai-nv: Updated

genai-perf/docs/goodput_tutorial.md
AndyDai-nv: Updated the tutorial, skip replying the rest of this file.

genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
AndyDai-nv: Updated

genai-perf/genai_perf/parser.py
AndyDai-nv: Updated all the text change. Skip replying to the rest for simplicity.

genai-perf/tests/test_console_exporter.py
```diff
@@ -170,6 +181,115 @@ def test_embedding_output(self, monkeypatch, capsys) -> None:
             "│ Request latency (ms) │ 5.00 │ 4.00 │ 6.00 │ 5.98 │ 5.80 │ 5.50 │\n"
             "└──────────────────────┴──────┴──────┴──────┴──────┴──────┴──────┘\n"
             "Request throughput (per sec): 123.00\n"
+            "Request goodput (per sec): 100.00\n"
```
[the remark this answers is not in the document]
AndyDai-nv: Updated unit tests. Invert other test cases, only keep goodput ones. Skip replying to the rest for simplicity.

genai-perf/docs/goodput_tutorial.md
```diff
@@ -0,0 +1,134 @@
+<!--
```
dyastremsky, line 1: Please name this file consistently with the other tutorial docs. E.g. "goodput.md"

genai-perf/docs/goodput_tutorial.md
```diff
@@ -0,0 +1,134 @@
...
+PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+-->
+
+# Context
+
+Goodput, defined as the number of completed requests per second that meet the Service Level Objectives (SLOs), provides an enhanced measure of AI serving performance by accounting for both cost efficiency and user satisfaction.
```
dyastremsky, line 29: We use a line length max of 80. These lines should be split up.

genai-perf/docs/goodput_tutorial.md
```diff
@@ -0,0 +1,134 @@
...
+
+GenAI-Perf is continually evolving to meet the demands of next-generation GenAI workflows. We are excited to introduce our latest feature: goodput support for measuring Service Level Objectives.
```
dyastremsky, line 31: Please shorten these to be more technical. While briefly citing potential uses is good, a lot of these words/sentences don't say anything. Focus on making this high-quality technical documentation rather than high-quality marketing documents.

Also, please remember that this tutorial will need to be up for a long time after it is merged. So time-relevant things like about whether we are introducing a feature would go out in marketing materials, not in our technical documentation.

genai-perf/docs/goodput_tutorial.md
dyastremsky: You should avoid using terms that are not defined if they are not ubiquitous. Specifically, SLOs are really just constraints that the user specifies. If you can define them first, you can then use them in the documentation. However, since this is more technical documentation, the preferred approach is to use simple language (e.g. constraints) to allow users to skim the documentation to find exactly what they need without reading start to end.

genai-perf/docs/goodput_tutorial.md
```diff
@@ -0,0 +1,134 @@
...
+
+This new feature empowers you to more precisely evaluate the efficiency and effectiveness of your AI models and services. By measuring goodput, GenAI-Perf delivers deeper insights into how well your services adhere to your defined SLOs, enabling you to optimize both performance and cost efficiency while prioritizing the user experience.
+
+# Tutorials
+
+This is a tutorial on how to use our goodput feature by examples.
```
dyastremsky, line 37:
```suggestion
This is a tutorial on how to use benchmark models using goodput.
```

genai-perf/docs/goodput_tutorial.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Examples
+
+- [LLM Examples](#LLM)
+
+- [Embedding Model Examples](#embeddings)
+
+## Profile GPT2 Goodput<a id="LLM"></a>
```
dyastremsky, line 45: You should aim to be consistent here. Since you use LLM throughout, the title should be Profile LLM Goodput so that it's obvious that this is the "LLM Examples" section.

I'd recommend using the terminology "Chat Completions" instead though to be more precise and clear with which API is being used.

genai-perf/docs/goodput_tutorial.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Run GPT2 on OpenAI Chat Completions API-compatible server
+
+```bash
+docker run -it --net=host --rm --gpus=all vllm/vllm-openai:latest --model gpt2 --dtype float16 --max-model-len 1024
+```
+
+### Run GenAI-Perf with goodput constraints
```
dyastremsky, line 53: All titles should be capitalized.

dyastremsky (commented):
Great work! I reviewed the tutorial and will review the rest of the PR later.

Similar to @matthewkotila's suggestion, it would be great to break up this PR to make it easier to review. I understand if you do not want to break it up at this point, but if it's at all possible, that would help reviews be more thorough and help us review/iterate faster. It's unconventional for us to have a PR that's 2000 lines to review.

At a minimum, I'd pull tutorial.md into its own PR. You can break up this PR into implementation, testing, and documentation or similar. The PRs can target a feature branch so that each PR can be approved, then the final PR is just a list of already-approved PRs. If you choose to do this, let me know if you need any help doing so.

genai-perf/genai_perf/metrics/llm_metrics.py
```diff
@@ -26,22 +26,32 @@
...
         MetricMetadata("output_token_throughput_per_request", "tokens/sec"),
-        MetricMetadata("output_sequence_length", "tokens"),
-        MetricMetadata("input_sequence_length", "tokens"),
     ]
 
+    LLM_REQUEST_METRICS = (
+        LLM_REQUEST_TIME_METRICS
+        + LLM_REQUEST_THROUGHPUT_METRICS
+        + [
+            MetricMetadata("output_sequence_length", "tokens"),
```
dyastremsky, line 50: You should make a separate variable to hold LLM_REQUEST_OTHER_METRICS (including the other metrics e.g. output sequence length), so that we keep the abstraction level the same. Then, this variable can combine all three of those variables.

genai-perf/genai_perf/parser.py
```diff
@@ -245,6 +245,21 @@ def _check_load_manager_args(args: argparse.Namespace) -> argparse.Namespace:
...
+    """
+    Parse and check goodput args
+    """
+    if args.goodput:
+        args.goodput = parse_goodput(args.goodput)
+        for target_metric, target_val in args.goodput.items():
+            if target_val < 0:
+                raise ValueError(
+                    f"Invalid value found, {target_metric}: {target_val}. "
+                    f"The Service Level Objective value should be non-negative. "
```
dyastremsky, line 258:
```suggestion
                    f"The goodput constraint value should be non-negative. "
```

genai-perf/genai_perf/parser.py
```diff
@@ -652,6 +684,23 @@ def _add_other_args(parser):
...
 
+def _add_goodput_args(parser):
+    goodput_group = parser.add_argument_group("Goodput")
+
+    goodput_group.add_argument(
+        "--goodput",
+        "-g",
+        nargs="+",
+        required=False,
+        help="An option to provide Service Level Objectives to compute goodput. "
```
dyastremsky, line 695:
```suggestion
        help="An option to provide constraints in order to compute goodput, "
        "the throughput only including those requests that meet those constraints."
```

genai-perf/genai_perf/parser.py
```diff
@@ -652,6 +684,23 @@ def _add_other_args(parser):
...
+        "Specify goodput constraints as 'key:value' pairs, where the key is a "
+        "valid Service Level Objective name, and the value is a number representing "
```
dyastremsky, line 697:
```suggestion
        "valid metric name, and the value is a number representing "
```

genai-perf/genai_perf/parser.py
```diff
@@ -652,6 +684,23 @@ def _add_other_args(parser):
     )
 
 
+def _add_goodput_args(parser):
+    goodput_group = parser.add_argument_group("Goodput")
```
dyastremsky, line 688: We shouldn't be adding a whole group just for this specific feature. I would put this under the input group.

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
```diff
@@ -161,6 +163,19 @@ def _parse_requests(self, requests: dict) -> Metrics:
             chunked_inter_token_latencies,
         )
 
+        # request goodput
```
dyastremsky, line 166: Could you break the goodput calculation out into its own function that gets called here? This function is quite long and needs to be broken out at some point. I'd rather not extend it further if not needed.

genai-perf/genai_perf/profile_data_parser/profile_data_parser.py
```diff
@@ -136,11 +142,23 @@ def _parse_requests(self, requests: dict) -> Metrics:
         benchmark_duration = (max_res_timestamp - min_req_timestamp) / 1e9  # to seconds
         request_throughputs = [len(requests) / benchmark_duration]
 
-        return Metrics(
+        metric = Metrics(
             request_throughputs,
             request_latencies,
         )
 
+        # request goodput
```
dyastremsky, line 150: Could you break this out into its own function that gets called here?

genai-perf/tests/test_json_exporter.py
```diff
@@ -358,3 +359,555 @@ def test_generate_json_custom_export(
         }
       }
     """
+
+    def test_valid_goodput_json_output(
```
dyastremsky, line 363: Since these test cases are just testing goodput, can you avoid making large data structs specific to goodput and just check the goodput-specific lines you expect are in the output? You don't need the other checks.

genai-perf/tests/test_csv_exporter.py
```diff
@@ -228,3 +228,133 @@ def test_embedding_csv_output(
         ]
         returned_data = [data for _, data in mock_read_write]
         assert returned_data == expected_content
+
```
dyastremsky, line 231: Since these test cases are just testing goodput, can you just check the goodput-specific lines you expect are in the output? You don't need the other checks. They are already in the general purpose unit tests.

genai-perf/tests/test_llm_goodput_calculator.py
```diff
@@ -0,0 +1,146 @@
...
+            inter_token_latencies=[2, 1],
+            output_token_throughputs_per_request=[3 / ns_to_sec(7), 6 / ns_to_sec(9)],
+        )
+
+        gc_1 = LLMGoodputCalculator(
+            goodput_constraints=test_goodput_constraints,
+            metric=test_llm_metrics_1,
+            benchmark_duration=self.TEST_BENCHMARK_DURATION,
+        )
+        assert gc_1.goodput is None  # before computing
```
dyastremsky, line 128: These comments seem like they can be removed. The code is clear and self-commenting here.

genai-perf/tests/test_llm_profile_data_parser.py
```diff
@@ -136,6 +145,8 @@ def test_triton_llm_profile_data(self, mock_read_write: pytest.MonkeyPatch) -> N
         assert metrics.output_token_throughputs == pytest.approx(ott)
         assert metrics.output_sequence_lengths == [3, 6]
         assert metrics.input_sequence_lengths == [3, 4]
+        gp = [1 / ns_to_sec(10)]
```
dyastremsky, line 148: I recommend writing out the whole variable here. It's not immediately clear that gp == goodput.

genai-perf/docs/goodput_tutorial.md
AndyDai-nv: Please see updated tutorial in a separate PR

genai-perf/genai_perf/goodput_calculator/goodput_calculator.py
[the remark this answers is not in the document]
AndyDai-nv: I believe there is a reason that throughput and latency are lists. So I keep goodput in the same form

genai-perf/genai_perf/goodput_calculator/goodput_calculator.py
```diff
@@ -0,0 +1,106 @@
...
+        self._slo_names = {
+            "request_latency": "request_latencies",
+        }
+
+    def compute(self) -> None:
+        """
+        Compute the goodput result.
+
+        The GoodputCalculator class sets valid goodput constraints from users'
```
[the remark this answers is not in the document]
AndyDai-nv: Maybe change `GoodputCalculator class` to `compute method`? The content in the comment is for readability better understanding goodput computation process

genai-perf/genai_perf/goodput_calculator/goodput_calculator.py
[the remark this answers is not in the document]
dyastremsky: I see. It may be important to keep it consistent then.

If goodput is just a single value in the list, I'd add a comment stating that ._goodput is a list with the goodput as its only value.

genai-perf/genai_perf/goodput_calculator/goodput_calculator.py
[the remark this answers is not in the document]
dyastremsky: Sure! That sounds good.

genai-perf/docs/goodput_tutorial.md
dyastremsky: Perfect, nice job!

genai-perf/tests/test_llm_profile_data_parser.py
AndyDai-nv: Updated according to suggestions from above reviews, except for discussed ones.

dyastremsky (commented):
Two small comments. Will finish the review soon.

genai-perf/tests/test_llm_profile_data_parser.py
```diff
@@ -200,21 +201,140 @@ def test_triton_llm_profile_data(
...
+        # experiment 1 metrics & statistics
+        stat_obj = pd.get_statistics(infer_mode="concurrency", load_level="10")
+        metrics = stat_obj.metrics
+        stat = stat_obj.stats_dict
 
-        expected_metrics = LLMMetrics(**expected_metrics)
-        expected_statistics = Statistics(expected_metrics)
+        assert isinstance(metrics, LLMMetrics)
 
-        check_llm_metrics(metrics, expected_metrics)
```
dyastremsky, line 216: Why do you need to add all of the asserts instead of using the check functions? It's often a bad sign if we're making a test a lot longer, especially if the changes do not seem relevant to the feature (i.e. goodput).

dyastremsky (commented):
One more comment.

This looks pretty close to being ready to merge. Great work, Andy!

genai-perf/tests/test_llm_profile_data_parser.py
AndyDai-nv: I believe this is a recently merged update made by other people. I will solve the conflicts or reuse the function accordingly.

dyastremsky: Ah, I see. Thanks!

genai-perf/genai_perf/goodput_calculator/goodput_calculator.py
```diff
@@ -0,0 +1,103 @@
...
+    def __init__(
+        self,
+        goodput_constraints: Dict[str, float],
+        metric: Metrics,
+        benchmark_duration: float,
+    ) -> None:
+        self._goodput_constraints = goodput_constraints
+        self._benchmark_duration = benchmark_duration
+        self._metric = metric
+        # goodput is defined as the number of completed requests per second
```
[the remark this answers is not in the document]
AndyDai-nv: Updated according to all suggestions in this review cycle

dyastremsky (approved):
Fantastic work, Andy! Looks good to me. 🚀

PR #37, opened 2024-08-13
Add CLI argument for server metrics url

lkomali:
This PR adds a CLI argument to get server metrics url when the server is running on a machine different from the one genai-perf is running.

- Added a --server-metrics-url to endpoint arguments
- Check the server_metrics_url if all conditions are met
- Test different scenarios - default case, non-default case, invalid url
- Added a check if url is reachable or not to catch the error gracefully and display it

### DIFF
--- a/genai-perf/genai_perf/parser.py
+++ b/genai-perf/genai_perf/parser.py
@@ -38,6 +38,7 @@ from genai_perf.constants import (
     CNN_DAILY_MAIL,
     DEFAULT_ARTIFACT_DIR,
     DEFAULT_COMPARE_DIR,
+    DEFAULT_TRITON_METRICS_URL,
     OPEN_ORCA,
 )
 from genai_perf.llm_inputs.llm_inputs import (
@@ -765,9 +766,23 @@ def compare_handler(args: argparse.Namespace):
 
 
 def profile_handler(args, extra_args):
+    from genai_perf.telemetry_data.triton_telemetry_data_collector import (
+        TritonTelemetryDataCollector,
+    )
     from genai_perf.wrapper import Profiler
 
-    Profiler.run(args=args, extra_args=extra_args)
+    telemetry_data_collector = None
+    if args.service_kind == "triton":
+        # TPA-275: pass server url as a CLI option in non-default case
+        telemetry_data_collector = TritonTelemetryDataCollector(
+            server_metrics_url=DEFAULT_TRITON_METRICS_URL
+        )
+
+    Profiler.run(
+        args=args,
+        extra_args=extra_args,
+        telemetry_data_collector=telemetry_data_collector,
+    )
 
 
 ### Parser Initialization ###
--- a/genai-perf/genai_perf/telemetry_data/telemetry_data_collector.py
+++ b/genai-perf/genai_perf/telemetry_data/telemetry_data_collector.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python3
+
+# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of NVIDIA CORPORATION nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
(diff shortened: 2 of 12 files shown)


genai-perf/genai_perf/parser.py
```diff
@@ -246,6 +246,49 @@ def _check_load_manager_args(args: argparse.Namespace) -> argparse.Namespace:
     return args
 
 
+def is_valid_url(url):
+    """
+    Checks if a URL is valid. It must use 'http' or 'https', have a valid
+    netloc, optional port, and contain '/metrics' in the path.
```
debermudez, line 252: can we add this to the docstring:
```
<scheme>://<netloc>/<path>;<params>?<query>#<fragment>
```
so we know what the pattern is for the url?

genai-perf/genai_perf/parser.py
```diff
@@ -246,6 +246,49 @@ def _check_load_manager_args(args: argparse.Namespace) -> argparse.Namespace:
...
+        if not result.netloc:
+            return False
+
+        # Check path contains '/metrics'
+        if "/metrics" not in result.path:
+            return False
+
+        return True
+    except Exception:
+        return False
```
debermudez, line 271: Do we just want to return false here?

genai-perf/genai_perf/parser.py
```diff
@@ -246,6 +246,49 @@ def _check_load_manager_args(args: argparse.Namespace) -> argparse.Namespace:
...
+    Checks if the server metrics URL is valid based on the backend and service kind.
+    """
+
+    # Check if the URL is valid and contains the expected path
+    if (
+        args.service_kind == "triton"
+        and args.server_metrics_url
+        and not is_valid_url(args.server_metrics_url)
+    ):
+        parser.error("The --server-metrics-url option contains an invalid URL format.")
```
debermudez, line 287: I would move this exception into is_valid_url.
And update this method to check service kind and existence of the metrics url. Then return the result of _is_valid_url. 

This way, as we add more checks later, we can make the call to the url validation method and it will bubble up the exception if one is thrown.

genai-perf/genai_perf/parser.py
```diff
@@ -604,6 +647,14 @@ def _add_endpoint_args(parser):
         help="URL of the endpoint to target for benchmarking.",
     )
 
+    endpoint_group.add_argument(
+        "--server-metrics-url",
```
debermudez, line 651: Lets keep these alphabetized based on the option name.

genai-perf/genai_perf/telemetry_data/telemetry_data_collector.py
```diff
@@ -59,6 +59,16 @@ def stop(self) -> None:
             self._stop_event.set()
             self._thread.join()
 
+    def check_url_reachability(self) -> bool:
+        """Check if the server metrics URL is reachable and valid."""
+        if self._server_metrics_url:
+            try:
+                response = requests.get(self._server_metrics_url, timeout=5)
+                return response.status_code == 200
```
debermudez, line 67:
```suggestion
                return response.status_code == response.codes.ok
```

genai-perf/genai_perf/parser.py
```diff
@@ -604,6 +647,14 @@ def _add_endpoint_args(parser):
...
+        type=str,
+        default=None,
+        required=False,
+        help="URL of the server metrics endpoint. Required for 'openai' service kind. Defaults to the default URL if 'service_kind' is 'triton'.",
```
dyastremsky, line 655: This doesn't seem to be required. It'd be good to avoid adding required args.

dyastremsky (commented):
Great work! Can you add a unit test that tests the functionality you added (not just the parsing)?

genai-perf/genai_perf/parser.py
lkomali: This function returns if url is valid or not and the _check_server_metrics_url will display error message based on the return value.

genai-perf/genai_perf/parser.py
```diff
@@ -245,6 +245,85 @@ def _check_load_manager_args(args: argparse.Namespace) -> argparse.Namespace:
...
+    if args.service_kind == 'triton' and args.server_metrics_url and not is_valid_url(args.server_metrics_url):
+            parser.error("The --server-metrics-url option contains an invalid URL format.")
+
+    return args
+
+
+def is_valid_url(url):
+    """
+    Checks if a URL is valid. It must use 'http' or 'https', have a valid
+    netloc, optional port, and contain '/metrics' in the path.
```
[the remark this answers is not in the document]
lkomali: Yeah, I will update the comment. I had a check for port earlier but removed it after reading about netloc.

genai-perf/genai_perf/parser.py
lkomali: I will modify the comment.

PR #39, opened 2024-08-15
Add CLI argument for server-metrics-url

lkomali:
This is a new PR for adding a CLI argument to get server-metrics-url to collect telemetry data when the server is running on a different machine than genai-perf.
1. Add a new endpoint arg --server-metrics-url
2. Add checks for server-metrics-url
3. Add a check if url is reachable.
4. Unit tests

### DIFF
--- a/genai-perf/genai_perf/parser.py
+++ b/genai-perf/genai_perf/parser.py
@@ -31,6 +31,7 @@ import sys
 from enum import Enum, auto
 from pathlib import Path
 from typing import Tuple
+from urllib.parse import urlparse
 
 import genai_perf.logging as logging
 import genai_perf.utils as utils
@@ -247,6 +248,49 @@ def _check_load_manager_args(args: argparse.Namespace) -> argparse.Namespace:
     # If no concurrency or request rate is set, default to 1
     if not args.concurrency and not args.request_rate:
         args.concurrency = 1
+    return args
+
+
+def _is_valid_url(url: str, parser: argparse.ArgumentParser) -> None:
+    """
+    Validates a URL to ensure it meets the following criteria:
+    - The scheme must be 'http' or 'https'.
+    - The netloc (domain) must be present.
+    - The path must contain '/metrics'.
+
+    Raises:
+        `parser.error()` if the URL is invalid.
+
+    The URL structure is expected to follow:
+    <scheme>://<netloc>/<path>;<params>?<query>#<fragment>
+    """
+    parsed_url = urlparse(url)
+
+    if (
+        parsed_url.scheme not in ["http", "https"]
+        or not parsed_url.netloc
+        or "/metrics" not in parsed_url.path
+        or parsed_url.port is None
+    ):
+        parser.error(
+            "The URL passed for --server-metrics-url is invalid. "
+            "It must use 'http' or 'https', have a valid domain, "
+            "and contain '/metrics' in the path. The expected structure is: "
+            "<scheme>://<netloc>/<path>;<params>?<query>#<fragment>"
+        )
+
+
+def _check_server_metrics_url(
+    parser: argparse.ArgumentParser, args: argparse.Namespace
+) -> argparse.Namespace:
+    """
+    Checks if the server metrics URL passed is valid
+    """
+
+    # Check if the URL is valid and contains the expected path
+    if args.service_kind == "triton" and args.server_metrics_url:
+        _is_valid_url(args.server_metrics_url, parser)
+
     return args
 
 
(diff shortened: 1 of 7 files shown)


genai-perf/genai_perf/parser.py
```diff
@@ -269,6 +270,49 @@ def _check_goodput_args(args):
...
+def _check_server_metrics_url(
+    parser: argparse.ArgumentParser, args: argparse.Namespace
+) -> argparse.Namespace:
+    """
+    Checks if the server metrics URL passed is valid
+    """
+
+    # Check if the URL is valid and contains the expected path
+    if args.service_kind == "triton" and args.server_metrics_url:
+        _is_valid_url(args.server_metrics_url, parser)
```
debermudez, line 311: I would keep the order consistent.
This method call takes something then the parser but the method call _check_server_metrics_url does parser then something else. 
Lets standardize on an order.

genai-perf/genai_perf/parser.py
```diff
@@ -650,6 +694,14 @@ def _add_endpoint_args(parser):
         "you must specify an api via --endpoint-type.",
     )
 
+    endpoint_group.add_argument(
+        "--server-metrics-url",
+        type=str,
+        default=None,
+        required=False,
+        help="The full URL to access the server metrics endpoint. This argument is required if the metrics are available on a different machine than localhost (where GenAI-Perf is running).",
```
debermudez, line 702: This is inconsistent with url + endpoint above.
The problem is that the url + endpoint should be passed in as a single arg and we slice it internally but for now its inconsistent.
This isnt wrong but it might cause UX issues.

genai-perf/genai_perf/wrapper.py
```diff
@@ -101,7 +101,9 @@ def build_cmd(args: Namespace, extra_args: Optional[List[str]] = None) -> List[s
             "image_height_mean",
             "image_height_stddev",
             "image_format",
+            "server_metrics_url",
             "goodput",
+            "server_metrics_url",
```
debermudez, line 106: Not your code but can you re alphabetize this section?

genai-perf/tests/test_cli.py
```diff
@@ -917,3 +917,75 @@ def test_get_extra_inputs_as_dict(self, extra_inputs_list, expected_dict):
...
+                    "triton",
+                    "--server-metrics-url",
+                    INVALID_URL,
+                ],
+                None,
+                INVALID_URL_ERROR_MESSAGE,
+            ),
+        ],
+    )
+    def test_server_metrics_url_for_triton(
```
debermudez, line 978: do we have any testing for properly formed urls that are not actually a reachable endpoint?

lkomali: Yes, there are tests in test_telemetry_data_collector.py that uses valid URLs. There's one test for success scenario and one test for all kinds of failure scenarios like 404, 403, 500, etc. These are already merged to main. 

This is the URL 
` TEST_SERVER_URL = "http://testserver:8080/metrics"`

This is the test
image

debermudez: perfect. thank you for catching me up here.

genai-perf/tests/test_cli.py
```diff
@@ -917,3 +917,75 @@ def test_get_extra_inputs_as_dict(self, extra_inputs_list, expected_dict):
         namespace.extra_inputs = extra_inputs_list
         actual_dict = parser.get_extra_inputs_as_dict(namespace)
         assert actual_dict == expected_dict
+
+    TEST_TRITON_METRICS_URL = "http://custom-metrics-url:8002/metrics"
```
dyastremsky, line 921: Is there a reason for creating these global variables? I feel like this could be confusing. I'd need to scroll down quite a bit to see why they are used. It'd be better to have these as local variables inside the one test.

genai-perf/tests/test_cli.py
```diff
@@ -917,3 +917,75 @@ def test_get_extra_inputs_as_dict(self, extra_inputs_list, expected_dict):
...
+                INVALID_URL_ERROR_MESSAGE,
+            ),
+        ],
+    )
+    def test_server_metrics_url_for_triton(
+        self, args_list, expected_url, expected_error, monkeypatch, capsys
+    ):
+        monkeypatch.setattr("sys.argv", args_list)
+
+        if expected_error:
```
dyastremsky, line 983: When you start using branching logic like this, it suggests this should be two tests. Do you want to add the valid test to the test_cli valid arg test and then have a separate one just for the error testing? That assumes there's not a generic error test that you can reuse... which if we don't have, we should refactor sometime soon to add.

dyastremsky (commented):
Great work! A couple of comments.

genai-perf/tests/test_cli.py
lkomali: we have test_conditional_errors, that handles error scenarios. 
We can use it for the invalid url case, I think

debermudez (approved):

PR #40, opened 2024-08-16
Support image retrieval endpoint with OpenAI chat endpoint

nv-hwoo:
Some main changes in the PR are...
1. Add `image_retrieval` endpoint
2. Use openai chat api request format for image_retrieval endpoint
3. Refactor openai chat request input generation

### DIFF
--- a/genai-perf/genai_perf/llm_inputs/llm_inputs.py
+++ b/genai-perf/genai_perf/llm_inputs/llm_inputs.py
@@ -51,6 +51,7 @@ class OutputFormat(Enum):
     OPENAI_EMBEDDINGS = auto()
     OPENAI_VISION = auto()
     RANKINGS = auto()
+    IMAGE_RETRIEVAL = auto()
     TENSORRTLLM = auto()
     VLLM = auto()
     TENSORRTLLM_ENGINE = auto()
--- a/genai-perf/genai_perf/llm_inputs/llm_inputs.py
+++ b/genai-perf/genai_perf/llm_inputs/llm_inputs.py
@@ -322,7 +322,19 @@ class LlmInputs:
             input_file_dataset = cls._get_input_dataset_from_rankings_files(
                 queries_filename, passages_filename, batch_size, num_of_output_prompts
             )
-
+            generic_dataset_json = (
+                cls._convert_input_synthetic_or_file_dataset_to_generic_json(
+                    input_file_dataset
+                )
+            )
+        elif output_format == OutputFormat.IMAGE_RETRIEVAL:
+            if input_type != PromptSource.FILE:
+                raise GenAIPerfException(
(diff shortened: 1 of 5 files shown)


genai-perf/genai_perf/llm_inputs/llm_inputs.py
```diff
@@ -783,41 +771,6 @@ def _convert_generic_json_to_output_format(
...
-        output_tokens_deterministic: bool,
-        model_name: list = [],
-        model_selection_strategy: ModelSelectionStrategy = ModelSelectionStrategy.ROUND_ROBIN,
-    ) -> Dict:
-        # TODO (TMA-1757): Implement a way to select a role for `text_input`
-        (
-            system_role_headers,
-            user_role_headers,
-            _,
-        ) = cls._determine_json_feature_roles(dataset_json)
```
nv-hwoo, line 804: ### WHY
Removed this from chat completions because genai-perf _only_ supports user role in the openai chat endpoint anyway. There's no way for users to set other roles such as system or assistant, and if we need to support other roles, then we likely need some abstraction around openai chat message to make it more easy to maintain and work with.

### FOLLOW UP WORK
I only removed this for chat completions endpoint in this PR, but other code needs to be cleaned as well because it makes no sense for other endpoints to have this (since this is chat completions specific code). For instance, completions endpoint does not support system, user, assistant roles anyway. Same for Trition endpoints but since we were moving too fast and was just copying the same code, the code lives everywhere.

genai-perf/genai_perf/llm_inputs/llm_inputs.py
```diff
@@ -783,41 +771,6 @@ def _convert_generic_json_to_output_format(
 
         return output_json
 
-    @classmethod
-    def _convert_generic_json_to_openai_chat_completions_format(
```
nv-hwoo, line 787: With the removal of `_determine_json_feature_roles`, this method can be just merged with `_populate_openai_chat_completions_output_json`

genai-perf/genai_perf/llm_inputs/llm_inputs.py
```diff
@@ -637,37 +645,18 @@ def _get_prompts_from_input_file(
...
+                    images.append(image.strip() if image else image)
         return prompts, images
 
     @classmethod
     def verify_file(cls, input_filename: Path) -> None:
         if not input_filename.exists():
             raise FileNotFoundError(f"The file '{input_filename}' does not exist.")
 
-    @classmethod
-    def _convert_to_openai_multi_modal_content(
```
nv-hwoo, line 650: Moved as a new method `_extract_chat_contents` that's used when converting from generic json to chat format. Since this is endpoint specific logic, this doesn't need to be here.

debermudez (approved):

PR #42, opened 2024-08-20
Update console to move metrics into the table

debermudez:
Updating the console to move all metrics into the table. Tests were also updated.

Pre:
Screenshot 2024-08-19 at 16 11 00

Post:
Screenshot 2024-08-21 at 12 14 08

### DIFF
--- a/genai-perf/genai_perf/export_data/console_exporter.py
+++ b/genai-perf/genai_perf/export_data/console_exporter.py
@@ -83,7 +83,7 @@ class ConsoleExporter:
 
         for metric in self._metrics.system_metrics:
             metric_str = metric.name.replace("_", " ").capitalize()
-            metric_str = metric_str.replace("throughput", "tput")
+            # metric_str = metric_str.replace("throughput", "tput")
             metric_str += f" ({metric.unit})" if metric.unit != "tokens" else ""
             row_values = [metric_str]
             for stat in self.STAT_COLUMN_KEYS:
--- a/genai-perf/genai_perf/export_data/console_exporter.py
+++ b/genai-perf/genai_perf/export_data/console_exporter.py
@@ -43,14 +43,16 @@ class ConsoleExporter:
         self._args = config.args
 
     def _get_title(self):
+        title = "NVIDIA GenAI-Perf | "
         if self._args.endpoint_type == "embeddings":
-            return "Embeddings Metrics"
+            title += "Embeddings Metrics"
         elif self._args.endpoint_type == "rankings":
-            return "Rankings Metrics"
+            title += "Rankings Metrics"
         elif self._args.endpoint_type == "image_retrieval":
-            return "Image Retrieval Metrics"
+            title += "Image Retrieval Metrics"
         else:
-            return "LLM Metrics"
+            title += "LLM Metrics"
+        return title
 
     def export(self) -> None:
         table = Table(title=self._get_title())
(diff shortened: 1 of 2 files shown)


debermudez:
> I remember you mentioned there are some customers who are directly parsing the console output and so we need to confirm with them before making any changes to the console. Do you know who they are and should we run through them before making this change?

They were visually parsing it for a quick visual verification. 
Any automated parsing via scripts is using the csv or json. This is why I didnt update the csv in this quick PR.

genai-perf/genai_perf/export_data/console_exporter.py
```diff
@@ -84,6 +79,19 @@ def _construct_table(self, table: Table) -> None:
 
             table.add_row(*row_values)
 
+        for metric in self._metrics.system_metrics:
+            metric_str = metric.name.replace("_", " ").capitalize()
+            metric_str = metric_str.replace("throughput", "tput")
```
nv-hwoo, line 84: why are we replacing throughput to tput?

debermudez: The console gets too wide otherwise and risks cutting off the data.
The data seemed more important than the non abbreviated metric text.

nv-hwoo: I agree, but I feel like this needs a separate discussion because then we're losing consistency in our visual display. Since console is solely for visual look, and people who need the data can look for csv or json, could we keep the full name for consistency for now? If this becomes a problem, then we can bring the discussion to the table along with other metrics as well.

genai-perf/genai_perf/export_data/console_exporter.py
matthewkotila, line 1: @debermudez thoughts on perhaps another column explaining the set over which the particular statistic is aggregated? For instance, the last two are one value per benchmark whereas others might be one value per request.

genai-perf/genai_perf/export_data/console_exporter.py
debermudez: I am confused by how we are losing consistency in our visual display?
This is the only place where it is displayed.

genai-perf/genai_perf/export_data/console_exporter.py
debermudez: I thought the N/A effectively conveyed that information. 
Do you think it would be more clear if the data was replicated in all of the columns?

genai-perf/genai_perf/export_data/console_exporter.py
nv-hwoo: what I mean is no other metrics are abbreviated except the two added metrics.

debermudez: I see. At the moment, we have sign off from product who I think would be the final call on this. Still happy to discuss offline if that is not sufficient for your concerns.

nv-hwoo: I'm okay with change if product decides to do so. Just wanted to discuss about this before we change anything on screen that users see directly.

Could you update the screenshot above with the `tput` output? 

cc @harryskim for final call

genai-perf/genai_perf/export_data/console_exporter.py
matthewkotila: Discussed offline, no strong opinion one way or the other. Just wanted to suggest it to see if anyone else thought it was a good idea.

genai-perf/genai_perf/export_data/console_exporter.py
debermudez: Updated the screenshot

debermudez: reverted the abbreviation. should be all set now

nv-hwoo (dismissed):
LGTM! Thanks for making this change 🚀

nv-hwoo (approved):

PR #47, opened 2024-08-21
Graceful handling of exception thrown by genai-perf if metrics url is unreachable 

lkomali:
This is a small PR to gracefully handle the exception thrown by GenAI-Perf and it is logged as a warning that the url is not reachable.

image

### DIFF
--- a/genai-perf/genai_perf/telemetry_data/telemetry_data_collector.py
+++ b/genai-perf/genai_perf/telemetry_data/telemetry_data_collector.py
@@ -47,11 +47,11 @@ class TelemetryDataCollector(ABC):
         self._thread: Optional[Thread] = None
 
     def is_url_reachable(self) -> bool:
-        TIMEOUT_SECONDS = 5
+        timeout_seconds = 5
         if self._server_metrics_url:
             try:
                 response = requests.get(
-                    self._server_metrics_url, timeout=TIMEOUT_SECONDS
+                    self._server_metrics_url, timeout=timeout_seconds
                 )
                 return response.status_code == requests.codes.ok
             except requests.RequestException:
@@ -59,20 +59,17 @@ class TelemetryDataCollector(ABC):
         return True
 
     def start(self) -> None:
-        """Start the telemetry data collection thread."""
         if self._thread is None or not self._thread.is_alive():
             self._stop_event.clear()
             self._thread = Thread(target=self._collect_metrics)
             self._thread.start()
 
     def stop(self) -> None:
-        """Stop the telemetry data collection thread."""
         if self._thread is not None and self._thread.is_alive():
             self._stop_event.set()
             self._thread.join()
 
     def _fetch_metrics(self) -> str:
-        """Fetch metrics from the metrics endpoint"""
         response = requests.get(self._server_metrics_url)
         response.raise_for_status()
         return response.text
@@ -83,7 +80,7 @@ class TelemetryDataCollector(ABC):
         pass
 
     def _collect_metrics(self) -> None:
-        """Continuously collect telemetry metrics at for every second"""
+        """Continuously collect telemetry metrics for every second"""
         while not self._stop_event.is_set():
             metrics_data = self._fetch_metrics()
             self._process_and_update_metrics(metrics_data)
@@ -91,10 +88,8 @@ class TelemetryDataCollector(ABC):
 
     @property
     def metrics(self) -> TelemetryMetrics:
-        """Return the collected metrics."""
         return self._metrics
 
     @property
     def metrics_url(self) -> str:
-        """Return server metrics url"""
         return self._server_metrics_url
--- a/genai-perf/genai_perf/telemetry_data/telemetry_data_collector.py
+++ b/genai-perf/genai_perf/telemetry_data/telemetry_data_collector.py
@@ -46,11 +46,13 @@ class TelemetryDataCollector(ABC):
         self._stop_event = Event()
         self._thread: Optional[Thread] = None
 
-    def check_url_reachability(self) -> bool:
-        """Check if the server metrics URL is reachable"""
+    def is_url_reachable(self) -> bool:
+        TIMEOUT_SECONDS = 5
         if self._server_metrics_url:
             try:
-                response = requests.get(self._server_metrics_url, timeout=5)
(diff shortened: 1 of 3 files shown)


genai-perf/genai_perf/telemetry_data/telemetry_data_collector.py
```diff
@@ -46,6 +46,16 @@ def __init__(
         self._stop_event = Event()
         self._thread: Optional[Thread] = None
 
+    def check_url_reachability(self) -> bool:
```
dyastremsky, line 49:
```suggestion
    def is_url_reachable(self) -> bool:
```

genai-perf/genai_perf/telemetry_data/telemetry_data_collector.py
```diff
@@ -46,6 +46,16 @@ def __init__(
...
+        """Check if the server metrics URL is reachable"""
+        if self._server_metrics_url:
+            try:
+                response = requests.get(self._server_metrics_url, timeout=5)
```
dyastremsky, line 53: Set the timeout value to a separate variable that clarifies the units. We want to avoid magic numbers.

genai-perf/genai_perf/telemetry_data/telemetry_data_collector.py
```diff
@@ -81,3 +91,8 @@ def _collect_metrics(self) -> None:
     def metrics(self) -> TelemetryMetrics:
         """Return the collected metrics."""
         return self._metrics
+
+    @property
+    def metrics_url(self) -> str:
+        """Return server metrics url"""
```
dyastremsky, line 97: Do we want comments for every function? Some of these might be redundant.

genai-perf/genai_perf/wrapper.py
```diff
@@ -152,7 +152,12 @@ def run(
     ) -> None:
         try:
             if telemetry_data_collector is not None:
-                telemetry_data_collector.start()
+                if telemetry_data_collector.check_url_reachability():
+                    telemetry_data_collector.start()
+                else:
+                    logger.warning(
+                        f"The metrics url ({telemetry_data_collector.metrics_url}) is unreachable, cannot collect telemetry data"
```
dyastremsky, line 159: A little too long. Also a run-on sentence. We might want to make it shorter.
Something like:
```
The metrics url ({telemetry_data_collector.metrics_url}) is unreachable.
GenAI-Perf cannot collect telemetry data.
```

genai-perf/tests/test_telemetry_data_collector.py
```diff
@@ -143,3 +143,44 @@ def test_collect_metrics(
...
+        collector: MockTelemetryDataCollector,
+    ) -> None:
+        mock_get.return_value.status_code = requests.codes.ok  # 200
+        assert collector.check_url_reachability() is True
+
+    @patch("requests.get")
+    def test_url_reachability_check_failure(
+        self, mock_get: MagicMock, collector: MockTelemetryDataCollector
+    ) -> None:
+        # Simulate a 404 Not Found error
```
dyastremsky, line 160: Are the number important here? I feel like we could remove all of the comments, as the request codes names are clear and have associated documentation.

dyastremsky (commented):
Fantastic work! I left some comments.

genai-perf/genai_perf/telemetry_data/telemetry_data_collector.py
dyastremsky: With boolean-returning functions, naming like this is helpful to make it clear that the function returns a boolean and what it is.

genai-perf/genai_perf/telemetry_data/telemetry_data_collector.py
```diff
@@ -46,6 +46,18 @@ def __init__(
         self._stop_event = Event()
         self._thread: Optional[Thread] = None
 
+    def is_url_reachable(self) -> bool:
+        TIMEOUT_SECONDS = 5
```
dyastremsky, line 50: Should this be all caps? I think you only do this with module-level constants, not function variables.

lkomali: yep, sorry i moved it to the function but didn't make it lowercase.

genai-perf/tests/test_telemetry_data_collector.py
lkomali: Missed it sorry

genai-perf/genai_perf/wrapper.py
```diff
@@ -152,7 +152,13 @@ def run(
     ) -> None:
         try:
             if telemetry_data_collector is not None:
-                telemetry_data_collector.start()
+                if telemetry_data_collector.is_url_reachable():
+                    telemetry_data_collector.start()
+                else:
+                    logger.warning(
+                        f"The metrics URL ({telemetry_data_collector.metrics_url}) is unreachable."
```
dyastremsky, line 159: Nit: Could you add a space at the end of this? Otherwise, there won't be one.

E.g. 
```
f"The metrics URL ({telemetry_data_collector.metrics_url}) is unreachable. "
"GenAI-Perf cannot collect telemetry data."
```
This makes it so that instead of "...unreachable.GenAI-Perf...", the message reads "...unreachable. GenAI-Perf..."

dyastremsky (commented):
One nit, then everything looks good from my end.

Have you run it and received the error message? Does the error message look good? If so, it'd be nice to add a screenshot to this PR but is not mandatory.

genai-perf/genai_perf/wrapper.py
lkomali: Added the space.

lkomali:
Attached the screenshot of how the error message looks on the console.

genai-perf/genai_perf/wrapper.py
dyastremsky: Did you push it? I don't see it on my end.

lkomali: yes, I did. Let me know if you still cannot see the change. There was some problem initially when I tried to update my branch and this change was overwritten I think.

dyastremsky (approved):

PR #49, opened 2024-08-22
Add doc generation script with doc templates

debermudez:
Trying to centralize the doc maintenance.

Now there is a perf_analyzer/templates directory. In that directory exists template_vars.yaml, generate_docs.py, and a directory of genai-perf jinja2 templates for the docs.

All of the templates are in normal markdown except where a release number or variable that needs to be updated exists. When you run generate_docs.py in the templates directory, the script will iterate over the files in the template_vars.yaml file, populating all fields in the template with the respective value in the yaml file. 
Finally, the generated docs are written out to their current location to avoid manual copying. 

Now, instead of scouring the codebase for release values to update every month, we can update 1 file and regenerate the docs. Only the docs that change need to be committed.

The next step is to expand this to perf_analyzer's docs where this is even more helpful.

### DIFF
--- a/templates/genai-perf-templates/README_template
+++ b/templates/genai-perf-templates/README_template
@@ -1,0 +1,567 @@
+<!--
+Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+ * Redistributions of source code must retain the above copyright
+   notice, this list of conditions and the following disclaimer.
+ * Redistributions in binary form must reproduce the above copyright
+   notice, this list of conditions and the following disclaimer in the
+   documentation and/or other materials provided with the distribution.
+ * Neither the name of NVIDIA CORPORATION nor the names of its
+   contributors may be used to endorse or promote products derived
+   from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+-->
+
+# GenAI-Perf
+
+GenAI-Perf is a command line tool for measuring the throughput and latency of
+generative AI models as served through an inference server.
+For large language models (LLMs), GenAI-Perf provides metrics such as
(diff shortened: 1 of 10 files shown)


templates/genai-perf-templates/compare_template
matthewkotila, line 1: Is the intended workflow that once per release, someone manually runs generate_docs.py and then it updates the actual md files in the repo and then they commit those changes?

templates/genai-perf-templates/README_template
matthewkotila, line 1: Is the intended workflow that once per release, someone manually runs generate_docs.py and then it updates the actual md files in the repo and then they commit those changes?

debermudez: yes. that is the intention for now.
I can see a future where we automate this in the CI further. For now, this should centralize a lot of the update work.

templates/template_vars.yaml
```diff
@@ -0,0 +1,43 @@
+README:
+  filename: README.md
+  template: genai-perf-templates/README_template
+  release: 24.08
```
dyastremsky, line 4: Great work!

Is there a way to centralize the variables here? It seems like this moves us from needing to find the variables to update to being able to update them in one file then run the script. The next step for me would be to just having one centralized list of variables (maybe in this file?) to then know we need to update each one once. We'd go to the top of the and update release version, CLI version, etc. once and be done.

debermudez: This was a conscientious decision on my part to provide flexibility for every file while still centralizing all variables.
I can see going forward making a central list and then only overwrite if it is specified for a file. However, I did not want to over engineer this right now when ops is likely to take this completely out of this repo.
I wanted a short term solution that had room to evolve for later.

For now, search and replace in this one file is still very straight forward.

fpetrini15: Yeah I had the same suggestion. I had your branch pulled and was working on it when the suggestion came through. Seems like it's not as straight-forward as I'd hoped, though I'm no YML expert. I think we can extend this in the future.

templates/generate_docs.py
```diff
@@ -0,0 +1,55 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+import os
+from pathlib import Path
+
+import yaml
+from jinja2 import Environment, FileSystemLoader
+
```
fpetrini15, line 35: Might be nice to be able to run this script from top-level. Pulling inspiration from Triton's documentation generation script, we could add:

```python3
# Global constants
PA_ABSLT_PATH = os.environ.get("PA_ABSLT_PATH", os.getcwd())
PA_TEMPLATES_ABSLT_PATH = os.path.join(PA_ABSLT_PATH, "templates")

# Change working directory to perf_analyzer/templates.
os.chdir(PA_TEMPLATES_ABSLT_PATH)
.
.
# Render logic
.
.
# Change working directory to perf_analyzer.
os.chdir(PA_ABSLT_PATH)
```

fpetrini15:
Great work @debermudez! In my testing, many of the files are unchanged because there is no version to modify. I'm assuming you expected this and are just generating templates in advance? Would it be better to only add template files as needed to reduce bloat?

debermudez:
> Great work @debermudez! In my testing, many of the files are unchanged because there is no version to modify. I'm assuming you expected this and are just generating templates in advance? Would it be better to only add template files as needed to reduce bloat?

That is expected behavior yes. Most of the docs are not going to need updates.
My intention here was to set up templates to make it easy to update in the future without much work. If you would prefer I only template the files that need updating now, i can remove the ones that are not changing.

templates/generate_docs.py
debermudez: Added this. Now we can run the script from the root directory.

templates/template_vars.yaml
debermudez: Updated the script to centralize the variables to  a single general set.
Now each file may specialize a variable which will take precedence over the general variable if both are defined.
Thank you for the suggestion, @dyastremsky  and @fpetrini15!

fpetrini15 (approved):
Thanks again for taking this on! 🚀

templates/genai-perf-templates/README_template
matthewkotila: Whoops, didn't realize this was gonna get merged so soon.

Thoughts on automating the filling of the template to happen during the pre-commit? It could even be set up to squash the change into the latest commit so it's "transparent".

Also, thoughts on keeping the version completely separate in a `version.txt` in the root of the repo (or `genai-perf` directory), similar to PyTorch: https://github.com/pytorch/pytorch/blob/main/version.txt

Then pulling from that.

PR #55, opened 2024-08-26
Consolidate input arguments

debermudez:
This pull request changes Inputs to use a config instead of passing around long lists of arguments across class functions. It includes the following actions:
- Creates a new InputsConfig data class to carry options around the Inputs functions.
- Updates methods and testing to use the new InputsConfig.
- Moves from using static calls to `Inputs.<method>` to instead instantiate an Inputs object that has a public create_inputs function, which is handled with the internal private functions. This will help with scalability in the future.
- Removes add_model_name, since it is never false.


genai-perf/genai_perf/parser.py
```diff
@@ -35,18 +35,11 @@
...
     DEFAULT_TRITON_METRICS_URL,
-    OPEN_ORCA,
-)
-from genai_perf.inputs.inputs import (
-    Inputs,
-    ModelSelectionStrategy,
-    OutputFormat,
-    PromptSource,
 )
+from genai_perf.inputs import input_constants as ic
```
nv-braf, line 42: Is there a reason for the inconsistency in how we import constants? This is different than in inputs.py

genai-perf/genai_perf/inputs/inputs_config.py
```diff
@@ -53,89 +53,112 @@
...
-      num_prompts:
-          The number of synthetic output prompts to generate
-      random_seed:
-          Seed used to generate random values
+    ####################
+    # General Parameters
+    ####################
 
-    """
+    # add_stream:
```
nv-braf, line 62: This is definitely an improvement! But, it seems like listing the variable name above the comment is redundant and could be removed while keeping the same level of readability.

debermudez: You are totally right. 
Got rid of the redundancy.

nv-braf (approved):

PR #56, opened 2024-08-27
create retriever factory

debermudez:


genai-perf/genai_perf/inputs/inputs.py
```diff
@@ -59,360 +48,28 @@ def create_inputs(self) -> Dict:
...
-        with open(self.config.input_filename, mode="r", newline=None) as file:
-            for line in file:
-                if line.strip():
-                    # None if not provided
-                    prompt = load_json_str(line).get("text_input")
-                    image = load_json_str(line).get("image")
-                    prompts.append(prompt.strip() if prompt else prompt)
-                    images.append(image.strip() if image else image)
-        return prompts, images
+        self._check_for_dataset_name_if_input_type_is_url()
```
dyastremsky, line 62: Why was the try/catch block removed?

genai-perf/tests/test_input_embeddings.py
```diff
@@ -55,9 +59,9 @@ def test_get_input_dataset_from_embeddings_file(self, mock_file, mock_exists):
             batch_size=batch_size,
             num_prompts=100,
         )
-        inputs = Inputs(config)
+        factory = InputRetrieverFactory(config)
```
dyastremsky, line 62: Could you use a different variable name? Factory is too generic.

genai-perf/tests/test_input_rankings.py
```diff
@@ -76,8 +77,8 @@ def test_get_input_dataset_from_rankings_file(self, mock_file):
             batch_size=batch_size,
             num_prompts=100,
         )
-        inputs = Inputs(config)
-        dataset = inputs._get_input_dataset_from_rankings_files(
+        irf = InputRetrieverFactory(config)
```
dyastremsky, line 80: Could you use a different variable name and keep it consistent with the other test file? Irf is unintuitive, especially if you see it later in the code.

dyastremsky (commented):
Fantastic work! A few small comments.

genai-perf/tests/test_input_embeddings.py
debermudez: I have often seen short tests use generic names.
Ill update this and avoid that practice in the future.

Updated to `input_retriever_factory`

genai-perf/tests/test_input_rankings.py
debermudez: Updated to `input_retriever_factory`

genai-perf/tests/test_input_embeddings.py
dyastremsky: Thanks! Could you do this for all of the tests? I think test_input_retriever_factory is the last one.

I think people used to prioritize conciseness more, but readability is always important, especially if you might use the variable more than 1-2 lines below its declaration. However, with code completion tools now widely available, the emphasis is shifting (at least in my opinion) to be readability >>> conciseness.

debermudez: done.

I agree with you. 
I didn't think this was really unreadable but if you are saying it was harder to follow, that is justification enough to change. Thanks for calling it out.

dyastremsky (approved):
Excellent work!

PR #58, opened 2024-08-27
Add goodput support for image retrieval endpoint

AndyDai-nv:
Add goodput support for image retrieval endpoint, supported SLOs include: image latency, request latency and image throughput.
see examples below:

```bash 
genai-perf profile \
        -m nv-yolox-structured-images-v1 \ 
        --service-kind openai \
        --endpoint-type image_retrieval \
        --input-file images.jsonl  \
        --goodput image_latency:28
```

Example output

image


genai-perf/genai_perf/goodput_calculator/llm_goodput_calculator.py
```diff
@@ -45,7 +46,7 @@ class LLMGoodputCalculator(GoodputCalculator):
     def __init__(
         self,
         goodput_constraints: Dict[str, float],
-        metric: Union[LLMMetrics, Metrics],
+        metric: Union[LLMMetrics, ImageRetrievalMetrics, Metrics],
```
nv-hwoo, line 49: can we set this to base class?
```suggestion
        metric: Metrics,
```

genai-perf/genai_perf/profile_data_parser/image_retrieval_profile_data_parser.py
```diff
@@ -76,9 +82,26 @@ def _parse_requests(self, requests: dict) -> ImageRetrievalMetrics:
...
+            llm_goodput_calculator = LLMGoodputCalculator(
+                self._goodput_constraints,
+                self._image_metric,
+                benchmark_duration,
+            )
+
+            llm_goodput_calculator.compute()
+            self._image_metric.request_goodputs = llm_goodput_calculator.goodput
+        else:
+            return
```
nv-hwoo, line 107: all `ProfileDataParser`, `LLMProfileDataParser`, and `ImageRetrievalProfileDataParser` shares this identical code block. let's only implement this in the base class and let all the derived classes use that one instead of having their own copy.

genai-perf/tests/test_image_retrieval_profile_data_parser.py
```diff
@@ -138,3 +139,71 @@ def test_image_retrieval_profile_data(
...
+        infer_mode,
+        load_level,
+        expected_metrics,
+    ) -> None:
+        """Collect image retrieval metrics from profile export data and check values.
+
+        Goodput Constraints
+        * image latency: 1.25e-6 ms
+        * request latency: 2.5e-6 ms
+
```
nv-hwoo, line 178: We don't need to reiterate this information here as it's already in the code.

```suggestion
```

dyastremsky:
LGTM. Question for @AndyDai-nv and maybe @nv-hwoo, with the modularization/refactor, is there a way to get it so that goodput is automatically added for new APIs? It looks like the code is sprawling a bit now with new metrics and profile export classes, so I'm wondering if we'll be able to make it so that new APIs don't need much change to each new API to support metrics including goodput.

dyastremsky (approved):

genai-perf/tests/test_image_retrieval_profile_data_parser.py
```diff
@@ -138,3 +139,71 @@ def test_image_retrieval_profile_data(
...
+            - [2 / (3 - 1), 3 / (7 - 3)] = [1 / ns_to_sec(1), 3 / ns_to_sec(4)]
+        * image latencies
+            - [(3 - 1) / 2, (7 - 3) / 3] = [1, 4/3]
+        * request goodput
+            - [1 / (7 - 1)] = [1 / ns_to_sec(6)]
+
+        """
+        test_goodput_constraints = {
+            "image_latency": 1.25e-6,  # ms
+            "request_latency": 2.5e-6,  # ms
```
nv-hwoo, line 194: is this msec? or sec?

genai-perf/tests/test_image_retrieval_profile_data_parser.py
```diff
@@ -43,6 +43,7 @@ def check_image_retrieval_metrics(
     assert m1.request_throughputs == pytest.approx(m2.request_throughputs)
     assert m1.image_latencies == m2.image_latencies
     assert m1.image_throughputs == pytest.approx(m2.image_throughputs)
+    assert m1.request_goodputs is None or m1.request_goodputs == m2.request_goodputs
```
dyastremsky, line 46: Why is there an or here? Shouldn't only one assertion be true?

genai-perf/genai_perf/profile_data_parser/image_retrieval_profile_data_parser.py
nv-hwoo: @dyastremsky making this change, I believe, will address your question.

genai-perf/tests/test_image_retrieval_profile_data_parser.py
AndyDai-nv: msec, the testing data is in ns unit.
`1.25e-6` will further be 1.25ns to be able to compare with `[1, 4/3]`

genai-perf/tests/test_image_retrieval_profile_data_parser.py
AndyDai-nv: This is for existing tests with no goodput computation. I will try to find a way to get rid of this

genai-perf/tests/test_image_retrieval_profile_data_parser.py
AndyDai-nv: Wouldn't this be more readable? As you can see clearly in the comment to see which request satisfy the constraints and hence quickly understand the process of computing goodput.

nv-hwoo: I don't think this is about readability but more about are you giving any new information through docstring. For example, the docstring about metrics is useful because it shows you how the numbers are derived, instead of just reiterating the number. If you would like to make it more clear, then I suggest you add more information (e.g. how you are using goodput constraints to arrive at the final metric, etc).

AndyDai-nv: Updated, add computation process in the comment

genai-perf/genai_perf/profile_data_parser/image_retrieval_profile_data_parser.py
AndyDai-nv: @dyastremsky @nv-hwoo See below whether the new iteration addresses your question. 
Updated according to all suggestions in this round.

genai-perf/genai_perf/profile_data_parser/image_retrieval_profile_data_parser.py
```diff
@@ -76,9 +81,17 @@ def _parse_requests(self, requests: dict) -> ImageRetrievalMetrics:
         benchmark_duration = (max_res_timestamp - min_req_timestamp) / 1e9  # to seconds
         request_throughputs = [len(requests) / benchmark_duration]
 
-        return ImageRetrievalMetrics(
+        self._image_metric = ImageRetrievalMetrics(
```
nv-hwoo, line 84: is there a reason why this needs to be an instance attribute?

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
```diff
@@ -163,22 +162,11 @@ def _parse_requests(self, requests: dict) -> Metrics:
...
-                benchmark_duration,
-            )
+            goodput_val = self._calculate_goodput(benchmark_duration, self._llm_metric)
+            self._llm_metric.request_goodputs = goodput_val
 
-            llm_goodput_calculator.compute()
-            self._llm_metric.request_goodputs = llm_goodput_calculator.goodput
-        else:
-            return
+        return self._llm_metric
```
nv-hwoo, line 169: same as above

genai-perf/tests/test_image_retrieval_profile_data_parser.py
nv-hwoo: nice! thanks for adding the details

nv-hwoo (commented):
few nit comments but otherwise, lgtm!

genai-perf/genai_perf/profile_data_parser/image_retrieval_profile_data_parser.py
AndyDai-nv: No, nice catch. 
Updated in the new version.

nv-hwoo (approved):
awesome. well done! 👍

nv-hwoo:
@AndyDai-nv please make sure that CI passed before merging. It seems like some jobs failed (which I think was due to a system hiccup, so I would respin).

dyastremsky (approved):

PR #59, opened 2024-08-27
Create converter factory for inputs generation

dyastremsky:
Pull out the logic that converts the generic JSON to the API-specific logic into its own factory class. Note that is a refactor meant to ease the next steps of refactoring inputs. It is not meant to fix anything beyond pulling the logic out of inputs into its own file.

Next steps may include cleaning up the conversion logic and breaking each converter into its own file in a converter directory.


genai-perf/genai_perf/inputs/output_format_converter.py
```diff
@@ -0,0 +1,780 @@
...
+        if (
+            header in system_role_headers
+            or header in user_role_headers
+            or header in text_input_headers
+        ):
+            new_prompt = content
+
+        return new_prompt
+
+    def _create_new_text_input(
```
nv-braf, line 296: If this class is being derived from BaseConverter and the methods are identical why aren't we just using the base converter class' method?
Also, it seems like _create_new_prompt and _create_new_text_input are identical. Maybe it would be better to create a generic method in BaseConverter and then just call this for both methods?

genai-perf/genai_perf/inputs/output_format_converter.py
```diff
@@ -0,0 +1,780 @@
...
+
+            self._add_optional_tags_to_openai_json(
+                pa_json["data"][index],
+                config,
+                iter_model_name,
+            )
+
+        return pa_json
+
+    def _create_new_prompt(
```
nv-braf, line 395: Same comment as previous - should avoid duplication of code

nv-braf (commented):
Looks good overall. My general feedback would be that I would like to see output_format_converter broken into separate files for each class (maybe that's a future story?)

genai-perf/genai_perf/inputs/output_format_converter.py
dyastremsky: Thanks for catching this! I fixed this. I was trying to use the lightest touch when pulling these out of the original file and missed this duplication.

genai-perf/genai_perf/inputs/output_format_converter.py
dyastremsky: Done!

dyastremsky:
> Looks good overall. My general feedback would be that I would like to see output_format_converter broken into separate files for each class (maybe that's a future story?)

Thanks so much for reviewing and catching the duplication! I 100% agree with that direction. This ticket was just focused on pulling out the code into its own file. There is an open ticket (TPA-365) for breaking these files into their own classes.

nv-braf (approved):

PR #64, opened 2024-08-29
Separate output format converters into their own files for better organization

dyastremsky:
Note: While this is a bigger PR, all it is really doing is splitting the classes into their own functions. The main things to review are the new directory structure, import technique (see `__init__.py`), and overall design. It may also be useful to skim base_converter.py and output_format_converter_factory.py.


nv-braf (approved):
This is great! Exactly what I was expecting to see.

dyastremsky:
Thanks, @nv-braf! You're so fast. ⏩ 

Would you be able to re-approve when you get a chance? I was writing up the description for you and realized the output format converter factory class needed to be renamed to have factory in the name.

genai-perf/genai_perf/inputs/converters/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+
+from .openai_chat_completions_converter import OpenAIChatCompletionsConverter
+from .openai_completions_converter import OpenAICompletionsConverter
+from .openai_embeddings_converter import OpenAIEmbeddingsConverter
+from .rankings_converter import RankingsConverter
+from .tensorrtllm_converter import TensorRTLLMConverter
+from .tensorrtllm_engine_converter import TensorRTLLMEngineConverter
+from .vllm_converter import VLLMConverter
+
+__all__ = [
```
debermudez, line 35: As this is currently defined, this basically redefines the imports above?
This might be a pain to maintain in the future, but I am not sure.
This is not blocking tho.

genai-perf/genai_perf/inputs/output_format_converter_factory.py
```diff
@@ -0,0 +1,53 @@
...
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from genai_perf.exceptions import GenAIPerfException
+from genai_perf.inputs.converters import *
```
debermudez, line 28: I have always been cautioned against wildcard imports but I really have a good sense why I should avoid them.
Can you explain why you used a wildcard here?

debermudez:
One question and one comment.
Overall, a very nice addition to the rest of the hard work you did this week on this. Thank you!

nv-braf (approved):

genai-perf/genai_perf/inputs/converters/__init__.py
dyastremsky: Responded to your other comment.

This just makes it so that imports are explicitly defined, which will also make code analyzers happier since it's better practice. It also means someone checking this file knows exactly what's imported. It does mean needing to add the class name into the file in two places, that's true. I'm open to other approaches. I don't have a better one now, so unless you do, we could also iterate on this if we find a better approach.

genai-perf/genai_perf/inputs/output_format_converter_factory.py
debermudez: I have concerns around the traceability of the imports in other files later on.
BUT as you have said, we likely need to iterate on this as we go AND this is the best proposed solution we have so far. Nice work.

genai-perf/genai_perf/inputs/converters/__init__.py
debermudez: addressed in the other thread. 
this is good to go and we can always iterate if we must later.

debermudez (approved):

PR #70, opened 2024-09-03
Create Telemetry Statistics

lkomali:
PR to create statistics for telemetry metrics.


debermudez:
adding @nv-hwoo  for this review as the stats/metrics expert

genai-perf/genai_perf/metrics/telemetry_statistics.py
```diff
@@ -0,0 +1,112 @@
...
+    def _calculate_percentiles(
+        self, data: List[float], attr: str, gpu_index: str
+    ) -> None:
+        percentile_results = StatisticsUtil.calculate_percentiles(data)
+        p25 = percentile_results[0]
+        p50 = percentile_results[1]
+        p75 = percentile_results[2]
+        p90 = percentile_results[3]
+        p95 = percentile_results[4]
+        p99 = percentile_results[5]
```
nv-hwoo, line 77:
```suggestion
        p25, p50, p75, p90, p95, p99 = percentile_results
```

genai-perf/genai_perf/metrics/statistics.py
```diff
@@ -80,51 +80,50 @@ def _should_skip(self, data: List[Union[int, float]], attr: str) -> bool:
...
     def _calculate_percentiles(self, data: List[Union[int, float]], attr: str) -> None:
-        p25, p50, p75 = np.percentile(data, [25, 50, 75])
-        p90, p95, p99 = np.percentile(data, [90, 95, 99])
+        percentile_results = StatisticsUtil.calculate_percentiles(data)
+        p25 = percentile_results[0]
+        p50 = percentile_results[1]
+        p75 = percentile_results[2]
+        p90 = percentile_results[3]
+        p95 = percentile_results[4]
+        p99 = percentile_results[5]
```
nv-hwoo, line 94:
```suggestion
        p25, p50, p75, p90, p95, p99 = percentile_results
```

genai-perf/genai_perf/main.py
```diff
@@ -165,9 +170,9 @@ def run():
             create_artifacts_dirs(args)
             tokenizer = get_tokenizer(args.tokenizer)
             generate_inputs(config_options)
-            args.func(args, extra_args)
+            telemetry_data_collector = args.func(args, extra_args)
```
debermudez, line 173: In the profile subcommand case, this line means that the call to the profile wrapper is returning the telemetry data collector.  That probably should not be coupled that way.

genai-perf/genai_perf/metrics/statistics_util.py
```diff
@@ -0,0 +1,55 @@
...
+from typing import List, Tuple, Union
+
+import numpy as np
+
+
+class StatisticsUtil:
+
+    @staticmethod
+    def calculate_mean(values: List[Union[int, float]]) -> float:
+        return float(np.mean(values))
```
nv-hwoo, line 38: What is the benefit of taking these single line numpy functions out into static methods under a separate class? Before we were directly calling the numpy functions, and now we are calling static wrapper methods around the numpy functions.

genai-perf/genai_perf/metrics/statistics.py
```diff
@@ -80,51 +80,50 @@ def _should_skip(self, data: List[Union[int, float]], attr: str) -> bool:
...
-        avg = np.mean(data)
+        avg = StatisticsUtil.calculate_mean(data)
         setattr(self, "avg_" + attr, avg)
-        self._stats_dict[attr]["avg"] = float(avg)
+        self._stats_dict[attr]["avg"] = avg
 
     def _calculate_percentiles(self, data: List[Union[int, float]], attr: str) -> None:
-        p25, p50, p75 = np.percentile(data, [25, 50, 75])
-        p90, p95, p99 = np.percentile(data, [90, 95, 99])
+        percentile_results = StatisticsUtil.calculate_percentiles(data)
```
debermudez, line 88: Do we want a wrapper around numpy like this?
Can you provide some insight into this portion of the design?

genai-perf/genai_perf/metrics/telemetry_statistics.py
```diff
@@ -0,0 +1,112 @@
...
+        setattr(self, f"p25_{attr}_{gpu_index}", p25)
```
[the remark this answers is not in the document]
lkomali: I'm not sure where it's used. 
It was added in statistics.py and I assumed it might be used somewhere, and followed it.

genai-perf/genai_perf/metrics/telemetry_statistics.py
```diff
@@ -0,0 +1,112 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from collections import defaultdict
+from typing import Any, DefaultDict, Dict, List
+
+from genai_perf.metrics.statistics_util import StatisticsUtil
+from genai_perf.metrics.telemetry_metrics import TelemetryMetrics
+
+
+class TelemetryStatistics:
```
debermudez, line 36: Is there sufficient overlap here to create a base class between this class and the Metrics class?

genai-perf/genai_perf/parser.py
```diff
@@ -896,6 +897,8 @@ def profile_handler(args, extra_args) -> None:
         telemetry_data_collector=telemetry_data_collector,
     )
 
+    return telemetry_data_collector
```
debermudez, line 900: This doesnt feel like the right place for this.
I think we need to decouple the telemetry collector creation from the profile command.

genai-perf/genai_perf/main.py
```diff
@@ -130,8 +135,8 @@ def report_output(data_parser: ProfileDataParser, args: Namespace) -> None:
     else:
         raise GenAIPerfException("No valid infer mode specified")
 
-    # TPA-274 - Integrate telemetry metrics with other metrics for export
     stats = data_parser.get_statistics(infer_mode, load_level)
+    telemetry_data_collector.get_statistics()
```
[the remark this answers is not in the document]
lkomali: we will pass it to reporter which will be in the next PR.

genai-perf/genai_perf/metrics/statistics_util.py
lkomali: @debermudez the above is the reason why I did it. 
Open to suggestions.

genai-perf/genai_perf/metrics/telemetry_statistics.py
lkomali: Are you asking about TelemetryMetrics and Metrics class?

genai-perf/genai_perf/main.py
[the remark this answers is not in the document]
debermudez: I think this line goes in the next PR then.
We dont know when that will get in and this line will exist in main until that time.

genai-perf/genai_perf/metrics/telemetry_statistics.py
debermudez: yes

genai-perf/genai_perf/parser.py
lkomali: The reason to create telemetry_data_collector in profile command is because we need to collect while PA is running.
And the return is because telemetry_data_collector had metrics decoupled, and so it calls the telemetry_statistics on the metrics.
Something that I could think of it to move the creation and url_reachability to main (maybe in a function) and then call profile_handler by passing the telemetry_data_collector as an argument.
Let me know if that seems like a good way to go?

genai-perf/genai_perf/metrics/telemetry_statistics.py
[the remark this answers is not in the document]
nv-hwoo: yeah we started with storing the values as attributes, but then migrated to using `stats_dict` instead. I'm not sure if the attributes for `Statistics` class is still being used anywhere in the code, but `TelemetryStatistics` doesn't need to follow that unless it's using it.

genai-perf/genai_perf/metrics/statistics_util.py
debermudez: I defer to @nv-hwoo judgement here. I dont know if this is necessary now but i am not sure.

genai-perf/genai_perf/parser.py
debermudez: I like that approach more.

genai-perf/genai_perf/metrics/statistics_util.py
nv-hwoo: >According to how the metrics are stored, this is the common code between the two. The way the values should be stored is different and thus, this is the design I could come up with.

In this case, if the two codes require different handling of the metrics, reducing the common code down to just a single line of numpy function, then IMO it's not really worth taking it out. This only adds extra hierarchy to the function call without any benefit of utility functions (e.g. removing duplicate code). 

+When I asked if it would be possible to reuse the `Statistics` class for the new telemetry stats class, I was wondering if we could leverage the current code in the statistics class to do the same for telemetry because they are trying to calculate the same thing. If the two functionally don't overlap, then I'm fine with the telemetrystatistics class having separate methods.

genai-perf/genai_perf/metrics/telemetry_statistics.py
```diff
@@ -0,0 +1,112 @@
...
+        p90 = percentile_results[3]
+        p95 = percentile_results[4]
+        p99 = percentile_results[5]
+        setattr(self, f"p25_{attr}_{gpu_index}", p25)
+        setattr(self, f"p50_{attr}_{gpu_index}", p50)
+        setattr(self, f"p75_{attr}_{gpu_index}", p75)
+        setattr(self, f"p90_{attr}_{gpu_index}", p90)
+        setattr(self, f"p95_{attr}_{gpu_index}", p95)
+        setattr(self, f"p99_{attr}_{gpu_index}", p99)
+        self._stats_dict[attr][gpu_index]["p99"] = p99
```
nv-hwoo, line 84: One quick idea that popped on my head, what do you think about restructuring the `stats_dict` into:
```python
self._stats_dict[gpu_index][attr]["p99"]
```
so that you index for gpu id first and then for attribute? If this restructuring is possible, then I think we can reuse `Statistics` method (although it would require a small refactor, but I think it's possible).

lkomali: Restructuring as self._stats_dict[gpu_index][attr]["p99"] is possible. I see what you mean.
Yeah.. I guess it can work. Thanks for the suggestion @nv-hwoo

genai-perf/genai_perf/metrics/telemetry_statistics.py
lkomali: The overlap is both use MetricMetadata to store metrics. Other than that, I can't find similarity in both of them. Let me know if I'm wrong.

genai-perf/genai_perf/metrics/telemetry_statistics.py
lkomali: I modified as per your suggestion and our offline discussion.
I made Statistics and TelemetryStatistics to have composition instead of Inheritance relationship as that made more sense with different metric types. Also, not calling super.__init__() while using inheritance gives a codeql error.
I was able to modify without changing the self._stats_dict structure in TelemetryStatistics.

genai-perf/genai_perf/parser.py
lkomali: I moved this part to main.py and renamed and modified the corresponding test.

nv-hwoo (approved):

PR #73, opened 2024-09-04
Tokens to TRTLLM Backend

IzzyPutterman:


genai-perf/genai_perf/inputs/input_constants.py
matthewkotila, line 1: @IzzyPutterman is this still something we want merged?

IzzyPutterman: Yes I think we do.  We can query the trtllm backend with this, when hosted by triton.

dyastremsky: Same question for Izzy a few months later. Is this feature already in with the direct C-API TRT-LLM engine endpoint-type? If so, we could close out this PR.

CC: @debermudez

PR #74, opened 2024-09-05
Add functionality to perform warmup before benchmarking

matthewkotila:
Example:
```bash
perf_analyzer --warmup-request-count=3 --request-count=2 ...
```

* Allows user to specify fixed number of "warmup" requests that happen before benchmarking
* In concurrency mode, warmup requests get sent at specified concurrency
  * Only `N` total warmup requests get sent via `--warmup-request-count=N`, regardless of concurrency
* In request rate mode, warmup requests get sent at specified request rate
* In custom intervals mode, warmup requests get sent at specified custom intervals
* `--warmup-request-count` works independent of `--request-count` mode being used

### DIFF
--- a/src/command_line_parser.cc
+++ b/src/command_line_parser.cc
@@ -824,71 +824,90 @@ CLParser::ParseCommandLine(int argc, char** argv)
   argc_ = argc;
   argv_ = argv;
 
+  constexpr int long_option_idx_base{256};
+
   // {name, has_arg, *flag, val}
   static struct option long_options[] = {
-      {"streaming", no_argument, 0, 0},
-      {"max-threads", required_argument, 0, 1},
-      {"sequence-length", required_argument, 0, 2},
-      {"percentile", required_argument, 0, 3},
-      {"data-directory", required_argument, 0, 4},
-      {"shape", required_argument, 0, 5},
-      {"measurement-interval", required_argument, 0, 6},
-      {"concurrency-range", required_argument, 0, 7},
-      {"latency-threshold", required_argument, 0, 8},
-      {"stability-percentage", required_argument, 0, 9},
-      {"max-trials", required_argument, 0, 10},
-      {"input-data", required_argument, 0, 11},
-      {"string-length", required_argument, 0, 12},
-      {"string-data", required_argument, 0, 13},
-      {"async", no_argument, 0, 14},
-      {"sync", no_argument, 0, 15},
-      {"request-rate-range", required_argument, 0, 16},
-      {"num-of-sequences", required_argument, 0, 17},
-      {"binary-search", no_argument, 0, 18},
-      {"request-distribution", required_argument, 0, 19},
-      {"request-intervals", required_argument, 0, 20},
-      {"shared-memory", required_argument, 0, 21},
-      {"output-shared-memory-size", required_argument, 0, 22},
-      {"service-kind", required_argument, 0, 23},
-      {"model-signature-name", required_argument, 0, 24},
-      {"grpc-compression-algorithm", required_argument, 0, 25},
-      {"measurement-mode", required_argument, 0, 26},
-      {"measurement-request-count", required_argument, 0, 27},
-      {"triton-server-directory", required_argument, 0, 28},
-      {"model-repository", required_argument, 0, 29},
-      {"sequence-id-range", required_argument, 0, 30},
-      {"ssl-grpc-use-ssl", no_argument, 0, 31},
-      {"ssl-grpc-root-certifications-file", required_argument, 0, 32},
-      {"ssl-grpc-private-key-file", required_argument, 0, 33},
-      {"ssl-grpc-certificate-chain-file", required_argument, 0, 34},
-      {"ssl-https-verify-peer", required_argument, 0, 35},
-      {"ssl-https-verify-host", required_argument, 0, 36},
-      {"ssl-https-ca-certificates-file", required_argument, 0, 37},
-      {"ssl-https-client-certificate-file", required_argument, 0, 38},
-      {"ssl-https-client-certificate-type", required_argument, 0, 39},
-      {"ssl-https-private-key-file", required_argument, 0, 40},
-      {"ssl-https-private-key-type", required_argument, 0, 41},
-      {"verbose-csv", no_argument, 0, 42},
-      {"enable-mpi", no_argument, 0, 43},
-      {"trace-level", required_argument, 0, 44},
-      {"trace-rate", required_argument, 0, 45},
-      {"trace-count", required_argument, 0, 46},
-      {"log-frequency", required_argument, 0, 47},
-      {"collect-metrics", no_argument, 0, 48},
-      {"metrics-url", required_argument, 0, 49},
-      {"metrics-interval", required_argument, 0, 50},
-      {"sequence-length-variation", required_argument, 0, 51},
-      {"bls-composing-models", required_argument, 0, 52},
-      {"serial-sequences", no_argument, 0, 53},
-      {"input-tensor-format", required_argument, 0, 54},
-      {"output-tensor-format", required_argument, 0, 55},
-      {"version", no_argument, 0, 56},
-      {"profile-export-file", required_argument, 0, 57},
-      {"periodic-concurrency-range", required_argument, 0, 58},
-      {"request-period", required_argument, 0, 59},
-      {"request-parameter", required_argument, 0, 60},
-      {"endpoint", required_argument, 0, 61},
-      {"request-count", required_argument, 0, 62},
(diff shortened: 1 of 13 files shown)


src/command_line_parser.cc
```diff
@@ -824,71 +824,90 @@ CLParser::ParseCommandLine(int argc, char** argv)
   argc_ = argc;
   argv_ = argv;
 
+  constexpr int long_option_idx_base{256};
```
matthewkotila, line 827: We started having so many long options that they indices were overlapping with short options (e.g. ascii `?` is integer 63), so I created a base value beyond all ascii values to start the long options from

src/command_line_parser.cc
```diff
@@ -1688,13 +1716,14 @@ CLParser::ParseCommandLine(int argc, char** argv)
       }
     }
     catch (const std::invalid_argument& ia) {
-      if (opt >= 'A') {  // short options
+      if (opt < long_option_idx_base) {  // short options
```
matthewkotila, line 1719: Reworked logic here because as mentioned [here](https://github.com/triton-inference-server/perf_analyzer/pull/74/files#r1744758068), we start all long options at 256 now, and all options below that are short options.

src/command_line_parser.cc
```diff
@@ -1688,13 +1716,14 @@ CLParser::ParseCommandLine(int argc, char** argv)
...
         Usage(
             "Failed to parse -" + std::string{(char)opt} +
             ". Invalid value provided: " + std::string{optarg});
       } else {
         Usage(
-            "Failed to parse --" + std::string{long_options[opt].name} +
+            "Failed to parse --" +
+            std::string{long_options[opt - long_option_idx_base].name} +
```
matthewkotila, line 1726: In order to get the name of long options now, we need to subtract away the base index to get the actual index of the long option we want in the `getopt.h` data structure for holding long options.

src/concurrency_manager.cc
```diff
@@ -82,6 +82,16 @@ ConcurrencyManager::InitManagerFinalize()
   }
 }
 
+cb::Error
+ConcurrencyManager::PerformWarmup(
```
matthewkotila, line 86: Each load manager has its own `PerformWarmup()` since each load manager has a different function call for beginning load (i.e. `ChangeConcurrencyLevel()`, `ChangeRequestRate()`, and `InitCustomIntervals()`

src/custom_load_manager.cc
```diff
@@ -175,4 +183,11 @@ CustomLoadManager::ReadTimeIntervalsFile(
   return cb::Error::Success;
 }
 
+void
+CustomLoadManager::WaitForWarmupAndCleanup()
+{
+  RequestRateManager::WaitForWarmupAndCleanup();
+  custom_intervals_.clear();
```
matthewkotila, line 190: CustomLoadManager has an additional data structure that needs to be reset after warmup.

src/load_manager.cc
```diff
@@ -272,6 +270,20 @@ LoadManager::StopWorkerThreads()
...
+{
+  for (std::thread& thread : threads_) {
+    if (thread.joinable()) {
+      thread.join();
+    }
+  }
+  threads_.clear();
+  threads_config_.clear();
+  threads_stat_.clear();
+  workers_.clear();
```
matthewkotila, line 284: These are the four data structures that all load managers share that need to be reset before actual benchmarking begins.

src/load_manager.h
```diff
@@ -147,7 +150,6 @@ class LoadManager {
   bool using_json_data_;
 
   std::shared_ptr<DataLoader> data_loader_;
-  std::unique_ptr<cb::ClientBackend> backend_;
```
matthewkotila, line 150: This wasn't actually used--random cleanup.

src/load_manager.cc
```diff
@@ -213,8 +213,6 @@ LoadManager::InitManagerInputs(
     const size_t string_length, const std::string& string_data,
     const bool zero_input, std::vector<std::string>& user_data)
 {
-  RETURN_IF_ERROR(factory_->CreateClientBackend(&backend_));
```
matthewkotila, line 216: This code had no downstream effects--this `backend_` wasn't used anywhere--random cleanup.

src/request_rate_manager.h
```diff
@@ -152,8 +160,6 @@ class RequestRateManager : public LoadManager {
 
   size_t DetermineNumThreads();
 
-  std::vector<std::shared_ptr<ThreadConfig>> threads_config_;
```
matthewkotila, line 155: See [here](https://github.com/triton-inference-server/perf_analyzer/pull/74/files#r1744761776)

src/load_manager.h
```diff
@@ -158,6 +160,8 @@ class LoadManager {
   std::vector<std::thread> threads_;
   // Contains the statistics on the current working threads
   std::vector<std::shared_ptr<ThreadStat>> threads_stat_;
+  // Contains the configs for the current working threads
+  std::vector<std::shared_ptr<ThreadConfig>> threads_config_;
```
matthewkotila, line 164: See [here](https://github.com/triton-inference-server/perf_analyzer/pull/74/files#r1744761776)

src/inference_profiler.cc
```diff
@@ -561,6 +561,9 @@ InferenceProfiler::Profile(
   is_stable = false;
   meets_threshold = true;
 
+  RETURN_IF_ERROR(
+      dynamic_cast<ConcurrencyManager*>(manager_.get())
+          ->PerformWarmup(concurrent_request_count, warmup_request_count));
```
nv-hwoo, line 566: How does stats and timestamps collected during the warmup get discarded? (I'm guessing this is achieved through `WaitForWarmupAndCleanup` by cleaning up all the worker threads that contain the record/data, but just want to confirm).

src/load_manager.h
debermudez: nice! get that unused variable out of here

src/load_manager.cc
debermudez: where is the backend created if not here?

src/inference_profiler.cc
debermudez: Do we want to clean up the warmup data or should it be collected in a separate portion of the export json?
I could see a case where user's would want that data too.

debermudez (commented):
Is there testing around this functionality?

src/inference_profiler.cc
matthewkotila: The stats are in `thread_stats_`, which gets cleared in `WaitForWarmupAndCleanup()`.

The timestamps only get saved when `ProfileHelper()` gets called, but warmup happens before that.

dyastremsky (approved):
Appreciate all the clean-up, Matt!

I have the [same question](https://github.com/triton-inference-server/perf_analyzer/pull/74#discussion_r1745946515) as Elias about where the backend creation happens, but otherwise, all looks good!

src/command_line_parser.cc
matthewkotila: Sure, the options deserve a deeper dive/redesign.

src/load_manager.cc
matthewkotila: There are several places a backend is created and later used. You can search for calls to `CreateClientBackend()` in the repository.

The one you're probably thinking of where the backend is used to send inference requests is here:

https://github.com/triton-inference-server/perf_analyzer/blob/a3e3d27895fab652e4d766e4fc86dd96a7723d6a/src/infer_context.h#L93

src/inference_profiler.cc
debermudez: But do we want to throw this data away?

src/command_line_parser.cc
matthewkotila, line 1: > @debermudez: Is there testing around this functionality?

(responding here so we have a thread)

Previously it was just manual testing by me. I now have added an end-to-end test.

src/inference_profiler.cc
matthewkotila: Yup. If someone wanted it they can skip using the warmup feature and just specify more requests via `--request-count`

nv-hwoo: sounds good to me. will defer to @debermudez on closing this comment.

nv-hwoo (approved):
nice and clean  🚀

src/command_line_parser.cc
debermudez: approved that test! thanks man.

src/inference_profiler.cc
debermudez: yea thats really reasonable.
appreciate the answer.

debermudez (dismissed):
Beautiful!

dyastremsky (approved):

PR #75, opened 2024-09-05
Export Telemetry Statistics

lkomali:
Export Telemetry Statistics
1. CSV output
image
2. JSON output
image
image
3. Console output
image
image


genai-perf/tests/test_console_exporter.py
```diff
@@ -288,49 +296,236 @@ def test_invalid_goodput_output(self, monkeypatch, capsys) -> None:
...
-            ("rankings", Metrics(), "NVIDIA GenAI-Perf | Rankings Metrics"),
-            ("vision", LLMMetrics(), "NVIDIA GenAI-Perf | VLM Metrics"),
-            (
-                "image_retrieval",
-                ImageRetrievalMetrics(),
-                "NVIDIA GenAI-Perf | Image Retrieval Metrics",
-            ),
-        ],
-    )
-    def test_console_title(
```
dyastremsky, line 309: Did you intend to delete this test?

genai-perf/tests/test_csv_exporter.py
```diff
@@ -335,3 +341,151 @@ def test_invalid_goodput_csv_output(
...
+        config.stats = stats.stats_dict
+        config.telemetry_stats = telemetry_stats.stats_dict
+        config.metrics = stats.metrics
+        config.artifact_dir = Path(".")
+        config.args = args
+
+        exporter = CsvExporter(config)
+        exporter.export()
+
+        expected_content = [
```
dyastremsky, line 394: Since other tests already test the non-Telemetry parts, could you update the tests to only test that the telemetry part is included in the final output? That should be the scope of the Telemetry-specific tests, rather than repeating what other tests are checking.

genai-perf/tests/test_csv_exporter.py
```diff
@@ -335,3 +341,151 @@ def test_invalid_goodput_csv_output(
...
+        ]
+
+        expected_filename = "profile_export_genai_perf.csv"
+        returned_data = [
+            data
+            for filename, data in mock_read_write
+            if os.path.basename(filename) == expected_filename
+        ]
+        if returned_data == []:
+            raise Exception(
```
dyastremsky, line 488: You shouldn't be raising exceptions in a test. The test should be asserting a failure.

genai-perf/genai_perf/export_data/json_exporter.py
```diff
@@ -43,6 +46,7 @@ class JsonExporter:
 
     def __init__(self, config: ExporterConfig):
         self._stats: Dict = config.stats
+        self._telemetry_stats: Optional[Dict] = config.telemetry_stats
```
dyastremsky, line 49: Could you make this type hint more specific? What kind of dict is it?

genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
```diff
@@ -0,0 +1,190 @@
...
+
+    @staticmethod
+    def merge_telemetry_stats_json(
+        telemetry_stats: Optional[Dict], stats_and_args: Dict
+    ) -> None:
+        if telemetry_stats is not None:
+            stats_and_args.update({"telemetry_stats": telemetry_stats})
+
+    @staticmethod
+    def export_telemetry_stats_csv(telemetry_stats: Optional[Dict], csv_writer) -> None:
```
dyastremsky, line 71: Can you add the type hint for csv_writer? I know we don't do it elsewhere yet, but it would be good to start doing.

e.g.
```
from csv import writer as CsvWriter
def export_telemetry_stats_csv(telemetry_stats: Optional[Dict], csv_writer: CsvWriter) -> None:
```

genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
```diff
@@ -0,0 +1,190 @@
...
+            TelemetryDataExporterUtil._write_dynamic_telemetry_stats(
+                telemetry_stats, csv_writer
+            )
+            TelemetryDataExporterUtil._write_constant_telemetry_stats(
+                telemetry_stats, csv_writer
+            )
+
+    @staticmethod
+    def export_telemetry_stats_console(
+        telemetry_stats: Optional[Dict], STAT_COLUMN_KEYS: List[str], console: Console
```
dyastremsky, line 82: Is this supposed to be capitalized? I think anywhere you are using something as a parameter rather than a constant, it should be lower snake case to adhere to PEP8. Could you go through the changes to do so, if you think that makes sense?

genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
```diff
@@ -0,0 +1,190 @@
...
+            metric_str += f" ({metric_data['unit']})"
+            for gpu in metric_data.keys():
+                if gpu == "unit":
+                    continue
+                value = metric_data[gpu]["avg"]
+                csv_writer.writerow([metric_str, gpu, f"{value:.2f}"])
+
+    @classmethod
+    def _format_metric_name(cls, metric_name: str) -> str:
+        abbreviations = ["gpu"]
```
dyastremsky, line 183: This should be a constant. You don't want magic values in your code.

Also, can all the constants be moved into their own file for maintainability?

genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
```diff
@@ -0,0 +1,190 @@
...
+
+    @classmethod
+    def _construct_telemetry_stats_subtable(
+        cls,
+        table: Table,
+        metric_data: Dict[str, Dict[str, float]],
+        metric_name: str,
+        STAT_COLUMN_KEYS: List[str],
+    ) -> None:
+        gpu_indices = sorted([key for key in metric_data.keys() if key != "unit"])
```
dyastremsky, line 127: Can you review this function? You're sorting and then using a nested loop to get the values in this dictionary. I wonder if there's a more efficient way to do this.

dyastremsky (commented):
Great work! This is a very solid design.

Left a few questions/comments.

genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
```diff
@@ -0,0 +1,190 @@
...
+    @staticmethod
+    def export_telemetry_stats_console(
+        telemetry_stats: Optional[Dict], STAT_COLUMN_KEYS: List[str], console: Console
+    ) -> None:
+        if telemetry_stats:
+            TelemetryDataExporterUtil._construct_telemetry_stats_table(
+                telemetry_stats, STAT_COLUMN_KEYS, console
+            )
+
+    @classmethod
```
debermudez, line 89: Why are some class methods and others are static methods?

genai-perf/tests/test_console_exporter.py
lkomali: Did I delete it? Let me check. I didn't wanna delete any test. It might have happened by mistake.

genai-perf/tests/test_csv_exporter.py
lkomali: I followed what the other tests are doing. 
Should I remove all the Exception lines from all the tests?

debermudez (commented):
looks pretty good to me. Echoing @dyastremsky's comment about the test.

genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
lkomali: So, I made all the utility functions for exporting data which do not interact with any class members as staticmethods.
The ones I made classmethods, I wanted them to be called without an object and they interact with class members, like the variables I created for headers etc. So, I made them classmethods.

genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
```diff
@@ -0,0 +1,190 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from typing import Dict, List, Optional
+
+from rich.console import Console
+from rich.table import Table
+
+
+class TelemetryDataExporterUtil:
```
nv-hwoo, line 33: is there a reason why all these util functions are defined under `TelemetryDataExporterUtil` class? I think it's relatively common in the python world to define all the utility functions as just plain functions:
- https://github.com/google/jax/blob/main/jax/tools/build_utils.py
- https://github.com/pytorch/pytorch/blob/main/torch/_functorch/utils.py
- https://github.com/fastapi/fastapi/blob/master/fastapi/dependencies/utils.py

This avoids any need to care about `classmethod` or `staticmethod` (since the class attributes can just be global variables). Another benefit with this approach is we can abbreviate the long prefix when importing this utils module:
```python
from ... import telemetry_data_exporter_utils as telem_utils

telem_utils.export_telemetry_stats_csv(...)
```
instead of 
```python
from <...>.telemetry_data_exporter_utils import TelemetryDataExporterUtil

TelemetryDataExporterUtil.export_telemetry_stats_csv(...)
```
which IMO is less readable than the former.

genai-perf/tests/test_csv_exporter.py
nv-hwoo: yes, I agree with @dyastremsky. Could you change this to below?
```python
assert expected_filename in returned_data
```

genai-perf/tests/test_console_exporter.py
```diff
@@ -288,49 +296,236 @@ def test_invalid_goodput_output(self, monkeypatch, capsys) -> None:
...
+            "│ ┡━━━━━━━━━━━╇━━━━━╇━━━━━╇━━━━━╇━━━━━╇━━━━━╇━━━━━┩ │\n"
+            "│ │      gpu0 │  78 │  75 │  80 │  80 │  80 │  79 │ │\n"
+            "│ └───────────┴─────┴─────┴─────┴─────┴─────┴─────┘ │\n"
+            "└───────────────────────────────────────────────────┘\n"
+        )
+
+        returned_data = capsys.readouterr().out
+        assert returned_data == expected_content
+
+    def test_invalid_telemetry_verbose(self, monkeypatch, capsys) -> None:
```
nv-hwoo, line 478: what is the test trying to check? the test name sounds like we should be catching some error but I only see normal console output checks.

genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
lkomali: It's of type csv._writer. I think since it's an internal class, and the reason why we don't do it in other places. 
Should I add csv._writer?

genai-perf/tests/test_csv_exporter.py
```diff
@@ -335,3 +341,151 @@ def test_invalid_goodput_csv_output(
...
+            "--service-kind",
+            "triton",
+            "--streaming",
+            "--server-metrics-url",
+            "http://tritonserver:8002/metrics",
+        ]
+        monkeypatch.setattr("sys.argv", argv)
+        args, _ = parser.parse_args()
+
+        metrics = LLMMetrics(
```
nv-hwoo, line 362: I see a lot of tests with these duplicate lines... is there a way to avoid this? (maybe using `pytext.fixture` could help?)

genai-perf/tests/test_json_exporter.py
```diff
@@ -682,3 +682,399 @@ def test_invalid_goodput_json_output(
...
+                f"Expected file {expected_filename} not found in written data."
+            )
+        assert len(written_data) == 1
+        output_data_dict = json.loads(written_data[0])
+
+        assert output_data_dict["telemetry_stats"] == json.loads(
+            expected_telemetry_json_output
+        )
+
+    def test_triton_telemetry_output_failure(
```
nv-hwoo, line 952: same here. the test name "failure" makes it sounds like we need to catch some kind of error. I think more appropriate name would be something like `test_triton_telemetry_no_output`.

genai-perf/tests/test_json_exporter.py
```diff
@@ -682,3 +682,399 @@ def test_invalid_goodput_json_output(
         assert output_data_dict["input_config"]["goodput"] == json.loads(
             expected_invalid_goodput_json_config
         )
+
+    def test_triton_telemetry_output_success(
+        self, monkeypatch, mock_read_write: pytest.MonkeyPatch
+    ) -> None:
+        stats = {
```
nv-hwoo, line 689: this seems to be reused over and over in each tests. can we take it out using `pytest.fixture`?

genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
lkomali: This doesn't work. It gives an error. 
It's of type csv._writer and do we want to use it? Asking this because this seems like a private method.

genai-perf/tests/test_console_exporter.py
lkomali: I removed tests that do not contain telemetry data as they are covered by other tests as @dyastremsky suggested.

genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
dyastremsky: Oh, that's strange. What's the error?

lkomali: Function "_csv.writer" is not valid as a typeMypy[valid-type](https://mypy.readthedocs.io/en/latest/_refs.html#code-valid-type)
Perhaps you need "Callable[...]" or a callback protocol?Mypy

lkomali: I tried a couple
```
import csv
csv_writer: csv.writer
``` gives MyPy error
```
import csv
csv_writer: csv._writer
``` gives error in the console
```
import _csv
csv_writer: _csv._writer
```
gives error in the console

File: genai-perf/tests/test_json_exporter.py (hunk 14 above)
Comment (lkomali) [reply genai-perf/tests/test_json_exporter.py]: I didn't see that the stats is a class-level member. So, I re-organized and used it.

File: genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py (hunk 5 above)
Comment (nv-hwoo) [reply genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py]: Looks like the csv writer object type is not accessable as it's defined in the C source code. But the following code seems to work with `mypy`:
```python
# foo.py
import csv
import _csv

with open("file.txt", "w") as f:
    w: _csv._writer = csv.writer(f)
```
and run 
```bash
mypy foo.py
```

ref: https://stackoverflow.com/questions/51264355/how-to-type-annotate-object-returned-by-csv-writer

genai-perf/tests/test_csv_exporter.py
lkomali: expected_filename is not in returned_data. It's only used to read from.

genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
lkomali: In type annotation, I'm unable to use _csv._writer or csv._writer or csv.writer.

genai-perf/tests/test_json_exporter.py
```diff
@@ -361,6 +255,112 @@ def test_generate_json_custom_export(
...
+        expected_filename = "profile_export_genai_perf.json"
+        written_data = [
+            data
+            for filename, data in mock_read_write
+            if os.path.basename(filename) == expected_filename
+        ]
+        if written_data == []:
+            raise Exception(
+                f"Expected file {expected_filename} not found in written data."
+            )
```
nv-hwoo, line 315:
```suggestion
        assert expected_filename in written_data
```

genai-perf/tests/test_json_exporter.py
```diff
@@ -361,6 +255,112 @@ def test_generate_json_custom_export(
...
+        config.artifact_dir = args.artifact_dir
+        json_exporter = JsonExporter(config)
+        json_exporter.export()
+        written_data = [
+            data for filename, data in mock_read_write if filename == expected_filename
+        ]
+        if written_data == []:
+            raise Exception(
+                f"Expected file {expected_filename} not found in written data."
+            )
```
nv-hwoo, line 356:
```suggestion
        assert expected_filename in written_data
```

genai-perf/tests/test_json_exporter.py
```diff
@@ -682,3 +682,184 @@ def test_invalid_goodput_json_output(
...
+        expected_filename = "profile_export_genai_perf.json"
+        written_data = [
+            data
+            for filename, data in mock_read_write
+            if os.path.basename(filename) == expected_filename
+        ]
+        if written_data == []:
+            raise Exception(
+                f"Expected file {expected_filename} not found in written data."
+            )
```
nv-hwoo, line 859:
```suggestion
        assert expected_filename in written_data
```

genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
nv-hwoo: that's strange.. seems like this is a bit rare case so I'm fine dealing with the proper type hint later, but will defer to @dyastremsky's opinion on this.

nv-hwoo (commented):
looks great. just a few additional comments and I think it's ready for approval

genai-perf/tests/test_json_exporter.py
lkomali: expected_filename not in written_data. Removed exceptions.
Thanks for catching it.

genai-perf/tests/test_json_exporter.py
nv-hwoo: @lkomali I think you forgot this one

nv-hwoo (approved):
nice work 👍

PR #76, opened 2024-09-05
Add SearchParameters class

nv-braf:
Ports the SearchParameters class from MA to GAP.

I've also created some necessary collateral (ModelSpec and ConfigCommand) that was required to properly unit test this class.

Not all code is ready yet (as either more code is needed outside of this class or features, like BLS, that will come online after phase 1).

For now I have added: "TODO: Optimize" and commented out the code (rather than delete it).

Note: I have also moved the requirement from Python 3.8 -> Python 3.10 (to all for TypeAlias support)

### DIFF
--- a/genai-perf/genai_perf/config/generate/search_parameters.py
+++ b/genai-perf/genai_perf/config/generate/search_parameters.py
@@ -30,7 +30,7 @@ class SearchParameters:
     Contains information about all configuration parameters the user wants to search
     """
 
-    # These map to the run-config-search fields
+    # These map to the various fields that can be set for PA and model configs
     # See github.com/triton-inference-server/model_analyzer/blob/main/docs/config.md
     exponential_range_parameters = [
         "model_batch_size",
--- a/genai-perf/genai_perf/config/generate/search_parameters.py
+++ b/genai-perf/genai_perf/config/generate/search_parameters.py
@@ -15,10 +15,13 @@
 from math import log2
 from typing import Any, Dict, List, Optional, Tuple, Union
 
+from genai_perf.config.generate.search_parameter import (
+    ParameterCategory,
+    ParameterUsage,
+    SearchParameter,
+)
 from genai_perf.config.input.config_command import ConfigCommand, Range
 from genai_perf.exceptions import GenAIPerfException
-
-from genai_perf.config.generate.search_parameter import ParameterCategory, ParameterUsage, SearchParameter
 
 
 class SearchParameters:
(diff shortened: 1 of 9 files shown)


genai-perf/genai_perf/config/generate/search_parameters.py
```diff
@@ -0,0 +1,382 @@
...
+
+    def get_type(self, name: str) -> ParameterUsage:
+        return self._search_parameters[name].usage
+
+    def get_category(self, name: str) -> ParameterCategory:
+        return self._search_parameters[name].category
+
+    def get_range(self, name: str) -> Range:
+        return Range(
+            min=self._search_parameters[name].min_range,  # type: ignore
```
debermudez, line 84: why do we need to ignore typing here?

genai-perf/genai_perf/config/generate/search_parameters.py
```diff
@@ -0,0 +1,382 @@
...
+    ParameterCategory,
+    ParameterList,
+    ParameterUsage,
+    SearchParameter,
+)
+from genai_perf.config.input.config_command import ConfigCommand, Range
+from genai_perf.exceptions import GenAIPerfException
+
+
+class SearchParameters:
```
debermudez, line 28: 2 classes that only differ because of pluralization feels like a easy way to confuse the two.
This may be the best name, I dont know. I do know that I will be confused by this in the future.

genai-perf/genai_perf/config/input/config_command.py
```diff
@@ -0,0 +1,117 @@
...
+from typing import List, Optional, TypeAlias, Union
+
+
+def default_field(obj):
+    return field(default_factory=lambda: copy(obj))
+
+
+# TODO: OPTIMIZE
+# These will be moved to RunConfig once it's created
+@dataclass(frozen=True)
```
debermudez, line 26: 👨🏼‍🍳 💋 
Such a cool way to encapsulate constants into a logical grouping.

genai-perf/genai_perf/config/generate/search_parameters.py
nv-braf: The underlying type is Optional[int], because the parameter could be a list instead of a range (in which case min/max_range would be None). But, if we are calling this method then you know that you are using a range and not a list. However, there's no way for me to communicate this to the static type checker...

genai-perf/genai_perf/config/generate/search_parameters.py
nv-braf: Yeah, this was brought up in the MA review as well and my answer there was that no one will ever use the SearchParameter class except for SearchParameters, so there really shouldn't be any confusion in the future.

genai-perf/genai_perf/config/generate/search_parameters.py
debermudez: I feel like there should be a way, albeit probably very complex, to indicate this.
This is not a sticking point tho worth gating an approval.
Thanks for explaining it to me.

genai-perf/genai_perf/config/generate/search_parameters.py
debermudez: I vaguely remember that. 
Thanks!

debermudez (approved):

PR #77, opened 2024-09-06
Add synthetic data retriever class

nv-hwoo:
This PR is to extract out the synthetic data generation part into a separate API.
This PR only handles addition of the new API and does _not_ contain
 - lifting the dependencies to the endpoint (will have follow up PR)
 - integrating into the current code as it requires @debermudez's factory code to be in place.

### DIFF
--- a/genai-perf/genai_perf/inputs/synthetic_data_retriever.py
+++ b/genai-perf/genai_perf/inputs/synthetic_data_retriever.py
@@ -1,0 +1,64 @@
+# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of NVIDIA CORPORATION nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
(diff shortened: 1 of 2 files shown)


genai-perf/genai_perf/inputs/synthetic_data_retriever.py
```diff
@@ -0,0 +1,64 @@
...
+        synthetic_dataset = []
+        for _ in range(self.config.num_prompts):
+            prompt = SyntheticPromptGenerator.create_synthetic_prompt(
+                self.config.tokenizer,
+                self.config.prompt_tokens_mean,
+                self.config.prompt_tokens_stddev,
+            )
+            data = {"text_input": prompt}
+
+            if self.config.output_format == OutputFormat.OPENAI_VISION:
```
dyastremsky, line 53: Thanks for including images in your initial work! Are you going to refactor this out later? We need to have the data format be standard, so we'll need to have a way to create synthetic images for any APIs that use them.

I think we do not want any logic specific to the output formats on the retriever side when we finish modularization.

dyastremsky:
Also, are you planning to pull the current synthetic code out in a future PR? It looks like it was not deleted in this PR.

genai-perf/genai_perf/inputs/synthetic_data_retriever.py
nv-hwoo: yeah sorry for not adding any explanation on the PR. This PR is just for splitting up the synthetic part into a separate API, so we will have a follow up PR for further refining.

nv-hwoo:
@dyastremsky you are right this PR did not remove the current code. That requires input factory code to be in place so this PR only adds the new API and will be integrated into the new workflow in the follow up PRs.

genai-perf/genai_perf/inputs/synthetic_data_retriever.py
dyastremsky: Got it, thanks for clarifying!

dyastremsky (approved):

PR #78, opened 2024-09-06
Port Records and ModelConfigMeasurement classes

nv-braf:
Ports the Record and ModelConfigMeasurement classes into GAP.

The Record class is used to store individual metrics (like latency or throughput).
The MCM class is used to store all the perf metrics for a model configuration and has methods to compare configurations, change objectives and checkpoint.

### DIFF
--- a/genai-perf/tests/test_model_config_measurement.py
+++ b/genai-perf/tests/test_model_config_measurement.py
@@ -1,0 +1,242 @@
+# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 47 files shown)


genai-perf/tests/test_model_config_measurement.py
```diff
@@ -0,0 +1,207 @@
+# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
debermudez, line 1: With so many additions, there is not a lot of unit testing.
Do you exercise every type of record?

nv-braf: The record class has never had direct unit testing (it relies on indirect testing as it is used in a variety of unit testbenches). The MCM unit testing covers almost all of the base classes methods. And, the individual records will be exercised in unit tests once we port/create the code that handles the conversion of PA results to RCM/MCM.

nv-braf: Good catch. There actually is some record type unit testing that I missed porting over. I'll add this on Monday.

nv-braf: I've added the missing testing (and it also uncovered that I missed a file!). Please re-review when you get a chance.

debermudez (approved):

PR #79, opened 2024-09-07
Update doc templates

debermudez:
Update templates to match current docs.
Add the genai-perf version to the generation script.

### DIFF
--- a/templates/template_vars.yaml
+++ b/templates/template_vars.yaml
@@ -1,6 +1,7 @@
 General:
   release: 24.08
   triton_cli_version: 0.0.8
+  genai_perf_version: 0.0.6dev
 
 README:
   filename: README.md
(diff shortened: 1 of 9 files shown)


templates/template_vars.yaml
```diff
@@ -41,3 +41,9 @@ tutorial:
   filename: tutorial.md
   template: genai-perf-templates/tutorial_template
   output_dir: ../genai-perf/docs/
+
+version:
+  filename: __init__.py
+  template: genai-perf-templates/version_template
+  output_dir: ../genai-perf/genai_perf/
+  version: 0.0.6dev
```
fpetrini15, line 49: Can we add this version to the `General` variables as well?

debermudez: added

fpetrini15 (approved):
LGTM!

PR #81, opened 2024-09-09
Update model used in chat tutorials

dyastremsky:
GPT2 has no default template for vLLM, so it now returns an error when used:
```
ERROR 09-08 11:58:06 serving_chat.py:113] Error in applying chat template from request: As of transformers v4.44, default chat template is no longer allowed, so you must provide a chat template if the tokenizer does not define one.
INFO:     127.0.0.1:48058 - "POST /v1/chat/completions HTTP/1.1" 400 Bad Request
```

vLLM uses Zephyr in their tutorials. It has a chat template, so I am updating our documents to use it so that the tutorials continue to work.

I also updated a few of the nearby tutorial output tables to the current format. The goal of this PR isn't to entirely update the output tables, which can be done in a separate PR.


genai-perf/docs/tutorial.md
```diff
@@ -180,17 +180,17 @@ Output token throughput (per sec): 290.24
...
+### Run Zephyr on [OpenAI Chat Completions API](https://platform.openai.com/docs/api-reference/chat)-compatible server
 
 <details>
 <summary>See instructions</summary>
 
 Run the vLLM inference server:
 
 ```bash
-docker run -it --net=host --gpus=all vllm/vllm-openai:latest --model gpt2 --dtype float16 --max-model-len 1024
+docker run -it --net=host --gpus=all vllm/vllm-openai:latest --model HuggingFaceH4/zephyr-7b-beta --dtype float16 --max-model-len 1024
```
nv-hwoo, line 193: the context window of zephyr is much longer than 1024 (e.g. 32,768) so we may want to just remove the `--max-model-len 1024` part
```suggestion
docker run -it --net=host --gpus=all vllm/vllm-openai:latest --model HuggingFaceH4/zephyr-7b-beta --dtype float16
```

dyastremsky: Great catch. Thanks for reviewing so quickly and well, Hyunjae!

dyastremsky:
I updated the tutorial template. There's not an existing goodput template, so I created the ticket TPA-407 to add one in a separate PR.

debermudez (approved):

PR #82, opened 2024-09-10
fix bug in tensorrtllm converter

nv-hwoo:


matthewkotila (approved):

debermudez (approved):

genai-perf/genai_perf/inputs/converters/tensorrtllm_converter.py
```diff
@@ -96,6 +96,7 @@ def _populate_trtllm_output_json(
             pa_json = self._add_optional_tags_to_trtllm_json(
                 pa_json,
                 index,
+                config,
```
rmccorm4, line 99: Is there a test for streaming inference in Triton?

dyastremsky: Yes, [here](https://github.com/triton-inference-server/perf_analyzer/blob/a55259530eb6a7e0a3cece768826cb095e33ecb9/genai-perf/tests/test_triton_tensorrtllm_converter.py#L73). I don't know why this wasn't caught. We have been iterating very quickly with limited resources and had some CI issues, though this should have been caught in a Python package hook.

I'm rewriting these tests now as part of the customizable front-ends work.

PR #85, opened 2024-09-17
Add ability to change git remote repository in build

jbkyang-nvi:
Add git remote for long term builds that use non-git repos


CMakeLists.txt
matthewkotila, line 1: I don't quite understand what this is for--can you give a more concrete example?

Otherwise happy to defer to other reviewer(s)

fpetrini15 (approved):
Good catch Katherine! I see that the variable already exists in the PA src directory as well.

matthewkotila (approved):

PR #86, opened 2024-09-17
Add file input retriever class for reading from file source

nv-hwoo:
This PR
- introduces the new class `FileInputRetriever` that handles reading input data from files
- moves all the file related methods under `InputRetrieverFactory` class to the `FileInputRetriever` class
- moves all the unit tests related to the `FileInputRetriever` to `test_file_input_retriever.py`

This PR does _NOT_
- completely decouples endpoint logics from the retriever (this refining will be done in the follow up PR as there's a high change that will make the PR too big)

### DIFF
--- a/genai-perf/genai_perf/inputs/file_input_retriever.py
+++ b/genai-perf/genai_perf/inputs/file_input_retriever.py
@@ -1,0 +1,205 @@
+# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of NVIDIA CORPORATION nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import random
+from pathlib import Path
+from typing import Any, Dict, List, Tuple
+
+from genai_perf.exceptions import GenAIPerfException
+from genai_perf.inputs.input_constants import (
+    DEFAULT_BATCH_SIZE,
+    OutputFormat,
+    PromptSource,
+)
+from genai_perf.inputs.inputs_config import InputsConfig
+from genai_perf.utils import load_json_str
+
+
+class FileInputRetriever:
+    """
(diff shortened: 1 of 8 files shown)


genai-perf/genai_perf/inputs/file_input_retriever.py
```diff
@@ -0,0 +1,185 @@
...
+import random
+from pathlib import Path
+from typing import Any, Dict, List, Tuple
+
+from genai_perf.inputs.input_constants import DEFAULT_BATCH_SIZE, OutputFormat
+from genai_perf.inputs.inputs_config import InputsConfig
+from genai_perf.utils import load_json_str
+
+
+class FileInputRetriever:
```
dyastremsky, line 36: Do you feel it might still be a good idea to have a base retriever class? Then, we could put all the retrievers in their own repo and have them use the same high-level functions. That could clean up some code and add organization now, plus make it easier later to add new retrievers.

This may be part of this PR or a future PR. This PR is moving away from using the generalized factory though.

genai-perf/genai_perf/inputs/input_retriever_factory.py
```diff
@@ -62,44 +59,21 @@ def get_input_data(self) -> Dict:
...
-            input_file_dataset = self._get_input_dataset_from_file()
-            input_file_dataset = self._encode_images_in_input_dataset(
-                input_file_dataset
-            )
+        if self.config.output_format in [
+            OutputFormat.OPENAI_EMBEDDINGS,
+            OutputFormat.RANKINGS,
+            OutputFormat.IMAGE_RETRIEVAL,
+        ]:
+            # TODO: remove once the factory fully integrates retrievers
```
dyastremsky, line 67: Remove what here? Does this mean you're planning to have a factory that creates the correct retriever based on the input source in the future?

*If so, my comment above about a base retriever class can be resolved, since you're already planning to go that route.

genai-perf/genai_perf/inputs/inputs.py
```diff
@@ -81,6 +76,18 @@ def _write_json_to_file(self, json_in_pa_format: Dict) -> None:
         with open(str(filename), "w") as f:
             f.write(json.dumps(json_in_pa_format, indent=2))
 
+    def _check_for_supported_input_type(self) -> None:
```
dyastremsky, line 79: This is a nice addition. Long-term, it'd be good to move toward a more generic approach toward supported/unsupported types. We could even move it to a file that's easy for users to update when adding a new task.

e.g.
``` python
unsupported_combinations = {
            OutputFormat.OPENAI_EMBEDDINGS: [PromptSource.DATASET, PromptSource.SYNTHETIC],
            OutputFormat.RANKINGS: [PromptSource.DATASET, PromptSource.SYNTHETIC],
        }
if input_type in unsupported_combinations.get(output_format, []):
            raise GenAIPerfException(
                f"{dataset_name} does not have a corresponding URL in the dataset_url_map."
                f"{output_format.to_lowercase()} does not support input type `{input_type.to_lowercase()}`"
            )
```

dyastremsky (approved):
Spectacular work here, Hyunjae. This is a great reorganization of the code.

Added a few comments with discussions around long-term direction. The code looks good to merge into the feature branch.

genai-perf/genai_perf/inputs/file_input_retriever.py
nv-hwoo: I think that's a good idea, but I would make it a future PR since other retrievers need to be changed and currently I don't think there's much overlap between the retrievers to share.

genai-perf/genai_perf/inputs/input_retriever_factory.py
nv-hwoo: You are correct 😄  We want the factory to handle all the retrievers in the future based on the input source, but that requires all the retrievers to be implemented. Consider this change as a step towards that direction without breaking the current overall behavior.

genai-perf/genai_perf/inputs/inputs.py
nv-hwoo: Nice suggestion! Where do you think is a good place to add this? Atm, I can't think of other locations in mind other than `inputs.py`.

dyastremsky: Thanks! A generic format like this could even go within this function. Alternatively, it could be supported combinations, though I think that list would be much longer.

Longer-term, we'd want to centralize it somewhere. We could drop it in one of the constants files, but I think we'll ultimately want to think through the workflow of adding a new task and go from there.

One idea is the converter could be responsible for doing the check, since it's tied to output format and that means there's one fewer file the user has to touch to add a task. Lots of options though! Just seeding some ideas. This doesn't have to be resolved in this PR.

genai-perf/genai_perf/inputs/input_retriever_factory.py
dyastremsky: Perfect, thanks for clarifying! That sounds excellent.

genai-perf/genai_perf/inputs/file_input_retriever.py
dyastremsky: Awesome. I'll resolve this conversation. You might want to create a small ticket to refactor to use a retriever base class once all the retrievers are created if one does not yet exist (so that you don't need to remember it on your end).

genai-perf/genai_perf/inputs/inputs.py
nv-hwoo: Sounds good. Created TPA-425.

genai-perf/genai_perf/inputs/file_input_retriever.py
nv-hwoo: Created TPA-423

PR #88, opened 2024-09-18
PA: Fixed scheduler manager

IzzyPutterman:
image


src/inference_profiler.cc
```diff
@@ -619,8 +619,8 @@ InferenceProfiler::Profile(
   is_stable = false;
   meets_threshold = true;
 
-  RETURN_IF_ERROR(dynamic_cast<RequestRateManager*>(manager_.get())
-                      ->ChangeRequestRate(request_rate, request_count));
+  RETURN_IF_ERROR(dynamic_cast<FixedTimeManager*>(manager_.get())
```
IzzyPutterman, line 622: temp WAR as the constructors for the Inference Profiler conflict.

src/custom_load_manager.cc
```diff
@@ -67,10 +67,10 @@ CustomLoadManager::CustomLoadManager(
     const std::unordered_map<std::string, cb::RequestParameter>&
         request_parameters)
     : RequestRateManager(
-          async, streaming, Distribution::CUSTOM, batch_size,
-          measurement_window_ms, max_trials, max_threads, num_of_sequences,
-          shared_memory_type, output_shm_size, serial_sequences, parser,
-          factory, request_parameters),
+          async, streaming, Distribution::CUSTOM, std::vector<float>{},
```
debermudez, line 70: Adding the new param early in the list has some down stream effects on the tests that make use of default values.
This should get added to the end of the param list.

debermudez (commented):
This seems sane and I think works.
But I am interested to see how this works if the schedule falls behind.
Once PA gets behind, I dont think it can catch up. I think that behavior needs to also change to enable this schedule functionality to work.

src/command_line_parser.cc
```diff
@@ -1832,15 +1847,17 @@ CLParser::VerifyOptions()
       params_->using_custom_intervals};
   if (std::count(load_modes.begin(), load_modes.end(), true) > 1) {
     Usage(
-        "Cannot specify more then one inference load mode. Please choose only "
+        "Cannot specify more then one inference load mode. Please choose "
```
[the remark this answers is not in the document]
IzzyPutterman: Im pretty sure I got the same one as in genai-perf, but I can double check

IzzyPutterman:
> This seems sane and I think works. But I am interested to see how this works if the schedule falls behind. Once PA gets behind, I dont think it can catch up. I think that behavior needs to also change to enable this schedule functionality to work.

With one thread and the schedule I had in the slack thread I definitely saw it fall behind on one of the requests, but I think the subsequent deltas remained consistent.  Not sure how to deal with this unless we overprovision the threads or allow for dynamic allocation

src/command_line_parser.cc
matthewkotila, line 1: Can we get this rebased onto main? There are some merge conflicts that I want to review how they are resolved.

src/command_line_parser.cc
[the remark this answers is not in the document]
matthewkotila: +1, clang-format-16 should be used and it should be following the `.clang-format` file in the repo root

src/command_line_parser.cc
matthewkotila, line 1: In general, this kind of functionality is already supported via custom interval mode:

https://github.com/triton-inference-server/perf_analyzer/blob/main/docs/inference_load_modes.md#custom-interval-mode

Is this just a quality of life enhancement so that someone can provide timestamps (instead of interval durations) via the command line (instead of a file)?

IzzyPutterman: I mean the logic I truly implemented here is only a few lines, but the goal here is to imitate a request distribution exactly.  Something like this for example: https://raw.githubusercontent.com/kvcache-ai/Mooncake/refs/heads/main/mooncake_trace.jsonl

debermudez: The goal is to replicate a dataset exactly. 
But I still have the lingering question about what happens if PA falls behind.
We need a way to account for that or disable that behavior to ensure that requests are sent nearly exactly according to the schedule.

matthewkotila: @debermudez that's right that PA will stay fallen behind if the model is perpetually slower than the request rate--this PR doesn't implement a catching up feature

That would have to be a (probably much larger) piece of work. But, perhaps a user-side workaround is if someone sees a lot of delayed requests, they can increase the `--max-threads`. That should allow more time for a particular thread (for all threads) to wait between when it sends its current and next request (since the entire schedule is split up over more threads).

matthewkotila: @IzzyPutterman gotcha, so this is basically what I said, a quality of life enhancement. Makes perfect sense considering the datasets are written out in the format you mentioned. I'm on board with this.

debermudez: @matthewkotila makes sense to me. 
That seems like a sensible work around until we get a feature request stating this isnt viable.

src/command_line_parser.cc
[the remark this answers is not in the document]
matthewkotila: Friendly ping

src/command_line_parser.cc
matthewkotila: Friendly ping

src/test_request_rate_manager.cc
matthewkotila, line 1: I see changes here, but these aren't testing the new functionality. Can you add unit testing to the codebase for this new functionality?

src/perf_analyzer.cc
```diff
@@ -242,7 +242,7 @@ PerfAnalyzer::CreateAnalyzerObjects()
         pa::RequestRateManager::Create(
             params_->async, params_->streaming, params_->measurement_window_ms,
             params_->max_trials, params_->request_distribution,
-            params_->batch_size, params_->max_threads,
+            params_->schedule, params_->batch_size, params_->max_threads,
```
matthewkotila, line 245: IMO, this should be a new load manager, could be a subclass of RequestRateManager, which takes a schedule.

We already have CustomLoadManager, which takes a text file with intervals.

This PR is like CustomLoadManager but instead of a file, it takes the times on the command line, and instead of _intervals_ (i.e. time _between_ request sends), it takes absolute timestamps, meaning like

`--schedule=1,2,4.5`

sends a request at 1 second after the start of the benchmark, 2 seconds after benchmark start, and 4.5 seconds after benchmark start.

src/inference_profiler.cc
```diff
@@ -622,12 +622,25 @@ InferenceProfiler::Profile(
...
-                      ->PerformWarmup(request_rate, warmup_request_count));
-  RETURN_IF_ERROR(dynamic_cast<RequestRateManager*>(manager_.get())
-                      ->ChangeRequestRate(request_rate, request_count));
-  std::cout << "Request Rate: " << request_rate
-            << " inference requests per seconds" << std::endl;
+  auto custom_manager =
+      dynamic_cast<CustomRequestScheduleManager*>(manager_.get());
+  auto request_rate_manager = dynamic_cast<RequestRateManager*>(manager_.get());
+
+  if (custom_manager) {
```
nicolasnoble, line 629: This `if` flips between essentially the same code, just with different object types from a common pointer type using `dynamic_cast`. Given that `PerformWarmup` and `ChangeRequestRate` are being made virtuals already in this PR, this shouldn't be needed at all. The only difference lies in the debug message being emitted, which could also just become a virtual function returning a string, such as `GetManagerName`, to embed in a status message such as this one. This means `dynamic_cast` here is completely redundant with the rest of the code.

src/custom_request_schedule_manager.cc
```diff
@@ -0,0 +1,137 @@
...
+        request_parameters)
+{
+  std::unique_ptr<CustomRequestScheduleManager> local_manager(
+      new CustomRequestScheduleManager(
+          async, streaming, measurement_window_ms, max_trials, schedule,
+          batch_size, max_threads, num_of_sequences, shared_memory_type,
+          output_shm_size, serial_sequences, parser, factory,
+          request_parameters));
+
+  *manager = std::move(local_manager);
```
nicolasnoble, line 51: This is technically an anti-pattern. This should (1) use `std::make_unique`, and (2) directly store it straight into the output storage.

nicolasnoble (changes requested):

src/custom_request_schedule_manager.h
```diff
@@ -0,0 +1,140 @@
...
+  /// \param num_of_sequences The number of concurrent sequences to maintain on
+  /// the server \param shared_memory_type The type of shared memory to use for
+  /// inputs \param output_shm_size The size of the shared memory to allocate
+  /// for the output \param serial_sequences Enable serial sequence mode \param
+  /// parser The ModelParser object to get the model details \param factory The
+  /// ClientBackendFactory object used to create client to the server \param
+  /// manager Returns a new CustomRequestScheduleManager object \param
+  /// request_parameters Custom request parameters to send to the server \return
+  /// cb::Error object indicating success or failure
+  static cb::Error Create(
```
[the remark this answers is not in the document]
nicolasnoble: To clarify a bit my stance: I am not against smart pointers in general, just that `shared_ptr` are usually `unique_ptr` in disguise. They are non-committal about the actual ownership of the object, whereas `unique_ptr` has a clearer definition of it.

src/inference_profiler.cc
lkomali: I think dynamic cast is required. `RequestRate` inherits from `LoadManage`r class. These function are specific to `RequestRate` and classes that inherits from it. If I remove dynamic cast, the type for manager would be LoadManager which would throw an error when I try to call the `PerformWarmup` and `ChangeRequestRate` functions.

src/custom_request_schedule_manager.cc
lkomali: This gives an error because the constructor is protected and `std::make_unique` calls the constructor from outside of class scope, so for this to work, we should make constructor public.

src/custom_request_schedule_manager.h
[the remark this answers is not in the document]
lkomali: The shared_ptr is for client backend factory and I have no idea how much of an effort it is to modify everything to unique_ptr. I can create a ticket for that effort.
Coming to the constructor call, I agree with all your points. As far as I understand, your suggestion is to pass params object instead of the individual arguments. It makes sense to me. Again, I'm not sure if I'm going to break something by modifying the API. Let me know if I should add a ticket to investigate and modify the all the related APIs because as you said, it might look awkward if we modify for some of them and it's not propagated until some time after.

src/custom_request_schedule_manager.h
```diff
@@ -0,0 +1,140 @@
...
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#pragma once
+
+#include "load_manager.h"
+#include "request_rate_manager.h"
+
+namespace triton { namespace perfanalyzer {
```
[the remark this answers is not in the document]
lkomali: The general trend followed in all other files is 
`namespace triton { namespace perfanalyzer {`
Should I modify it to `namespace triton::perfanalyzer {` only in this particular file for now or modify it everywhere?

src/inference_profiler.cc
nicolasnoble: Right, this means we only need a single `dynamic_cast` to `RequestRateManager`. We might want to ensure it's not `nullptr` for safety, in case the manager object we get at this point is indeed one of the two types, but otherwise the same remark should apply.

src/custom_request_schedule_manager.cc
nicolasnoble: Fiiiiiiiiine. It's okay this way then.

src/custom_request_schedule_manager.h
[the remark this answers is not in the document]
nicolasnoble: Right, let's just try and plumb the parameters object in this one instance, and keep in mind we should do this for newer code. If we have time, we should go back and clean older API incrementally. You're right: let's not break existing code right now, but let's avoid introducing new code which follows this older pattern.

For the unique_ptr vs shared_ptr, this was more of a general remark that we should keep in mind when adding new shared_ptr in the codebase. At some point we should review the existing usage and clean it up, but not in this one PR.

src/custom_request_schedule_manager.h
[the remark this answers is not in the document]
nicolasnoble: No let's just do it incrementally too. This is new syntax from C++17, and it's cleaner in general. If you spot easy to change locations elsewhere, feel free to do so, but let's work incrementally towards more readable code.

nicolasnoble (approved):
Looks good now, thank you so much!

src/custom_request_schedule_manager.cc
```diff
@@ -0,0 +1,121 @@
...
+
+void
+CustomRequestScheduleManager::GenerateSchedule(
+    const double request_rate, const std::vector<float>& schedule)
+{
+  std::vector<float> scaled_schedule;
+  scaled_schedule.reserve(schedule.size());
+  if (schedule.size() > 0) {
+    for (const auto& value : schedule) {
+      scaled_schedule.push_back(value / static_cast<float>(request_rate));
```
matthewkotila, line 91: @IzzyPutterman why are the values from schedule (e.g. `--schedule=1,2,3,4,4.5,6,9,10,11,13,15,18` being divided by the request rate?

I thought those values are supposed to be absolute timestamps?

cc @lkomali @debermudez

IzzyPutterman: I added a random feature so I could scale by the request-rate if I wanted.  So without needing to change the schedule file I could speedup the sending by 2x or slowdown by 2x.

matthewkotila: I think this should be a new option, it kinda breaks the previously-communicated use of the `--request-rate-range` option.

PR #90, opened 2024-09-19
Throw warning in case of Triton API minor version mismatch

lkomali:
Currently, PA fails if there is a version mismatch between API_VERSION defined in core and client. 
This PR modifies this behavior to 
  1. fail when Triton API major version mismatch
  2. display a warning on the console when Triton API minor version mismatch

The reason for this difference is because major version change denotes breaking changes, minor version denotes additions.


src/client_backend/triton_c_api/triton_loader.cc
```diff
@@ -244,16 +244,25 @@ TritonLoader::StartTriton()
   REPORT_TRITONSERVER_ERROR(
       api_version_fn_(&api_version_major, &api_version_minor),
       "unable to get api version");
-  if ((TRITONSERVER_API_VERSION_MAJOR != api_version_major) ||
-      (TRITONSERVER_API_VERSION_MINOR > api_version_minor)) {
+
+  if (TRITONSERVER_API_VERSION_MAJOR != api_version_major) {
```
ganeshku1, line 248: Can we do something like this?

```
std::string createErrorMessage(const std::string& message, int expected, int actual) {
    return message + "Expected version: " + std::to_string(expected) + "\n" +
           "Actual version: " + std::to_string(actual) + "\n";
}

if (TRITONSERVER_API_VERSION_MAJOR != api_version_major) {
    std::string errorMessage = createErrorMessage(
        "Triton server API major version mismatch.\n",
        TRITONSERVER_API_VERSION_MAJOR, api_version_major
    );
    return Error(errorMessage);
}

if (TRITONSERVER_API_VERSION_MINOR > api_version_minor) {
    std::string warningMessage = createErrorMessage(
        "Warning: Triton server API minor version mismatch.\n",
        TRITONSERVER_API_VERSION_MINOR, api_version_minor
    );
    std::cerr << warningMessage;
}

```

The createErrorMessage function avoids redundant code by consolidating string message construction.
Use of std::string: Instead of std::stringstream, std::string is used with std::to_string for simplicity and efficiency.
Reduced redundancy by eliminating the need for separate std::stringstream instances.

lkomali: I don't have an idea which is better, I used existing. I will consider this, research a bit once and make changes as you suggested.

lkomali:
I missed one of the checks, will add it.

src/client_backend/triton_c_api/triton_loader.cc
matthewkotila, line 248: Add the case for where `TRITONSERVER_API_VERSION_MINOR` is > `api_version_minor`

We wanna fail in that case cuz PA is expecting a later minor version, meaning it could be using an API not found in the older minor version.

src/client_backend/triton_c_api/triton_loader.cc
```diff
@@ -244,16 +244,25 @@ TritonLoader::StartTriton()
...
+            << "Actual major version: " << api_version_major;
     return Error(sstream.str());
   }
+
+  if (TRITONSERVER_API_VERSION_MINOR > api_version_minor) {
+    std::stringstream sstream;
+    sstream << "Warning: Triton server API minor version mismatch. \n"
+            << "Expected minor version: " << TRITONSERVER_API_VERSION_MINOR
+            << "\n"
+            << "Actual minor version: " << api_version_minor;
```
matthewkotila, line 262: Can you add something to message saying attempting to run anyway, undefined behavior may happen?

src/client_backend/triton_c_api/triton_loader.cc
lkomali: Isn't it the other way round? When TRITONSERVER_API_VERSION_MINOR < api_version_minor, we fail it?

src/client_backend/triton_c_api/triton_loader.cc
```diff
@@ -244,16 +244,30 @@ TritonLoader::StartTriton()
...
   if ((TRITONSERVER_API_VERSION_MAJOR != api_version_major) ||
-      (TRITONSERVER_API_VERSION_MINOR > api_version_minor)) {
-    std::stringstream sstream;
-    sstream << "triton server API version mismatch. \n"
-            << "Expected version major:" << TRITONSERVER_API_VERSION_MAJOR
-            << ", minor:" << TRITONSERVER_API_VERSION_MINOR << "\n"
-            << "  Actual version major:" << api_version_major
-            << ", minor:" << api_version_minor;
-    return Error(sstream.str());
+      (api_version_minor > TRITONSERVER_API_VERSION_MINOR)) {
```
dyastremsky, line 255: You're printing an error message for a major version mismatch when there's only a minor version mismatch here, I think. Be careful with the branching. You want separate messages for the three cases. You can also be more specific with the error messages this way too (e.g. state that the server API version is newer than the client version for that case).

lkomali: Oh makes sense!

dyastremsky (dismissed):

src/client_backend/triton_c_api/triton_loader.cc
matthewkotila: `TRITONSERVER_API_VERSION_MINOR` is what PA is compiled to work with.

If it's higher than what it finds in the .so, it could mean it might use a not-yet-existing API.

That's why it's a failure for when `TRITONSERVER_API_VERSION_MINOR` > `api_version_minor`.

If you think this is true, can you provide reasoning?

lkomali:
@matthewkotila and I discussed offline and we decided to throw a warning incase of a minor version mismatch since the updates to Triton C API PA is likely to be rare and the probability of an undefined behavior is pretty low.

src/client_backend/triton_c_api/triton_loader.cc
matthewkotila: Harshini and I discussed offline, makes more sense from UX to only warn for minor mismatch, regardless of direction.

Reason being is that PA has been using the same Triton Core APIs since like 3 years ago when the feature was implemented. It's going to be highly unlikely for someone to be using a version of PA and Triton more than 3 years apart in version. I want to avoid the situation where someone has a server container a few months old and pip installs latest PA and gets an error because of that.

matthewkotila (approved):

dyastremsky (approved):

PR #93, opened 2024-09-20
Refactor non-chat LLM endpoint converters into template style design

nv-hwoo:
This PR
- refactors non-chat based LLM endpoint converters into simpler template-style implementation
- removes chat API specific logics (e.g. system/user roles) from non-chat APIs

This PR does _NOT_
- change any behavior of the overall codebase

### DIFF
--- a/genai-perf/genai_perf/inputs/converters/base_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/base_converter.py
@@ -25,7 +25,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import random
-from typing import Dict, List, Tuple
+from typing import Dict, List
 
 from genai_perf.exceptions import GenAIPerfException
 from genai_perf.inputs.input_constants import ModelSelectionStrategy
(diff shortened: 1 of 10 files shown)


genai-perf/genai_perf/inputs/converters/openai_completions_converter.py
```diff
@@ -25,95 +25,52 @@
...
-            generic_dataset,
-            system_role_headers,
-            user_role_headers,
-            text_input_headers,
-            config,
-        )
+    # TODO (TPA-430): This works for great for synthetic and input file approaches
+    # but a bit tedious for dataset case as we need to specify the content names
+    # for each dataset. This is because a dataset can be used differently depending
+    # on the endpoint (e.g. chat vs non-chat).
```
nv-hwoo, line 40: In order to remove chat API specific logics out from the non-chat APIs, these three non-chat converters need to specify the list of content keys/names it supports. This is mainly due to the complexity of supporting arbitrary public datasets. For example, each dataset comes with unique content names (e.g. "system_prompt", "question", and "response" in openorca). And depending on the endpoint, how we want to construct the payload changes. For chat endpoints, it makes sense to distinguish each contents from openorca dataset because it understands system prompt concept. However, in non-chat LLM API, there's no difference between the "system_prompt" content and "question" content in the openorca dataset, so we just simply concatenate them, but at the same time, we need to _ignore_ "response" content, hence specifying only "system_prompt" and "question" content names under `_CONTENT_NAMES` attribute. 

We probably want more scalable approach if we want to support arbitrary datasets, but that is likely out of scope of this PR. Also, given that the dataset approach is the least used feature in genai-perf, this is probably not an urgent issue.

genai-perf/genai_perf/inputs/converters/base_converter.py
```diff
@@ -54,62 +52,11 @@ def _select_model_name(
...
-                pa_json["data"][index]["text_input"][0] = new_text_input
-
-        return pa_json
+    def _construct_text_payload(self, input_data: Dict) -> str:
+        """
+        Construct text payload content for non-chat based LLM converters.
+        Since there are no roles or turns in non-chat LLM endpoints, all the
+        (pre-defined) text contents are concatenated into a single text prompt.
+        """
+        contents = [v for k, v in input_data.items() if k in self._CONTENT_NAMES]
```
debermudez, line 61: Well this is so much more simple. This all works? 
Crazy good simplification.

debermudez (approved):

debermudez (approved):

PR #96, opened 2024-09-25
Create Optuna Objective Generator Class

nv-braf:
Creates the OptunaObjectiveGenerator class:

- Takes in the baseline metrics (RunConfigMeasurement) and the search space (SearchParameters)
- Outputs a new set of parameters that should be profiled (ObjectiveParameters)

We will then profile using these parameters and pass back in the results (as a RunConfigMeasurement). This will be compared/scored against the baseline and used by Optuna to suggest a new set of parameters to profile. 

In addition I made a number of other (smallish) changes:

- Moved the TypeAliases to a central location (types.py)
- Changed some naming of variables and methods to better reflect what they are doing
- Updated the config file to hold fields needed by this class

NOTE: I use `#FIXME: OPTIMIZE` to indicate code that will be uncommented in the near future as more code gets ported. This can be safely ignored (for now) as these changes are being merged onto a branch and will be addressed before the top-level branch is merged to main.

### DIFF
--- a/genai-perf/genai_perf/config/generate/optuna_objective_generator.py
+++ b/genai-perf/genai_perf/config/generate/optuna_objective_generator.py
@@ -11,16 +11,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 from dataclasses import dataclass
 from random import randint
 from typing import Generator, Optional
 
+import genai_perf.logging as logging
 import optuna
 from genai_perf.config.generate.objective_parameter import ObjectiveParameters
-from genai_perf.config.generate.search_parameters import SearchParameters
 from genai_perf.config.input.config_command import ConfigCommand
 from genai_perf.measurements.run_config_measurement import RunConfigMeasurement
+from genai_perf.types import ModelSearchParameters
+
+logger = logging.getLogger(__name__)
 
 
 @dataclass(frozen=True)
@@ -30,7 +32,7 @@ class OptunaObjectiveGeneratorDefaults:
 
 class OptunaObjectiveGenerator:
     """
-    Generates the next set of objectives to try using Optuna's hyperparameter
+    Generates the next set of objectives to profile using Optuna's hyperparameter
     algorithm
     """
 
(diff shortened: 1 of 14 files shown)


genai-perf/pyproject.toml
```diff
@@ -59,6 +59,7 @@ dependencies = [
   "pyyaml",
   "responses",
   "pillow",
+  "oputna==3.6.1",
```
dyastremsky, line 62: It looks like this version is unable to be found in the GitHub hook. The check is failing due to being unable to build and test as a result.

nv-braf: Yes, I will need to figure this out before we merge to main, but for now I don't think this is a blocking issue (as we're only merging to a branch).

dyastremsky: Sounds good.

genai-perf/genai_perf/config/generate/optuna_objective_generator.py
```diff
@@ -0,0 +1,468 @@
...
+        total_num_of_possible_configs = self._calculate_num_of_configs_in_search_space()
+        min_trials_based_on_percentage_of_search_space = int(
+            total_num_of_possible_configs
+            * self._config.optimize.search_space_percentage.min
+            / 100
+        )
+
+        return min_trials_based_on_percentage_of_search_space
+
+    def _decide_min_between_percentage_and_trial_count(
```
dyastremsky, line 204: In a future refactor, I'd love to see the min/max functions refactored to be one, since they're mostly the same. There could be a flag to indicate whether to use the max or the min.

dyastremsky (approved):
Looks good. Wonderful work, Brian!

nicolasnoble (commented):
Thank you for looping me in these PRs, this definitely gives me context. I am not seeing anything that'd make me want to comment as of now, but of course, this is mostly because I still need to learn about the whole product.

genai-perf/genai_perf/config/generate/optuna_objective_generator.py
nv-braf: I actually gave that some thought because it's almost a copy/paste, but there were enough differences (especially around all the debug) that I couldn't figure out a way to combine them cleanly.

nicolasnoble: If I may give my own two cents, personally, I usually live by the principle that code needs to be refactored only after the third occurrence, not before. Having two almost identical copies of the same code is still acceptable in my book, it's when we reach three that we may want to think about it more.

PR #97, opened 2024-09-25
Improve multithreading safety of TritonLoader

matthewkotila:
Small PR that adds locking around the setting of a data member of a shared instance that could be set by multiple threads.


matthewkotila:
Closed because this is not an exhaustive multithreading safety review of TritonLoader. Might as well add thread safety on a complete review.

PR #98, opened 2024-09-26
Update PA versions post 24.09

pvijayakrish:
Update version reference in PA post 24.09 release.

### DIFF
--- a/templates/template_vars.yaml
+++ b/templates/template_vars.yaml
@@ -1,7 +1,7 @@
 General:
   release: 24.09
   triton_cli_version: 0.0.12
-  genai_perf_version: 0.0.6
+  genai_perf_version: 0.0.7dev
 
 README:
   filename: README.md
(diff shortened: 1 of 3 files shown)


templates/template_vars.yaml
```diff
@@ -1,7 +1,7 @@
 General:
-  release: 24.08
-  triton_cli_version: 0.0.11
-  genai_perf_version: 0.0.6dev
+  release: 24.09
+  triton_cli_version: 0.0.12
```
nv-hwoo, line 3: just to confirm, will triton cli version be bumped to 0.0.12 along with 24.09 release? (I don't see any PR or branch for the change in the triton cli repo)

pvijayakrish: triton_cli_version will be updated later.

nv-hwoo: should we wait for the 24.09 image so we can update both triton cli version and trtllm version at the same time?

pvijayakrish: Also this PR will be merged after 24.09 containers are published.

nv-hwoo (approved):
LGTM

PR #100, opened 2024-09-26
remove dataset support

debermudez:
Removing support for a dataset from huggingface. 
It was growing in complexity to support while not getting a lot of use.


genai-perf/genai_perf/inputs/input_constants.py
matthewkotila, line 1: High level question: this PR mainly just removes code for dataset support.

Does the removal of it now pose an opportunity to organize/refactor code in a cleaner way that wasn't possible before while dataset support was required?

debermudez: It is one less piece that is getting refactored and updated in the new work.
Basically one less retrieval channel we need to support as we are refactoring. 
I think that is a long way of saying yes.

nv-hwoo (commented):
nice 👍 could you also update the README.md as well?

debermudez:
> nice 👍 could you also update the README.md as well?

done

nv-hwoo (approved):
LGTM

genai-perf/README.md
```diff
@@ -407,9 +400,9 @@ URL of the endpoint to target for benchmarking. (default: `None`)
 ##### `--batch-size <int>`
 ##### `--batch-size-text <int>`
 
-The text batch size of the requests GenAI-Perf should send.
+The batch size of the requests GenAI-Perf should send.
```
dyastremsky, line 403: This looks like it was lot during rebase (thanks for rebasing!). Could you return the word "text" here to differentiate from --batch-size-image?

genai-perf/README.md
```diff
@@ -407,9 +400,9 @@ URL of the endpoint to target for benchmarking. (default: `None`)
...
 This is currently only supported with the
-[embeddings](docs/embeddings.md) and
+[embeddings](docs/embeddings.md), image_retrieval, and
```
dyastremsky, line 405: Similarly, image_retrieval does not support text batching, since it only uses images. You can revert to what it was before.

dyastremsky (commented):
Great work! It'll be nice to have one fewer option to support for input sources.

genai-perf/README.md
debermudez: Thank you for catching this. Done

genai-perf/README.md
debermudez: done

dyastremsky (approved):

genai-perf/README.md
```diff
@@ -105,7 +105,9 @@ You can also build Perf Analyzer [from source](../docs/install.md#build-from-sou
 ### Install GenAI-Perf from source
 
 ```bash
-pip install git+https://github.com/triton-inference-server/perf_analyzer.git#subdirectory=genai-perf
+git clone https://github.com/triton-inference-server/perf_analyzer.git && cd perf_analyzer
+
+pip install -e genai-perf
```
nv-hwoo, line 110: why did we split the single command to multiple commands?

debermudez: what do you mean?

nv-hwoo: It was `pip install git+https://github.com/triton-inference-server/perf_analyzer.git#subdirectory=genai-perf` before.

nv-hwoo: now it's 
```
git clone https://github.com/triton-inference-server/perf_analyzer.git && cd perf_analyzer
pip install -e genai-perf
```

debermudez: Its possible that it was a discrepancy with the template vs the readme.
we can revert it if you prefer

nv-hwoo: yes, we want a single pip install command for this. We don't want the users to go through two steps of cloning the repo and then pip install it.

debermudez: Reverted

PR #101, opened 2024-09-26
Refactor embeddings/rankings/tensorrtllm_engine converters to template-syle design

nv-hwoo:
This PR
- refactors the three converters into simpler template-style implementation
- adds unit tests for the resulting request body of the three converters (checks default case and case with request parameters)
- adds optional validation functionality to the `load_json_str` util function. This allows us to check/validate whether the json line contains the proper key name (e.g. `text_input` for LLM endpoints and `text` for embeddings or rankings).
- moved various input validation checks from converter to retriever side. It makes more sense for the retriever to check for these input related conditions, and it's good to place the checks closer to where the variables are defined.


genai-perf/genai_perf/inputs/converters/openai_embeddings_converter.py
```diff
@@ -31,33 +31,25 @@
...
-        config: InputsConfig,
-    ) -> Dict:
-        pa_json: Dict[str, Any] = {"data": []}
 
-        for index, entry in enumerate(generic_dataset["rows"]):
-            iter_model_name = self._select_model_name(config, index)
-            payload = entry.get("payload", {})
-            input_values = payload.get("input")
+    def convert(self, generic_dataset: Dict, config: InputsConfig) -> Dict:
+        """
```
dyastremsky, line 36: Do we want to have these sorts of function descriptions in the base converter instead of copying them into each converter? Open either way on this, with a slight preference for just having the shared function descriptions in the base class.

genai-perf/genai_perf/parser.py
```diff
@@ -135,6 +135,9 @@ def _check_conditional_args(
     Check for conditional args and raise an error if they are not set.
     """
 
+    if args.num_prompts <= 0:
```
dyastremsky, line 138: Thanks for adding this check. Can you use ArgParse instead to make sure this arg is positive?

genai-perf/genai_perf/utils.py
```diff
@@ -74,9 +74,19 @@ def load_json(filepath: Path) -> Dict[str, Any]:
         return load_json_str(content)
 
 
-def load_json_str(json_str: str) -> Dict[str, Any]:
+def load_json_str(json_str: str, func: Callable = lambda x: x) -> Dict[str, Any]:
+    """
+    Deserealizes JSON encoded string into Python object.
```
dyastremsky, line 79:
```suggestion
    Deserializes JSON encoded string into Python object.
```

genai-perf/tests/test_rankings_converter.py
```diff
@@ -0,0 +1,196 @@
+# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
dyastremsky, line 1: If we're going to start making separate test files for each converter, we should see if we can have a subdirectory for the tests for the converters (same for the retriever tests). This can be in a separate PR, if necessary.

dyastremsky (commented):
Great work! Appreciate all the refactoring and clean-up you did here.

genai-perf/tests/test_rankings_converter.py
debermudez: I think we are going to have a few PRs soon that reorganize the file hierarchy.
I agree but i would prefer to have them be separate.

debermudez (commented):
This is SO much more readable.
Fantastic work!

genai-perf/genai_perf/inputs/converters/openai_embeddings_converter.py
nv-hwoo: removed the docstrings for the derived classes

genai-perf/genai_perf/parser.py
nv-hwoo: modified to use argparse for checking input values

genai-perf/genai_perf/utils.py
nv-hwoo: done

dyastremsky (approved):

genai-perf/tests/test_rankings_converter.py
nv-hwoo: will address this in a separate PR

PR #102, opened 2024-09-27
Add ability to enable code execution for custom tokenizers

dyastremsky:
Add the ability to execute code for custom tokenizers. This is necessary for custom tokenizers hosted on HuggingFace Hub as well as local tokenizers that require it to be enabled.

As part of these changes, I also added a flag to use a specific model tag. This is often necessary for production code to ensure you are using an approved version of a model. It is paramount for those users who would enable trusting remote code, as they may want a specific model version for enhanced security.

### DIFF
--- a/genai-perf/README.md
+++ b/genai-perf/README.md
@@ -522,16 +522,16 @@ The HuggingFace tokenizer to use to interpret token metrics from prompts and
 responses. The value can be the name of a tokenizer or the filepath of the
 tokenizer. (default: `hf-internal-testing/llama-tokenizer`)
 
+##### `--tokenizer-revision <str>`
+
+The specific tokenizer model version to use. It can be a branch
+name, tag name, or commit ID. (default: `main`)
+
 ##### `--tokenizer-trust-remote-code`
 
 Allow custom tokenizer to be downloaded and executed. This carries security
 risks and should only be used for repositories you trust. This is only
 necessary for custom tokenizers stored in HuggingFace Hub.  (default: `False`)
-
-##### `--tokenizer-revision <str>`
-
-The specific tokenizer model version to use. It can be a branch
-name, tag name, or commit ID. (default: `main`)
 
 ##### `-v`
 ##### `--verbose`
--- a/genai-perf/README.md
+++ b/genai-perf/README.md
@@ -522,6 +522,17 @@ The HuggingFace tokenizer to use to interpret token metrics from prompts and
 responses. The value can be the name of a tokenizer or the filepath of the
 tokenizer. (default: `hf-internal-testing/llama-tokenizer`)
 
+##### `--tokenizer-trust-remote-code`
+
+Allow custom tokenizer to be downloaded and executed. This carries security
+risks and should only be used for repositories you trust. This is only
+necessary for custom tokenizers stored in HuggingFace Hub.  (default: `False`)
+
+##### `--tokenizer-revision <str>`
+
+The specific tokenizer model version to use. It can be a branch
+name, tag name, or commit ID. (default: `main`)
+
 ##### `-v`
 ##### `--verbose`
 
--- a/genai-perf/genai_perf/main.py
+++ b/genai-perf/genai_perf/main.py
@@ -95,7 +95,9 @@ def create_config_options(args: Namespace) -> InputsConfig:
         random_seed=args.random_seed,
         num_prompts=args.num_prompts,
         add_stream=args.streaming,
-        tokenizer=get_tokenizer(args.tokenizer),
+        tokenizer=get_tokenizer(
+            args.tokenizer, args.tokenizer_trust_remote_code, args.tokenizer_revision
+        ),
         extra_inputs=extra_input_dict,
         batch_size=args.batch_size,
         output_dir=args.artifact_dir,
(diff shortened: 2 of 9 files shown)


genai-perf/README.md
```diff
@@ -522,6 +522,17 @@ The HuggingFace tokenizer to use to interpret token metrics from prompts and
...
 tokenizer. (default: `hf-internal-testing/llama-tokenizer`)
 
+##### `--tokenizer-revision <str>`
+
+The specific tokenizer model version to use. It can be a branch
+name, tag name, or commit ID. (default: `main`)
+
+##### `--tokenizer-trust-remote-code`
+
+Allow custom tokenizer to be downloaded and executed. This carries security
```
nv-hwoo, line 532: since it carries a security risk for the user, can we add a short warning log to the user when this flag has been set?

genai-perf/tests/test_tokenizer.py
```diff
@@ -38,6 +42,16 @@ def test_non_default_tokenizer(self):
...
         get_tokenizer(tokenizer_model)
 
+    def test_default_tokenizer_all_args(self):
+        tokenizer_model = DEFAULT_TOKENIZER
+        get_tokenizer(tokenizer_model, False, DEFAULT_TOKENIZER_REVISION)
+
+    def test_non_default_tokenizer_all_args(self):
+        tokenizer_model = "gpt2"
+        get_tokenizer(
+            tokenizer_model, False, "11c5a3d5811f50298f278a704980280950aedb10"
```
nv-hwoo, line 52: is it possible to check whether the tokenizer attributes are correctly set? 

e.g. `tokenizer.trust_remote_code == True`

genai-perf/README.md
dyastremsky: Great idea. Done!

genai-perf/tests/test_tokenizer.py
dyastremsky: Unfortunately, there doesn't seem to be a getter to run this check. The main way to test this would be to have a tokenizer locally or on Hugging Face that requires this to be set to true but then it wouldn't be a unit test (and would expose the test to some security risks). We could alternatively mock the initialization call to ensure it's called with the right args, but then we're just redoing what the function does.

genai-perf/genai_perf/main.py
```diff
@@ -193,7 +195,13 @@ def run():
...
     else:
         create_artifacts_dirs(args)
-        tokenizer = get_tokenizer(args.tokenizer)
+        tokenizer = (
+            get_tokenizer(
+                args.tokenizer,
+                args.tokenizer_trust_remote_code,
+                args.tokenizer_revision,
+            ),
+        )
```
nv-hwoo, line 204: forgot to add this to the previous review
```suggestion
        tokenizer = get_tokenizer(
            args.tokenizer,
            args.tokenizer_trust_remote_code,
            args.tokenizer_revision,
        )
```

genai-perf/tests/test_tokenizer.py
nv-hwoo: sounds good. yeah I was hoping if that was possible but I guess there's no simple way to test this... 

if there's no way to check the changes, is there any point in having these new tests?

dyastremsky: We'd be testing that the external code doesn't change and that this function doesn't change. These tests would break if:
- The function signature changes
- The HuggingFace tokenizer parameters change
So this would catch a regression but isn't strictly necessary. These are similar to the tests right above them. If you'd like though, I can remove them.

genai-perf/genai_perf/main.py
debermudez: this is probably fall out from the formatting.

dyastremsky: Good catch, fixed!

genai-perf/genai_perf/tokenizer.py
```diff
@@ -44,7 +45,9 @@ def __init__(self, name: str) -> None:
             with contextlib.redirect_stdout(
                 io.StringIO()
             ) as stdout, contextlib.redirect_stderr(io.StringIO()) as stderr:
-                tokenizer = AutoTokenizer.from_pretrained(name)
+                tokenizer = AutoTokenizer.from_pretrained(
```
debermudez, line 48: Nice!
I didnt know this.

debermudez (commented):
Final approval lies with @nv-hwoo on this one  but i think this is good to go

genai-perf/tests/test_tokenizer.py
nv-hwoo: I see. Yeah I think passing non default args could reveal if anything changes on the huggingface side. Instead of removing, can we just pass `True` so that we are passing non default for all arguments?
```python
get_tokenizer(
    tokenizer_model, True, "11c5a3d5811f50298f278a704980280950aedb10"
```

dyastremsky: If we pass True, we expose our GitLab runners to a security risk. The Tokenizer model should be safe with the hash there, but I didn't want to add that risk.

nv-hwoo: makes sense. we can keep it as is

nv-hwoo (approved):
LGTM 🚀

PR #104, opened 2024-09-27
Create Sweep Objective Generator

nv-braf:
This is the generator that will be used for analyze/sweep. It is most similar to the BruteRCG in MA.

It will exhaustively sweep the entire objective space creating one set of parameters to profile per call to the generator. Since this was so simple I've already added multi-model (think BLS/ensemble) support and testing.

### DIFF
--- a/genai-perf/genai_perf/config/generate/search_parameter.py
+++ b/genai-perf/genai_perf/config/generate/search_parameter.py
@@ -62,5 +62,5 @@ class SearchParameter:
             or self.category == SearchCategory.EXPONENTIAL
         ):
             return [value for value in range(self.min_range, self.max_range + 1)]  # type: ignore
-        else:
-            return []
+
+        assert True, f"{self.category} is not a known SearchCategory"
--- a/genai-perf/genai_perf/config/generate/search_parameter.py
+++ b/genai-perf/genai_perf/config/generate/search_parameter.py
@@ -47,3 +47,18 @@ class SearchParameter:
     # These are only applicable to INTEGER and EXPONENTIAL categories
     min_range: Optional[int] = None
     max_range: Optional[int] = None
+
+    def get_list(self) -> List[Any]:
+        """
+        Returns the list of all possible parameter values
+        """
+        if (
+            self.category == SearchCategory.STR_LIST
+            or self.category == SearchCategory.INT_LIST
+        ):
+            return self.enumerated_list  # type: ignore
+        elif (
+            self.category == SearchCategory.INTEGER
+            or self.category == SearchCategory.EXPONENTIAL
+        ):
+            return [value for value in range(self.min_range, self.max_range + 1)]  # type: ignore
--- a/genai-perf/genai_perf/config/generate/search_parameter.py
+++ b/genai-perf/genai_perf/config/generate/search_parameter.py
@@ -62,3 +62,5 @@ class SearchParameter:
             or self.category == SearchCategory.EXPONENTIAL
         ):
             return [value for value in range(self.min_range, self.max_range + 1)]  # type: ignore
+        else:
+            return []
--- a/genai-perf/genai_perf/config/generate/sweep_objective_generator.py
+++ b/genai-perf/genai_perf/config/generate/sweep_objective_generator.py
@@ -11,18 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from dataclasses import dataclass
 from itertools import product
 from typing import Any, Dict, Generator, List, TypeAlias
 
 import genai_perf.logging as logging
-from genai_perf.config.generate.objective_parameter import (
-    ObjectiveParameter,
-    ObjectiveParameters,
-)
-from genai_perf.config.generate.search_parameter import SearchCategory, SearchParameter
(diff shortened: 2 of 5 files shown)


genai-perf/genai_perf/config/generate/sweep_objective_generator.py
```diff
@@ -0,0 +1,187 @@
...
+                model_name
+            ].number_of_total_possible_configurations()
+
+        return num_of_configs_in_search_space
+
+    ###########################################################################
+    # Info/Debug Methods
+    ###########################################################################
+    def _print_debug_search_space_info(self) -> None:
+        logger.info("")
```
dyastremsky, line 167: Is this a debugging statement that was forgotten, or why do you need this?

nv-braf: Are you referring to the `logger.info("")`? If so, those are intentional and create newlines around the debug info.

nv-braf: The reason `info` (instead of `debug`) is used is that I wanted an actual newline vs. a line that was `DEBUG: `

dyastremsky: Could you use a newline in the quote below instead so that it's clearer what is being done? e.g.
```
num_of_configs_in_search_space = (
            self._calculate_num_of_configs_in_search_space()
        )
        logger.debug(
            f"\nNumber of configs in search space: {num_of_configs_in_search_space}"
        )
        self._print_debug_model_search_space_info()
```

And perhaps have another newline printed in the debug function? I guess I'm not sure why we want this. It seems misleading to me in terms of reading code and logs. Do you happen to have a screenshot of the logs (with/without these newlines) to help me understand the purpose here?

nv-braf: I don't have a screenshot anymore, but with your suggestion it looks like:
`DEBUG:`
`DEBUG: <config_info>`
`DEBUG:`

versus with the code as is the newlines are actual newlines w/o printing the `DEBUG:` which to my eye just looks cleaner.

nv-braf: Found an old PR from MA that highlights this: https://github.com/triton-inference-server/model_analyzer/pull/900
image

dyastremsky: Got it, thanks for explaining and providing a visual! I agree that this looks cleaner. Resolving this thread.

dyastremsky (approved):
Excellent work, Brian!

genai-perf/genai_perf/config/generate/search_parameter.py
```diff
@@ -47,3 +47,20 @@ class SearchParameter:
...
+            or self.category == SearchCategory.INT_LIST
+        ):
+            return self.enumerated_list  # type: ignore
+        elif (
+            self.category == SearchCategory.INTEGER
+            or self.category == SearchCategory.EXPONENTIAL
+        ):
+            return [value for value in range(self.min_range, self.max_range + 1)]  # type: ignore
+        else:
+            return []
```
nicolasnoble, line 66: If you wanted to silence the warning above, you could just remove the `else:` and simply have a `return []` at the end. Not really an important action item tho.

nicolasnoble (approved):

genai-perf/genai_perf/config/generate/search_parameter.py
nv-braf: The `#type: ignore` is due to the fact that min/max_range is `Optional[int]` and static checking complains since it can't understand that in this case I know that these values will not be none.

`genai-perf/genai_perf/config/generate/search_parameter.py:64: error: Argument 1 to "range" has incompatible type "int | None"; expected "SupportsIndex"  [arg-type]`

nicolasnoble: Right, I was talking about this warning here:

> Explicit returns mixed with implicit (fall through) returns
> Mixing implicit and explicit returns may indicate an error as implicit returns always return None.

It's not able to understand that all paths of that if/elseif/else end up with a return, and it'd prefer a non-conditional explicit return statement.

nv-braf: I can remove all the `#type: ignore` by changing the function to look like:

```
    def get_list(self) -> List[Any]:
        """
        Returns the list of all possible parameter values
        """
        if (
            self.category == SearchCategory.STR_LIST
            or self.category == SearchCategory.INT_LIST
        ):
            assert self.enumerated_list is not None
            return self.enumerated_list
        elif (
            self.category == SearchCategory.INTEGER
            or self.category == SearchCategory.EXPONENTIAL
        ):
            assert self.min_range is not None and self.max_range is not None
            return [value for value in range(self.min_range, self.max_range + 1)]

        assert True  # should never reach here
```

Do you think this is cleaner?

nicolasnoble: I mean I think we're talking about two different warnings here. I'm totally fine either way honestly anyway :)

nv-braf: I've changed the method to assert. I think this is cleaner and will help root cause an issue faster if things go off the rails and we somehow have an unknown SearchCategory.

Thanks for pointing this out!

PR #108, opened 2024-10-02
Print cURL command in PA

lkomali:
This PR modifies openai_client to print cURL command in extra verbose mode in PA for debugging purposes.

image

### DIFF
--- a/src/client_backend/openai/http_client.h
+++ b/src/client_backend/openai/http_client.h
@@ -118,6 +118,11 @@ class HttpRequest {
   // be valid during the transfer and can be freed once transfer is completed.
   struct curl_slist* header_list_{nullptr};
 
+  const std::deque<std::pair<uint8_t*, size_t>>& GetDataBuffers() const
+  {
+    return data_buffers_;
+  }
+
  protected:
   const bool verbose_{false};
 
@@ -142,7 +147,8 @@ class HttpClient {
   // Note that this function does not block
   void Send(CURL* handle, std::unique_ptr<HttpRequest>&& request);
 
- protected:
+  void PrintCurlCommand(CURL* handle, std::unique_ptr<HttpRequest>&& request);
+
   void AsyncTransfer();
 
   bool exiting_{false};
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -295,6 +295,7 @@ add_executable(
   test_ctx_id_tracker.cc
   test_profile_data_collector.cc
   test_profile_data_exporter.cc
+  test_http_client.cc
 )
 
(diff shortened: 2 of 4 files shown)


dyastremsky:
Do we not want the HTTP client to also print the cURL command?

src/client_backend/openai/http_client.h
```diff
@@ -142,7 +147,6 @@ class HttpClient {
   // Note that this function does not block
   void Send(CURL* handle, std::unique_ptr<HttpRequest>&& request);
 
- protected:
```
dyastremsky, line 145: Is there a reason we no longer want this information to be protected?

dyastremsky (commented):
Great work on this PR! Nice job creating the new test as well.

Left a couple of questions.

src/client_backend/openai/http_client.h
nicolasnoble: I'm guessing it's to access `total_input_byte_size_` from `ChatCompletionClient::PrintCurlCommand`. It'd probably be better to add a public accessor for it instead.

lkomali:
@dyastremsky That's what @nv-hwoo raised in the slack thread, as to why we want to print the command twice. 
So I retained the Print function in the openai_client.

src/client_backend/openai/http_client.h
lkomali: Aren't they in protected ? We have defined at line 140 and all the others that follow are protected right? 
I removed thinking it's a redundant declaration.

nicolasnoble: Oh, derp, totally missed the fact this `protected:` here is redundant. Proceed ^_^

lkomali: Is it why I removed the line 145? I think it's redundant and that's why I removed. we have one protected in line 140 and everything that follows till line 171 is protected?

nicolasnoble: Yep, our bad, we just haven't been paying attention.

nv-hwoo:
@lkomali could you add the output of the curl command (e.g. screenshot) to the PR? Would like to see the final look.

lkomali:
@nv-hwoo Attached the screenshot in the description.

dyastremsky:
> @dyastremsky That's what @nv-hwoo raised in the slack thread, as to why we want to print the command twice. So I retained the Print function in the openai_client.

OpenAI isn't the only client we use though. We also use the HTTP client. Wouldn't users want to see the cURL command for the HTTP client too? A solution that allows the cURL command to be printed for all of the relevant clients would be best.

dyastremsky (approved):

PR #109, opened 2024-10-02
Use generic retriever functions for embeddings endpoint

dyastremsky:
This PR removes the need for a custom embeddings file retriever by using the generalized retrievers. The ultimate goal is to not have any endpoint-specific retrievers. GenAI-Perf should eventually have retrievers solely based on the data source, not the output format.

Based on feedback, this PR also allows users to specify text in an input file with both "text" and "text_input" for now, to avoid breaking embeddings/rankings (text) and the other endpoints (text_input).

### DIFF
--- a/genai-perf/docs/embeddings.md
+++ b/genai-perf/docs/embeddings.md
@@ -36,18 +36,18 @@ GenAI-Perf allows you to profile embedding models running on an
 To create a sample embeddings input file, use the following command:
 
 ```bash
-echo '{"text": "What was the first car ever driven?"}
-{"text": "Who served as the 5th President of the United States of America?"}
-{"text": "Is the Sydney Opera House located in Australia?"}
-{"text": "In what state did they film Shrek 2?"}' > embeddings.jsonl
+echo '{"text_input": "What was the first car ever driven?"}
+{"text_input": "Who served as the 5th President of the United States of America?"}
+{"text_input": "Is the Sydney Opera House located in Australia?"}
+{"text_input": "In what state did they film Shrek 2?"}' > embeddings.jsonl
 ```
 
 This will generate a file named embeddings.jsonl with the following content:
 ```jsonl
-{"text": "What was the first car ever driven?"}
-{"text": "Who served as the 5th President of the United States of America?"}
-{"text": "Is the Sydney Opera House located in Australia?"}
-{"text": "In what state did they film Shrek 2?"}
+{"text_input": "What was the first car ever driven?"}
+{"text_input": "Who served as the 5th President of the United States of America?"}
+{"text_input": "Is the Sydney Opera House located in Australia?"}
+{"text_input": "In what state did they film Shrek 2?"}
 ```
 
 ## Start an OpenAI Embeddings-Compatible Server
--- a/genai-perf/genai_perf/inputs/converters/base_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/base_converter.py
@@ -64,7 +64,7 @@ class BaseConverter:
             return self._construct_text_payload(input_data)
         else:
             input_data = cast(List, input_data)
-            return self._construct_texts_payload(input_data)
+            return self._construct_batched_text_payload(input_data)
 
     def _construct_text_payload(self, input_data: Dict) -> str:
         """
@@ -75,7 +75,7 @@ class BaseConverter:
         contents = [v for k, v in input_data.items() if k in self._CONTENT_NAMES]
         return " ".join(contents)
 
-    def _construct_texts_payload(self, input_data: List) -> List:
+    def _construct_batched_text_payload(self, input_data: List) -> List:
         """
         Construct batched text payload content for non-chat based LLM converters.
         """
--- a/genai-perf/genai_perf/inputs/converters/base_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/base_converter.py
@@ -25,7 +25,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import random
-from typing import Dict, List
+from typing import Dict, List, Union, cast
 
 from genai_perf.exceptions import GenAIPerfException
 from genai_perf.inputs.input_constants import ModelSelectionStrategy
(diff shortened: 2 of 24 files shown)


genai-perf/genai_perf/inputs/converters/base_converter.py
```diff
@@ -52,6 +52,20 @@ def _select_model_name(self, config: InputsConfig, index: int) -> str:
                 f"Model selection strategy '{config.model_selection_strategy}' is unsupported"
             )
 
+    def _construct_text_payload_batch_agnostic(
```
dyastremsky, line 55: Note that we currently return different schemas in the retrievers when client-side batching is enabled (batch_size >= 2), which is why this function is necessary.

In a future refactor ticket, I may look at using the same schema (Dicts/Lists) for both. Tthat would need to be done after a redesign of image client-side batching. I think the current client-side batching approach batches images and texts equally, so we'd need to disambiguate the two first. Happy to have this conversation offline.

genai-perf/docs/embeddings.md
```diff
@@ -36,18 +36,18 @@ GenAI-Perf allows you to profile embedding models running on an
 To create a sample embeddings input file, use the following command:
 
 ```bash
-echo '{"text": "What was the first car ever driven?"}
-{"text": "Who served as the 5th President of the United States of America?"}
-{"text": "Is the Sydney Opera House located in Australia?"}
-{"text": "In what state did they film Shrek 2?"}' > embeddings.jsonl
+echo '{"text_input": "What was the first car ever driven?"}
```
nicolasnoble, line 39: Doesn't this technically constitute a breaking change in the tool? What's the general policy regarding breaking changes in general?

genai-perf/genai_perf/inputs/input_retriever_factory.py
```diff
@@ -78,12 +78,6 @@ def get_input_data(self) -> Dict:
             )
         else:
             if self.config.input_type == PromptSource.DATASET:
-                # (TMA-1990) support VLM input from public dataset
```
dyastremsky, line 81: Moved this check to the input type check function in inputs.py. We can start moving all similar checks (output_format vs input_type) there to centralize those checks.

genai-perf/docs/embeddings.md
dyastremsky: Great question. The tool is still in alpha, or unofficially beta. We have had talks to go GA. We unfortunately are still releasing some breaking changes regularly (e.g. changing CLI args), though we are trying to start getting that down to 0.

I'll need to discuss this with @debermudez and the team. We needed to publish embeddings and rankings support in about a week, and we ended up deviating the way the other endpoints support text versus this endpoint. We could modify the retrievers to support both text and text_input as equivalent fields if we wanted to include this. I think it'd still be good to update documentation to just use one approach so that we can eventually deprecate the other.

@debermudez @ganeshku1 @nv-hwoo @nicolasnoble What are your thoughts on supporting the use of both text and text_input as equivalent fields for now but updating the documentation to use text_input everywhere?

dyastremsky: Alternatively, text is shorter, so we could always support _text_ everywhere. I'm open. Just would prefer it be the same everywhere for clarity (and cleaner code later).

nicolasnoble: The excuse of the tool not being GA is a perfectly valid one to change things like this, to be fair :-) I just don't know if we have an expectation of notifying users of changes like this one.

dyastremsky: We usually add breaking changes to the release notes. For now, I added support for both text and text_input, so those using the older format can continue doing so. If we want to deprecate one of the two, we can make that decision when we go GA.

debermudez: This feels like the right approach to me.
That way we arent breaking users for now, and gives us some flexibility

debermudez (approved):

genai-perf/genai_perf/inputs/input_retriever_factory.py
dyastremsky: Thanks for the heads up! I can remove this if there ends up being a conflict.

genai-perf/docs/embeddings.md
nv-hwoo: > What are your thoughts on supporting the use of both text and text_input as equivalent fields for now but updating the documentation to use text_input everywhere?

I personally prefer `text` over `text_input` just because it's simpler and for better consistency with other inputs (e.g. users define image as `image`, not `image_input`).

genai-perf/genai_perf/inputs/converters/base_converter.py
```diff
@@ -60,3 +74,10 @@ def _construct_text_payload(self, input_data: Dict) -> str:
         """
         contents = [v for k, v in input_data.items() if k in self._CONTENT_NAMES]
         return " ".join(contents)
+
+    def _construct_texts_payload(self, input_data: List) -> List:
```
nv-hwoo, line 78: can we change the function name to something like `_construct_batched_text_payload` for better distinguish from the function above?

genai-perf/genai_perf/inputs/file_input_retriever.py
```diff
@@ -182,9 +157,16 @@ def _get_prompts_from_input_file(self) -> Tuple[List[str], List[str]]:
...
+                    data = load_json_str(line)
                     # None if not provided
-                    prompt = load_json_str(line).get("text_input")
-                    image = load_json_str(line).get("image")
+                    prompt = data.get("text_input")
+                    prompt_alt = data.get("text")
+                    if prompt and prompt_alt:
+                        raise ValueError(
+                            "Each data entry must have only one of 'text_input' or 'text' key name."
+                        )
```
nv-hwoo, line 167: I modified `load_json_str` to perform checks like this. You can do something like:
```python
def __check_keyname(x, keys):
    if "text_input" in x and "text" in x:
        raise ValueError(
            "Each data entry must have only one of 'text_input' or 'text' key name."
         )

data = load_json_str(line, func=__check_keyname)
```

dyastremsky: @nv-hwoo What's the benefit of doing it this way? It looks to have the same outcome but add an extra function definition.

genai-perf/docs/embeddings.md
dyastremsky: We can switch to text. I'd do that in its own PR, as changing `text_input` to `text` everywhere has a lot of points at which it can break code, so I want it to be individually reviewed.

Also, we should update parser.py and README at some point. It looks like --input-file needs to be updated to show that it accepts both text and image fields.

genai-perf/genai_perf/inputs/converters/base_converter.py
dyastremsky: Sure, done.

genai-perf/genai_perf/inputs/file_input_retriever.py
nv-hwoo: I just think it's more explicit (and as a result more clear) in terms of what we are checking. The function is directly checking whether the keywords are in the json dictionary, whereas the current if statement is checking the outcome of the `get` method.

It's a nit and a suggestion, so no need to be gated by this.

nv-hwoo (approved):
LGTM

genai-perf/genai_perf/inputs/file_input_retriever.py
dyastremsky: Thanks for explaining. I'll merge this now, but if we want to think it's clearer in the future, then we can change the approach.

PR #110, opened 2024-10-02
Create PA Config Class

nv-braf:
This class takes in options from the user (via the config) and the Objective generator output and creates the CLI command that will be used to run PA. 

It also has a method to create a unique representation of the PA command which will be used in checkpointing to recognize when a configuration has already been profiled.

### DIFF
--- a/genai-perf/genai_perf/config/generate/objective_parameter.py
+++ b/genai-perf/genai_perf/config/generate/objective_parameter.py
@@ -47,5 +47,4 @@ class ObjectiveParameter:
         elif self.category == ObjectiveCategory.EXPONENTIAL:
             return 2**self.value
 
-        assert True, f"{self.category} is not a known ObjectiveCategory"
-        return None
+        raise ValueError(f"{self.category} is not a known ObjectiveCategory")
--- a/genai-perf/genai_perf/config/generate/objective_parameter.py
+++ b/genai-perf/genai_perf/config/generate/objective_parameter.py
@@ -17,6 +17,7 @@ from enum import Enum, auto
 from typing import Any, Dict, TypeAlias
 
 from genai_perf.config.generate.search_parameter import SearchUsage
+from genai_perf.exceptions import GenAIPerfException
 
 
 class ObjectiveCategory(Enum):
@@ -47,4 +48,4 @@ class ObjectiveParameter:
         elif self.category == ObjectiveCategory.EXPONENTIAL:
             return 2**self.value
 
-        raise ValueError(f"{self.category} is not a known ObjectiveCategory")
+        raise GenAIPerfException(f"{self.category} is not a known ObjectiveCategory")
--- a/genai-perf/genai_perf/config/generate/objective_parameter.py
+++ b/genai-perf/genai_perf/config/generate/objective_parameter.py
@@ -48,3 +48,4 @@ class ObjectiveParameter:
             return 2**self.value
 
         assert True, f"{self.category} is not a known ObjectiveCategory"
+        return None
--- a/genai-perf/genai_perf/config/generate/objective_parameter.py
+++ b/genai-perf/genai_perf/config/generate/objective_parameter.py
@@ -37,3 +37,14 @@ class ObjectiveParameter:
     usage: SearchUsage
     category: ObjectiveCategory
     value: Any
+
+    def get_value_based_on_category(self) -> Any:
+        if (
+            self.category == ObjectiveCategory.INTEGER
+            or self.category == ObjectiveCategory.STR
+        ):
+            return self.value
+        elif self.category == ObjectiveCategory.EXPONENTIAL:
+            return 2**self.value
+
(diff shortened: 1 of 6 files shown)


genai-perf/genai_perf/config/generate/objective_parameter.py
```diff
@@ -37,3 +37,14 @@ class ObjectiveParameter:
...
+    def get_value_based_on_category(self) -> Any:
+        if (
+            self.category == ObjectiveCategory.INTEGER
+            or self.category == ObjectiveCategory.STR
+        ):
+            return self.value
+        elif self.category == ObjectiveCategory.EXPONENTIAL:
+            return 2**self.value
+
+        assert True, f"{self.category} is not a known ObjectiveCategory"
```
nicolasnoble, line 50: Feels like this could be a bit more idiomatic if this was:

```python
raise ValueError(f"{self.category} is not a known ObjectiveCategory"
```

instead

nv-braf: Good suggestion!

genai-perf/genai_perf/config/generate/perf_analyzer_config.py
```diff
@@ -0,0 +1,146 @@
...
+from typing import Any, Dict, List
+
+from genai_perf.config.generate.search_parameter import SearchUsage
+from genai_perf.config.input.config_command import ConfigCommand, ConfigPerfAnalyzer
+from genai_perf.types import ModelName, ModelObjectiveParameters
+
+
+class PerfAnalyzerConfig:
+    """
+    Contains the all the methods necessary for handling calls
```
dyastremsky, line 24: Nit: extra words here

genai-perf/genai_perf/config/generate/perf_analyzer_config.py
```diff
@@ -0,0 +1,146 @@
...
+    ):
+        self._model_name = model_name
+        self._set_options_based_on_config(config)
+        self._set_options_based_on_objective(model_objective_parameters)
+
+    ###########################################################################
+    # Set Options Methods
+    ###########################################################################
+    def _set_options_based_on_config(self, config: ConfigCommand) -> None:
+        self._config: ConfigPerfAnalyzer = config.perf_analzyer
```
dyastremsky, line 42: Nit: typo for Perf Analyzer.

genai-perf/genai_perf/config/generate/perf_analyzer_config.py
```diff
@@ -0,0 +1,146 @@
...
+    ) -> str:
+        cli_str_tokens = cli_string.split(" ")
+
+        try:
+            removal_index = cli_str_tokens.index(option_to_remove)
+            cli_str_tokens.pop(removal_index)
+
+            if with_arg:
+                cli_str_tokens.pop(removal_index)
+        except ValueError:
```
dyastremsky, line 142: What is this meant to catch? Could this cause a silent error?

genai-perf/genai_perf/config/generate/perf_analyzer_config.py
```diff
@@ -0,0 +1,146 @@
...
+        return parameter_args
+
+    def _convert_objective_to_cli_option(self, objective_name: str) -> str:
+        obj_to_cli_dict = {
+            "runtime_batch_size": "--batch-size",
+            "concurrency": "--concurrency-range",
+            "request-rate": "--request-rate-range",
+        }
+
+        return obj_to_cli_dict[objective_name]
```
dyastremsky, line 93: You might want to use error handling here if it is possible that the objective name is ever not in the dictionary.

nv-braf: Good suggestion!

genai-perf/genai_perf/config/generate/perf_analyzer_config.py
nv-braf: It's so if the option we are trying to remove doesn't exist in the string we can just ignore and return the string unmodified.
I've updated the comment to better reflect this.

dyastremsky: Would you be able to do this without a try-except by checking if the index is invalid first? And in the with_arg case, checking the index is valid before calling pop (or calling another function that doesn't cause an error if the index is invalid)?

I think try/excepts are best to avoid when possible (e.g. they are sometimes necessary for external code or non-deterministic scenarios). They can obfuscate errors and the except blocks are computationally expensive when executed.

nv-braf: Not a problem. I've replaced the try-except with a check to see if the substring is present.

dyastremsky: Great work!

dyastremsky (approved):

PR #111, opened 2024-10-02
Create separate batch size option for text and images

dyastremsky:
A user should be able to select different batch sizes for different types of inputs. Currently, that includes text and images.

This PR adds support for selecting a batch size for images while avoiding breaking the existing CLI arg for selecting the batch size for text.


genai-perf/README.md
```diff
@@ -407,13 +407,21 @@ URL of the endpoint to target for benchmarking. (default: `None`)
...
 
-The batch size of the requests GenAI-Perf should send.
+The text batch size of the requests GenAI-Perf should send.
 This is currently only supported with the
-[embeddings](docs/embeddings.md), image_retrieval, and
+[embeddings](docs/embeddings.md) and
 [rankings](docs/rankings.md) endpoint types.
 (default: `1`)
 
+##### `--batch-size-images <int>`
```
debermudez, line 418: nit: do we want to alphabetize these? image then text?

debermudez (approved):

nv-braf:
I'd like to better understand the reason for this change. Is there a case where there are both types of inputs (text and images) at the same time? And, therefore we need to allow for the differentiation between these batch sizes?

dyastremsky:
> I'd like to better understand the reason for this change. Is there a case where there are both types of inputs (text and images) at the same time? And, therefore we need to allow for the differentiation between these batch sizes?

For your first question, yes: Vision Language Models (do not yet support client-side batching). For your second question, not currently. With the customizable frontend work though, we're allowing users to create their own APIs. The image and text batch size should not be coupled IMO, as we need GenAI-Perf to be maximally extensible for new APIs. There may be ones where they want 2 texts and 3 images, for example, and the current approach does not support that. This is a minor code change but avoids this being a limiting factor in the future.

genai-perf/README.md
dyastremsky: I wasn't sure with this. `--batch-size-images` should come before `--batch-size-text`. However, if we look at the doc, the first arg for text is `-b` and the second is `--batch-size`, both of which would come before `--batch-size-images`. Happy to take either approach.

genai-perf/README.md
```diff
@@ -407,13 +407,21 @@ URL of the endpoint to target for benchmarking. (default: `None`)
 
 ##### `-b <int>`
 ##### `--batch-size <int>`
```
debermudez, line 409: should we remove this to avoid confusion?
or was this left in for backwards compatibility?

genai-perf/README.md
debermudez: Thats a good point that led to a different question.
This is fine, and you clearly considered it already before i asked. 
thanks for being so thorough.

nv-hwoo, line 418: nit but maybe `--batch-size-image` to be consistent with the text counterpart?

dyastremsky: Done!

templates/genai-perf-templates/README_template
```diff
@@ -407,13 +407,21 @@ URL of the endpoint to target for benchmarking. (default: `None`)
...
 ##### `--batch-size <int>`
+##### `--batch-size-text <int>`
 
 The batch size of the requests GenAI-Perf should send.
 This is currently only supported with the
 [embeddings](docs/embeddings.md), image_retrieval, and
 [rankings](docs/rankings.md) endpoint types.
 (default: `1`)
 
+##### `--batch-size-images <int>`
```
nv-hwoo, line 418: I missed this one :)

dyastremsky: Good catch! Fixed.

PR #112, opened 2024-10-03
Fix freezing between warmup and benchmarking bug

matthewkotila:
Moves unnecessary destruction from middle of execution to end.

### DIFF
--- a/src/client_backend/triton_c_api/triton_loader.cc
+++ b/src/client_backend/triton_c_api/triton_loader.cc
@@ -1363,8 +1363,11 @@ TritonLoader::GetSingleton()
 
 TritonLoader::~TritonLoader()
 {
-  TRITONSERVER_Error* allocator_err = response_allocator_delete_fn_(allocator_);
-  FAIL_IF_TRITONSERVER_ERROR(allocator_err, "deleting response allocator");
+  if (response_allocator_delete_fn_) {
+    TRITONSERVER_Error* allocator_err =
+        response_allocator_delete_fn_(allocator_);
+    FAIL_IF_TRITONSERVER_ERROR(allocator_err, "deleting response allocator");
+  }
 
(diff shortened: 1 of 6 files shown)


src/client_backend/triton_c_api/triton_loader.cc
```diff
@@ -1363,8 +1363,11 @@ TritonLoader::GetSingleton()
 
 TritonLoader::~TritonLoader()
 {
-  TRITONSERVER_Error* allocator_err = response_allocator_delete_fn_(allocator_);
```
matthewkotila, line 1366: Calling Delete() from main causes the static instance to be instantiated, which means at end of process, destructor gets called here, but if Triton C API mode wasn't used, none of these function pointers is initialized, so it caused a set fault. Fix was to guard against uninitialized function pointer.

nicolasnoble: Right, I see, yep. I hadn't spotted this one before. There's definitely some overall hardening we should be doing here, with better RAII-style usage, but this totally will do for now.

src/main.cc
```diff
@@ -24,6 +24,10 @@
 // (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+#ifdef TRITON_ENABLE_PERF_ANALYZER_C_API
```
nicolasnoble, line 27: Hah, I hadn't spotted the fact the API is conditional either. This is indeed an added complexity to the shutdown ordering problem.

nicolasnoble (dismissed):

nicolasnoble (approved):

PR #113, opened 2024-10-04
Use text, not text_input, as field name

dyastremsky:
Default to using text as field name, not text_input. This is simpler and consistent with the use of "image" as a field name (rather than image_input).


debermudez (approved):
approved assuming that the rankings NIM doesnt require text_input

dyastremsky:
> approved assuming that the rankings NIM doesnt require text_input

Thanks! Discussed offline. That endpoint uses `text` already and the changes maintain backward compatibility for all endpoints. Running CI now.

I rebased on previous PRs carefully and reviewed, but another set of eyes for a quick skim would be helpful if you have time.

debermudez (approved):

PR #114, opened 2024-10-04
Refactor OpenAI Chat converter into template-style design

nv-hwoo:
This PR
- simplifies and refactors `OpenAIChatCompletionsConverter` into template-style design
- create unit tests for the converter class that handles chat, vision, and image retrieval endpoints.

This PR does _NOT_
- change any code behavior (except that it assumes there's no dataset approach anymore, which is getting handled by #100 )


genai-perf/tests/test_input_image_retrieval.py
```diff
@@ -1,155 +0,0 @@
-# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
debermudez, line 1: why did this whole file get deleted?

nv-hwoo: From the PR description above,
> create unit tests for the converter class that handles chat, vision, and image retrieval endpoints.

Basically ported to/under `test_openai_chat_converter.py`

debermudez: oof. shouldn't have missed that.
thank you

genai-perf/tests/test_output_format_converter.py
```diff
@@ -23,6 +24,52 @@
 
 
 class TestOutputFormatConverter:
+
```
debermudez, line 27: nice test to ensure we don't break this.

debermudez (commented):
approval after the code ql fix

debermudez (approved):

PR #115, opened 2024-10-04
Pass Branch Names to Automatic CI

fpetrini15:
_Context:_ Currently, automatic CI for PRs is run using both the `main` branch of the PA repo and the client repo. This is an issue because it means the changes in the PR are not actually being tested by CI, risking a false positive from CI and merging a PR that causes breakages.

**Goal:** Pass the appropriate branch names to CI. The PA repo will use the branch name associated with the CI that triggered the pipeline and the client repo branch will default to `main` unless overwritten by the developer.

Note that it will be the developer's responsibility to restore any overwritten branch names prior to merging.

### DIFF
--- a/.github/workflows/trigger_ci.yml
+++ b/.github/workflows/trigger_ci.yml
@@ -17,4 +17,8 @@ jobs:
       - name: Trigger Pipeline
         run: |
           #!/bin/bash
-          curl --fail --request POST --form token=${{ secrets.PIPELINE_TOKEN }} -F ref=${GITHUB_HEAD_REF}  "${{ secrets.PIPELINE_URL }}"
+          curl --fail --request POST \
+            --form token=${{ secrets.PIPELINE_TOKEN }} \
+            --form variables[TRITON_CLIENT_REPO_TAG]=${GITHUB_HEAD_REF} \
+            --form variables[TRITON_CLIENT_REPO_TAG]=main \
+            -F ref=${GITHUB_HEAD_REF}  "${{ secrets.PIPELINE_URL }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger_ci.yml
```diff
@@ -17,4 +17,8 @@ jobs:
       - name: Trigger Pipeline
         run: |
           #!/bin/bash
-          curl --fail --request POST --form token=${{ secrets.PIPELINE_TOKEN }} -F ref=${GITHUB_HEAD_REF}  "${{ secrets.PIPELINE_URL }}"
+          curl --fail --request POST \
+            --form token=${{ secrets.PIPELINE_TOKEN }} \
+            --form variables[TRITON_PERF_ANALYZER_REPO_TAG]=${GITHUB_HEAD_REF} \
+            --form variables[TRITON_CLIENT_REPO_TAG]=fpetrini-dummy \
```
mc-nv, line 23: Isn't it a hardcoded value?

fpetrini15: `TRITON_CLIENT_REPO_TAG`? Yes, I envision this value being hardcoded to `main`. It would be the developer's responsibility to update this value when necessary and restore it prior to merging. This behavior is similar to the one we have adopted in gitlab.

mc-nv: I would suggest to set default value to the default `main` branch and communicate the same with developers.

fpetrini15: Yes, that's what I stated above. This is just a draft PR where I was trying to show that branch name changes will take effect. I can switch it back now and open it as a true PR if you find this solution to the slack thread acceptable.

debermudez (approved):

PR #117, opened 2024-10-07
Add Analyze to Search Parameters

nv-braf:
Since we are transitioning to supporting Analyze (versus Optimize) first I have added support for Analyze to the SearchParameters class (in addition to adding it to the temporary ConfigCommand class).

Example usage is:

Analyze:
   concurrency: [1, 4, 8, 16]
   num_prompts: 
       min: 10
       max: 100

As part of this I've differentiated between runtime options that are passed to PA and those that are used internally in GAP.

### DIFF
--- a/genai-perf/genai_perf/config/generate/search_parameters.py
+++ b/genai-perf/genai_perf/config/generate/search_parameters.py
@@ -46,9 +46,9 @@ class SearchParameters:
         "instance_count",
         "max_queue_delay",
     ]
-    runtime_parameters = ["runtime_batch_size", "concurrency", "request_rate"]
+    runtime_pa_parameters = ["runtime_batch_size", "concurrency", "request_rate"]
 
-    all_parameters = model_parameters + runtime_parameters
+    all_parameters = model_parameters + runtime_pa_parameters
 
     def __init__(
         self,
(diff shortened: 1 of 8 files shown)


genai-perf/genai_perf/config/generate/search_parameters.py
```diff
@@ -39,25 +39,34 @@ class SearchParameters:
...
         self,
-        config: ConfigCommand,
+        config: ConfigOptimize,
         is_bls_model: bool = False,
         is_ensemble_model: bool = False,
         is_composing_model: bool = False,
     ):
-        self._config = config.optimize
+        self._config = config
+        self._subcommand = (
```
nicolasnoble, line 67: Minor nit: it feels like this could be an `Enum` instead of a string.

nicolasnoble (approved):
LGTM, with maybe that very minor nit I mentioned, which you can ignore if you don't feel like changing this.

genai-perf/genai_perf/config/generate/search_parameters.py
nv-braf: No, this is a good suggestion. I've gone ahead and implemented this.

PR #119, opened 2024-10-08
Create GenAI-Perf Config Class

nv-braf:
The idea here is that we'll take in the user config options (ConfigCommand right now) as well as any options that can be set by Analyze/Optimize and will vary from profile to profile and output a new config that will be used by all the other modules in GAP.


debermudez (commented):
Signing off on the direction.
Leaving approval to @nicolasnoble

nicolasnoble (approved):
Yep, looks fine to me. Sorry for the delay, got distracted with other tasks.

PR #122, opened 2024-10-09
Pass PA Version to Build

fpetrini15:
The PA binary we have been creating since the migration is not versioned. This is because, though we are setting the PA version at the top-level, we are not passing it into the build process.

_Note:_ When testing locally, I believe I was also able to root-cause the `(commit unknown)` issue that has been around since pre-migration. It stems from github's notion of safe repositories. If the PA repo isn't added as a safe directory, it will refuse to execute git commands. In my testing, adding PA as a safe directory, I was able to extract the hash. I would need more time than currently available to find an acceptable solution to this problem. In the meantime, I would like to get this change merged and cherry-picked for `24.10`. (Tracked in `TPRD-530`)


debermudez (approved):

dyastremsky (approved):
Great work, Francesco!

PR #123, opened 2024-10-10
Create RunConfig class

nv-braf:
Adds the `RunConfig` class which holds all the state information about a single profile in GAP:

- Run name
- PA config info
- GAP config info
- Measurements (metrics coming back from PA)
- (in the future this will also hold information about triton/endpoint configuration when Optimize is coded)

As part of this story I've also added in the missing checkpoint support for the PA/GAP config classes. 

For high level context: 

Each instance of profiling (calling PA to get metrics) has a corresponding RunConfig and a group of these will be stored in (the next to be coded) `Results` class. The Results are then stored in the checkpoint file which will be read when GAP is called (by any subcommand). 

Built-in to this class are all the methods necessary to get configuration and metrics information for each time the model was profiled. This can be used both as inputs to other subcommands (like Visualize) or to skip over known results when re-profiling a model.

### DIFF
--- a/genai-perf/genai_perf/config/generate/genai_perf_config.py
+++ b/genai-perf/genai_perf/config/generate/genai_perf_config.py
@@ -12,13 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from copy import deepcopy
 from dataclasses import dataclass
+from typing import Any, Dict
 
 from genai_perf.config.generate.search_parameter import SearchUsage
 from genai_perf.config.input.config_command import (
     ConfigCommand,
     ConfigInput,
     ConfigOutputTokens,
+    ConfigSyntheticTokens,
 )
 from genai_perf.types import ModelObjectiveParameters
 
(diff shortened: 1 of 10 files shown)


genai-perf/genai_perf/config/generate/genai_perf_config.py
```diff
@@ -56,3 +59,37 @@ def _set_options_based_on_objective(
...
+    @classmethod
+    def read_from_checkpoint(
+        cls, genai_perf_config_dict: Dict[str, Any]
+    ) -> "GenAIPerfConfig":
+        """
+        Takes the checkpoint's representation of the class and creates (and populates)
+        a new instance of a GenAIPerfConfig
+        """
+        genai_perf_config = GenAIPerfConfig(
+            config=ConfigCommand([""]),
```
debermudez, line 84: Does it make sense to store the command that generated a run in the checkpoint?

debermudez (commented):
Overall looks really solid. 1 question inline and another based on your description.

You state in the desciption: "The Results are then stored in the checkpoint file which will be read when GAP is called (by any subcommand)."
I can understand the checkpoint file being read for most command but I understand why we would follow this process for the profile subcommand. Can you expand on that slightly?

genai-perf/genai_perf/config/generate/genai_perf_config.py
nv-braf: This is how (in future runs/subcommands) we will identify if the results have already been obtained.

debermudez: I misunderstood this piece of code.
Discussed offline and this is actually really well designed.

debermudez (approved):

nicolasnoble (approved):

PR #124, opened 2024-10-10
Create Generic Dataset

dyastremsky:
This PR pushes the initial work done by @debermudez and me to create a generic dataset class. As a proof of concept, we have added it to work here for the file input case for the chat endpoint.

The target feature branch will contain all of the code changes to support generic dataset class as the new approach. This branch sets up the initial class with this one working pathway, so that the remaining work can be parallelized.

The testing will be broken until all of the pathways are fixed and any relevant tests are updated.

Input command:
```bash
genai-perf profile   -m HuggingFaceH4/zephyr-7b-beta   --service-kind openai   --endpoint-type chat   --synthetic-input-tokens-mean 200   --synthetic-input-tokens-stddev 0   --output-tokens-mean 100   --output-tokens-stddev 0   --streaming   --tokenizer HuggingFaceH4/zephyr-7b-beta --input-file rankings_jsonl/queries.jsonl
```

image

### DIFF
--- a/genai-perf/genai_perf/inputs/converters/base_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/base_converter.py
@@ -25,18 +25,19 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import random
-from typing import Dict, List, Union, cast
+from typing import Dict, List, Union
 
 from genai_perf.exceptions import GenAIPerfException
 from genai_perf.inputs.input_constants import ModelSelectionStrategy
 from genai_perf.inputs.inputs_config import InputsConfig
+from genai_perf.inputs.retrievers.generic_dataset import GenericDataset
 
 
 class BaseConverter:
 
     _CONTENT_NAMES: List[str]
 
-    def convert(self, generic_dataset: Dict, config: InputsConfig) -> Dict:
+    def convert(self, generic_dataset: GenericDataset, config: InputsConfig) -> Dict:
         """
         Construct a request body using the endpoint specific request format.
         """
@@ -53,31 +54,27 @@ class BaseConverter:
             )
 
     def _construct_text_payload_batch_agnostic(
-        self, batch_size_text: int, input_data: Union[Dict, List]
+        self, batch_size_text: int, input_data: List
     ) -> Union[str, List]:
         """
         Construct text payload content for non-chat based LLM converters.
         Allow batched and unbatched input data.
         """
         if batch_size_text == 1:
-            input_data = cast(Dict, input_data)
             return self._construct_text_payload(input_data)
         else:
-            input_data = cast(List, input_data)
             return self._construct_batched_text_payload(input_data)
 
-    def _construct_text_payload(self, input_data: Dict) -> str:
+    def _construct_text_payload(self, input_data: List[str]) -> str:
         """
         Construct text payload content for non-chat based LLM converters.
         Since there are no roles or turns in non-chat LLM endpoints, all the
         (pre-defined) text contents are concatenated into a single text prompt.
         """
-        contents = [v for k, v in input_data.items() if k in self._CONTENT_NAMES]
-        return " ".join(contents)
+        return " ".join(input_data)
 
-    def _construct_batched_text_payload(self, input_data: List) -> List:
+    def _construct_batched_text_payload(self, input_data: List[str]) -> List:
         """
         Construct batched text payload content for non-chat based LLM converters.
         """
-        contents = [item["text"] for item in input_data]
-        return contents
+        return input_data
--- a/genai-perf/genai_perf/inputs/retrievers/generic_dataset.py
+++ b/genai-perf/genai_perf/inputs/retrievers/generic_dataset.py
@@ -35,9 +35,8 @@ GenericDatasetDict: TypeAlias = Dict[Filename, List[DataRowDict]]
 
 @dataclass
 class DataRow:
-    def __init__(self, texts: List[str], images: List[str]):
-        self.texts = texts
-        self.images = images
+    texts: List[str]
+    images: List[str]
 
     def to_dict(self) -> DataRowDict:
         """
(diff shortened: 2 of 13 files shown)


genai-perf/genai_perf/inputs/retrievers/generic_dataset.py
```diff
@@ -0,0 +1,93 @@
...
+from typing import Dict, List, TypeAlias
+
+Filename: TypeAlias = str
+TypeOfData: TypeAlias = str
+ListOfData: TypeAlias = List[str]
+DataRowDict: TypeAlias = Dict[TypeOfData, ListOfData]
+GenericDatasetDict: TypeAlias = Dict[Filename, List[DataRowDict]]
+
+
+class DataRow:
```
nv-braf, line 36: I would consider making simple classes like this `@dataclass`: https://docs.python.org/3/library/dataclasses.html

This will auto-create a ton of common methods for you and in general makes the code cleaner to use.

Also, I'm not sure if some of these classes are intended to be immutable (since I don't see any accessor methods).

genai-perf/genai_perf/inputs/retrievers/input_retriever_factory.py
```diff
@@ -68,57 +71,36 @@ def get_input_data(self) -> Dict:
...
             else:
                 raise GenAIPerfException("Input source is not recognized.")
 
-        return generic_dataset_json
+        return input_data
 
     def _convert_input_synthetic_or_file_dataset_to_generic_json(
         self, dataset: Dict
     ) -> Dict[str, List[Dict]]:
+        raise GenAIPerfException("STOP")
```
nv-braf, line 92: What is the purpose of this exception? It isn't descriptive and has no qualification.

genai-perf/genai_perf/inputs/retrievers/input_retriever_factory.py
```diff
@@ -68,57 +71,36 @@ def get_input_data(self) -> Dict:
...
-                        payload = self._encode_image(filename)
-                        content["image"] = payload
-            else:
-                filename = row["row"].get("image")
-                if filename:
-                    payload = self._encode_image(filename)
-                    row["row"]["image"] = payload
-
-        return input_file_dataset
+    def _encode_images_in_input_dataset(self, input_file_dataset: GenericDataset):
```
nv-braf, line 97: This seems like a method that should exist inside the GenericDataset class

genai-perf/genai_perf/inputs/converters/base_converter.py
```diff
@@ -25,18 +25,19 @@
...
 from genai_perf.inputs.inputs_config import InputsConfig
+from genai_perf.inputs.retrievers.generic_dataset import GenericDataset
 
 
 class BaseConverter:
 
     _CONTENT_NAMES: List[str]
 
-    def convert(self, generic_dataset: Dict, config: InputsConfig) -> Dict:
+    def convert(self, generic_dataset: GenericDataset, config: InputsConfig) -> Dict:
```
nv-braf, line 40: Just a general boy scout comment. It's good practice to not have naked Dict/List as types. If possible try to clean this up when you touch lines like this.

genai-perf/genai_perf/inputs/retrievers/generic_dataset.py
dyastremsky: Great suggestion. Done for all of these. They should be mutable, we just weren't sure whether there was a need for formal accessor methods.

genai-perf/genai_perf/inputs/retrievers/input_retriever_factory.py
dyastremsky: Whoops, this should not be in here. This was used during the initial debugging steps. I removed it. Thanks for catching this!

genai-perf/genai_perf/inputs/retrievers/input_retriever_factory.py
dyastremsky: Would we be able to do this in a future PR?

This is a great suggestion. I was actually looking to move it into the file retriever since it's specific to file data, though the generic dataset might also be a good home. However, it was not scoped in this PR, since this is essentially checkpointing the new data structure (GenericDataset) with one proof of concept for us to work off of in parallel. This PR is going to be the base for Elias and me to update everything to use the new structure, then clean up the input generation. We'll definitely move this function out of this file as things get cleaned up in the in the feature-generic-dataset PR.

genai-perf/genai_perf/inputs/converters/base_converter.py
dyastremsky: The output of this file is the whole dictionary with several nested data structures that we turn into a JSON. The dictionary schema is going to be convert-specific.

GenericDataset is supposed to make the earlier steps easier, but the data structure itself here is a very complicated schema. After this work does a lot of cleanup, it's possible that we could talk about if there's a better way to type hint for this. I was just updating the one relevant type hint for all the converters, but it's possible that made it look like I was doing full clean-up throughout. That wasn't the plan. That said, if you have ideas for how this could be type hinted, I'd be happy to open a ticket to do so. Ideally in a future PR, so that this PR doesn't blow up with all of the work for this feature and related cleanup.

dyastremsky:
After talking to @debermudez, I'm going to merge this. We wanted early feedback on the direction, so Brian's review was really helpful. I moved us over to dataclasses for the generic dataset, which looks much cleaner. We'll integrate the other feedback in future PRs.

We should hopefully have a feature branch PR once the retrievers and converters are updated to use the new dataclasses to simplify the code.

genai-perf/genai_perf/inputs/retrievers/input_retriever_factory.py
nv-braf: Yes, that's reasonable.

genai-perf/genai_perf/inputs/converters/base_converter.py
nv-braf: I would just create a type `Dict[any, any]` and make that the output of the function. It least it shows that you've thought about it and this is intended.

dyastremsky: Thanks for the suggestion! We'll add that.

CC: @debermudez for awareness

debermudez: this is exactly what we were thinking. thanks for the confirmation and call out @nv-braf

PR #125, opened 2024-10-10
hook up synthetic retriever to the new factory

debermudez:
This PR hooks up the synthetic retriever to the factory using the new data structure to hold data.

Moving rapidly, this is still going to fail unit testing and mypy until we are closer to the end of moving all of the files.

### DIFF
--- a/genai-perf/genai_perf/inputs/retrievers/synthetic_data_retriever.py
+++ b/genai-perf/genai_perf/inputs/retrievers/synthetic_data_retriever.py
@@ -25,9 +25,14 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
-from typing import Any, Dict, List
+from typing import List
 
 from genai_perf.inputs.input_constants import OutputFormat
+from genai_perf.inputs.retrievers.generic_dataset import (
+    DataRow,
+    FileData,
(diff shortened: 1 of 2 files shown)


genai-perf/genai_perf/inputs/retrievers/synthetic_data_retriever.py
```diff
@@ -62,7 +68,10 @@ def retrieve_data(self) -> List[Dict[str, Any]]:
                     image_height_stddev=self.config.image_height_stddev,
                     image_format=self.config.image_format,
                 )
-                data["image"] = image
+                row.images.append(image)
+            data_rows.append(row)
+        file_name = "synthetic_dataset"
```
dyastremsky, line 73: Do we want to make this a constant somewhere to avoid magic values?

dyastremsky (approved):
Very clean, great work! One comment. It shouldn't be a blocker but it'd be good to make sure we clean it up before the merge to main.

genai-perf/genai_perf/inputs/retrievers/synthetic_data_retriever.py
debermudez: I wasn't sure if we needed to unless we wanted to add multi file support to this path as well.
If we do, I will defer that til later on, after we do some more cleaning up.

PR #126, opened 2024-10-10
Add multi-file support to file retriever

dyastremsky:
Add multi-file support to file retriever. Now, an input directory can be passed and all JSONL files will be read in it and passed into GenericDataset.

In addition, this works gets rid of the rankings-specific retrieval code and updates the rankings converter to use GenericDataset.

image


dyastremsky:
Created this incorrectly.

PR #128, opened 2024-10-10
Add multi-file support in retriever

dyastremsky:
Add multi-file support to file retriever. Now, an input directory can be passed and all JSONL files will be read in it and passed into GenericDataset.

In addition, this works gets rid of the rankings-specific retrieval code and updates the rankings converter to use GenericDataset.

image

### DIFF
--- a/genai-perf/genai_perf/inputs/converters/rankings_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/rankings_converter.py
@@ -34,25 +34,48 @@ from genai_perf.inputs.retrievers.generic_dataset import GenericDataset
 class RankingsConverter(BaseConverter):
 
     def convert(self, generic_dataset: GenericDataset, config: InputsConfig) -> Dict[Any, Any]:
+        provided_filenames = list(generic_dataset.files_data.keys())
+        if "queries" not in provided_filenames or "passages" not in provided_filenames:
+            raise ValueError(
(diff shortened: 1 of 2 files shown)


genai-perf/genai_perf/inputs/converters/rankings_converter.py
```diff
@@ -34,26 +34,49 @@
...
+        passages_data = generic_dataset.files_data["passages"]
+
+        rows_of_passage_data = len(passages_data.rows)
+
         request_body: Dict[str, Any] = {"data": []}
 
-        for index, entry in enumerate(generic_dataset["rows"]):
-            model_name = self._select_model_name(config, index)
+        for query_index, query_row in enumerate(queries_data.rows):
+            if query_index > rows_of_passage_data:
```
debermudez, line 51: This sets the limit to the number of payloads to be based on the number of passages, in the event that more queries are provided than passages?

dyastremsky: It actually sets the limit to the min of (number of passage "payloads", number of query "payloads"). I use the term payloads because there are usually multiple passages per query. So it's based on rows, not the number of texts.

This avoids the need for a strict check that passage_rows == query_rows and instead uses the maximum usable subset in the case where the number differs, which should be uncommon.

debermudez (approved):

PR #133, opened 2024-10-13
Add client-side batching support for synthetic inputs

dyastremsky:
This pull request adds client-side batching support for synthetic inputs. As a proof of concept, it hooks up the embeddings converter, which now works for both file inputs and synthetic inputs with batching.

Batching with synthetic input data:
image

Batching with file input data:
image

### DIFF
--- a/genai-perf/genai_perf/inputs/retrievers/synthetic_data_retriever.py
+++ b/genai-perf/genai_perf/inputs/retrievers/synthetic_data_retriever.py
@@ -27,7 +27,7 @@
 
 from typing import List
 
-from genai_perf.inputs.input_constants import OutputFormat
+from genai_perf.inputs.inputs_config import InputsConfig
 from genai_perf.inputs.retrievers.generic_dataset import (
     DataRow,
     FileData,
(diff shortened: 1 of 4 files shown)


genai-perf/genai_perf/inputs/retrievers/synthetic_data_retriever.py
```diff
@@ -58,16 +58,17 @@ def retrieve_data(self) -> GenericDataset:
                 self.config.prompt_tokens_mean,
                 self.config.prompt_tokens_stddev,
             )
-            row.texts.append(prompt)
+            for _ in range(self.config.batch_size_text):
+                row.texts.append(prompt)
 
-            if self.config.output_format == OutputFormat.OPENAI_VISION:
```
dyastremsky, line 63: I removed this format-specific code. We don't want format-specific code in the retrievers. This would prevent new image converters from generating image data via the synthetic route.

There will be latency introduced by generating both texts and images. I created

dyastremsky:
Merging for efficiency (to avoid many PRs), since Elias is out and we'll get the final review on the feature branch.

PR #142, opened 2024-10-17
Update unit tests after GenericDataset redesign

dyastremsky:
After the GenericDataset redesign, tests need to be updated. Responsibility for tasks became specialized between different classes and some functions changed. This pull request updates the test, removes unnecessary tests, and adds new tests to ensure our unit tests are comprehensive.


genai-perf/genai_perf/inputs/converters/openai_embeddings_converter.py
matthewkotila, line 1: Not to be naggy, but is there any way this PR could be made smaller?

nicolasnoble: I thought he was saying he's going to break it down into sub PRs?

dyastremsky: Theoretically, yes. There are 15 unit tests updated, so there could be PRs with groupings of unit tests. And I would love to have smaller PRs. However, this was always on a very tight timeline, and now with it needing to get delivered by just one person rather than two by tomorrow, I didn't know how to break it up in a way that wouldn't lose slow things down. I didn't expect that most of the inputs testing would need to be redone.

If there was a second person working on this, we could have each done some of the testing and reviewed each other's code to keep up velocity, but the situation changed. If y'all want to each pick specific test files to review, that could work too. Open to alternatives. But we need to deliver this tomorrow.

CC: @ganeshku1 in case we can find ways to get time allocated for review of this work.

dyastremsky: @nicolasnoble That was actually a third PR. One was implementation (approved by Brian), this is unit testing, and the third will run the pre-commit hooks. The third will be branched off this one so that I can actually run CI on it and make sure all passes.

genai-perf/genai_perf/inputs/converters/rankings_converter.py
```diff
@@ -52,7 +52,7 @@ def convert(self, generic_dataset: GenericDataset, config: InputsConfig) -> Dict
         request_body: Dict[str, Any] = {"data": []}
 
         for query_index, query_row in enumerate(queries_data.rows):
-            if query_index > rows_of_passage_data:
+            if query_index >= rows_of_passage_data:
```
matthewkotila, line 55: Was this a bug fix?

genai-perf/tests/test_cli.py
```diff
@@ -568,70 +566,6 @@ def test_unrecognized_arg(self, monkeypatch, capsys):
                 ],
                 "The --output-tokens-mean-deterministic option is only supported with the Triton and TensorRT-LLM Engine service-kind",
             ),
-            (
```
matthewkotila, line 571: Why are these unnecessary now?

genai-perf/tests/test_file_input_retriever.py
```diff
@@ -37,193 +35,247 @@
...
-            read_data=file_contents.get(filepath, file_contents["queries.jsonl"])
-        )()
-
-    mock_open_obj = mock_open()
-    mock_open_obj.side_effect = open_side_effects
+        filename = Path(filepath).name
+        return mock_open(read_data=file_contents.get(filename))()
 
+    @staticmethod
+    @patch("genai_perf.inputs.retrievers.file_input_retriever.FileInputRetriever._encode_image", return_value="mock_base64_image")
```
matthewkotila, line 75: Not super familiar with this use case, but why is this method both a static method and also patching things?

genai-perf/tests/test_json_exporter.py
```diff
@@ -228,6 +228,7 @@ class TestJsonExporter:
           "output_tokens_mean_deterministic": false,
           "output_tokens_stddev": 0,
           "random_seed": 0,
+          "synthetic_input_files": null,
```
matthewkotila, line 231: `None`?

genai-perf/tests/test_openai_chat_converter.py
```diff
@@ -25,24 +25,54 @@
...
 
 
 class TestOpenAIChatCompletionsConverter:
 
+    @staticmethod
+    def create_generic_dataset(rows: List[Dict[str, Any]]) -> GenericDataset:
+        def clean_text(row):
+            text = row.get("text", [])
+            if isinstance(text, list):
+                return [t for t in text if t]
```
matthewkotila, line 42: What are these cleanings actually doing? Looks like they're getting rid of `None`/`False` objects in a list? Why would those be there?

genai-perf/tests/test_openai_chat_converter.py
```diff
@@ -104,7 +130,7 @@ def test_convert_with_request_parameters(self):
             model_name=["test_model"],
             model_selection_strategy=ModelSelectionStrategy.ROUND_ROBIN,
             output_format=OutputFormat.OPENAI_CHAT_COMPLETIONS,
-            add_stream=True,  # set streaming
```
matthewkotila, line 107: Aw, but that "set streaming" comment really helped me understand what `add_stream=True` was doing 😩

genai-perf/tests/test_openai_completions_converter.py
```diff
@@ -91,7 +96,8 @@ def test_convert_with_request_parameters(self):
             model_name=["test_model"],
             model_selection_strategy=ModelSelectionStrategy.ROUND_ROBIN,
             output_format=OutputFormat.OPENAI_COMPLETIONS,
-            add_stream=True,  # set streaming
+            add_stream=True,
+            output_tokens_mean=1234,
```
matthewkotila, line 100: Why is this added?

genai-perf/tests/test_rankings_converter.py
```diff
@@ -24,29 +24,45 @@
...
 from genai_perf.inputs.input_constants import ModelSelectionStrategy, OutputFormat
 from genai_perf.inputs.inputs_config import InputsConfig
+from genai_perf.inputs.retrievers.generic_dataset import GenericDataset, DataRow, FileData
+from typing import List, Optional
 
 
 class TestRankingsConverter:
 
+    @staticmethod
+    def create_generic_dataset(
```
matthewkotila, line 38: It looks like each of these classes implements its own `create_generic_dataset`, is the idea that each converter makes generic datasets a little differently for their own testing?

genai-perf/tests/test_rankings_converter.py
```diff
@@ -60,17 +76,17 @@ def test_convert_default(self):
                 {
                     "payload": [
                         {
-                            "query": {"text": "query 1"},
-                            "passages": [{"text": "passage 1"}, {"text": "passage 2"}],
+                            "query": "query 1",
+                            "passages": [{"text_input": "passage 1"}, {"text_input": "passage 2"}],
```
matthewkotila, line 80: Why the change from `text` to `text_input`? Is this a bug fix?

genai-perf/tests/test_synthetic_data_retriever.py
```diff
@@ -58,26 +95,29 @@ def test_synthetic_text(self, mock_prompt, output_format):
...
-            num_prompts=3,
+            num_prompts=2,
+            batch_size_text=1,
+            batch_size_image=1,
+            synthetic_input_filenames=["file1.jsonl", "file2.jsonl"],
             output_format=OutputFormat.OPENAI_VISION,
         )
         synthetic_retriever = SyntheticDataRetriever(config)
-        dataset = synthetic_retriever.retrieve_data()
+        dataset: GenericDataset = synthetic_retriever.retrieve_data()
```
matthewkotila, line 110: Just curious, why use type annotations? (at least here)

genai-perf/genai_perf/inputs/converters/rankings_converter.py
dyastremsky: Yes. I forgot Python is 0-indexes. The index needs to be greater or equal to the number of rows to break. As an example, if there are two queries, the second query row would have an index of 1... if there was one passage, the rows_of_passage_data would be 1... so it would not break (1 == 1) and would try to find a second passage then be out of bounds.

This was caught by the [new rankings test](https://github.com/triton-inference-server/perf_analyzer/pull/142/files#diff-047e86ccbac2c3fc8f4e3d037bc584ed6d49f0ea3bde4c21123ffb1d35213684R276) that checks mismatched numbers of queries and passage.

genai-perf/tests/test_cli.py
dyastremsky: These checks were moved to the converters, since they're format-specific. To allow GAP to be more extensible (and logically organized), format-specific checks have been moved to a check_config function implemented in the output format converters. See [this one](https://github.com/triton-inference-server/perf_analyzer/pull/142/files#diff-d25d2d07884eb6a98fb2c89b98a26c4669d89586646ceb103fe268a91deaac67R36) as an example.

genai-perf/tests/test_json_exporter.py
dyastremsky: I tried that and it returned an error in the testing (passed when changed to null). It looks like the other parameters that would have None also say null, so I figured it was consistent.

genai-perf/tests/test_openai_chat_converter.py
dyastremsky: Great catch, Matt. That's right. To be honest, I don't exactly remember. I know this solved a few issues to get the test to pass. It's possible that part of it was just passing single values vs lists to this function.

I can fix it for this PR or I can make a follow-up ticket to clean this up after merging. Do you have a preference?

genai-perf/tests/test_openai_completions_converter.py
dyastremsky: It fails otherwise. I don't know why this was working before, since the output_tokens_mean was never set, so it should have never been set in the payload.

genai-perf/tests/test_rankings_converter.py
dyastremsky: The generic dataset should theoretically be the same, but they all are testing slightly different things (e.g. rankings wants a multi-file setup with query and passage files, some tests want single/multiple images, some tests want single/multiple texts, etc.). Your comments is leading me to think it's likely there's a refactor opportunity to create a mock class for GenericDataset that gets reused.

For now, I just have this function as a fixture that sets up the generic datasets in each test so that we can test just the converters in isolation without making any calls to the retrievers or inputs.

genai-perf/tests/test_rankings_converter.py
dyastremsky: I can undo this. We initially supported text for rankings/embeddings and text_input for everything else. We then decided to consolidate around text_input. Finally, we consolidated around text for all inputs, with backwards compatible support for text_input until we go GA.

I must have gotten mixed up while writing this test and changed which field we're using. It'll work, but there's no reason to use the deprecated field name here.

genai-perf/tests/test_synthetic_data_retriever.py
dyastremsky: Very good catch. I was trying to make the type checker happy, but this actually got removed in [the pre-commit PR](https://github.com/triton-inference-server/perf_analyzer/pull/143), so it won't exist once everything is merged. Sorry for the confusion.

matthewkotila:
Looks good overall, just some clarification questions.

genai-perf/tests/test_file_input_retriever.py
dyastremsky: It's static since it doesn't rely on an instance of the class. It's just a utility function for the tests. The patching is to mock the dependencies (returns a mock value instead of actually calling encode, to avoid making the tests a lot less readable with actual encoded values).

genai-perf/tests/test_json_exporter.py
matthewkotila: I just don't understand how the Python interpreter reads this without an error:

```bash
$ python3.10
Python 3.10.12 (main, Sep 11 2024, 15:47:36) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> null
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'null' is not defined
```

genai-perf/tests/test_openai_chat_converter.py
matthewkotila: Thanks for explanation--follow up is fine 👍

genai-perf/tests/test_rankings_converter.py
matthewkotila: Sounds like going forward we prefer/support `text` but this test is for backwards compatibility with `text_input`?

dyastremsky: That's correct.

genai-perf/tests/test_json_exporter.py
dyastremsky: Maybe because it's a dictionary value? It looks like we use it throughout the this test.

matthewkotila: Oh, this is a triple-quoted string, that's a JSON identifier, which is valid.

matthewkotila (approved):

genai-perf/README.md
```diff
@@ -89,7 +89,7 @@ genai-perf --help
 Since GenAI-Perf depends on Perf Analyzer,
 you'll need to install the Perf Analyzer binary:
 
-### Install Perf Analyzer (Ubuntu, Python 3.8+)
+### Install Perf Analyzer (Ubuntu, Python 3.10+)
```
matthewkotila, line 92: @dyastremsky I think this change was missing its README_template associated change.

I fixed it in https://github.com/triton-inference-server/perf_analyzer/pull/150

dyastremsky: Awesome, thank you! It's been a bit of extra effort to keep the docs + doc templates in sync.

PR #143, opened 2024-10-18
Run pre-commit hooks after inputs redesign to use GenericDataset

dyastremsky:
Re-run pre-commit hooks after inputs redesign. Since the new analyze changes will only support Python 3.10+, updating the code here to also only support 3.10+ to use aliases.

### DIFF
--- a/.github/workflows/python-package-genai.yml
+++ b/.github/workflows/python-package-genai.yml
@@ -39,7 +39,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-22.04"]
-        python-version: ["3.8", "3.10"]
+        python-version: ["3.10"]
 
     steps:
     - uses: actions/checkout@v3
--- a/genai-perf/genai_perf/inputs/converters/rankings_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/rankings_converter.py
@@ -24,26 +24,31 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-from typing import Any, Dict
+from typing import Any, Dict, List, Union
 
+from genai_perf.exceptions import GenAIPerfException
 from genai_perf.inputs.converters.base_converter import BaseConverter
 from genai_perf.inputs.inputs_config import InputsConfig
 from genai_perf.inputs.retrievers.generic_dataset import GenericDataset
-from genai_perf.exceptions import GenAIPerfException
+
 
 class RankingsConverter(BaseConverter):
 
     def check_config(self, config: InputsConfig) -> None:
         if config.add_stream:
-            raise GenAIPerfException(f"The --streaming option is not supported for {config.output_format.to_lowercase()}.")
-    
-    def convert(self, generic_dataset: GenericDataset, config: InputsConfig) -> Dict[Any, Any]:
+            raise GenAIPerfException(
+                f"The --streaming option is not supported for {config.output_format.to_lowercase()}."
+            )
+
+    def convert(
+        self, generic_dataset: GenericDataset, config: InputsConfig
+    ) -> Dict[Any, Any]:
         provided_filenames = list(generic_dataset.files_data.keys())
         if "queries" not in provided_filenames or "passages" not in provided_filenames:
             raise ValueError(
                 "Both 'queries.jsonl' and 'passages.jsonl' must be present in the input datasets."
             )
-        
+
         queries_data = generic_dataset.files_data["queries"]
         passages_data = generic_dataset.files_data["passages"]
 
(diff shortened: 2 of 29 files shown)


.github/workflows/python-package-genai.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       fail-fast: false
```
matthewkotila, line 39: In the future, could we put all formatting-related changes in their own commit or even better in their own PR?

genai-perf/genai_perf/inputs/converters/rankings_converter.py
```diff
@@ -60,27 +65,25 @@ def convert(self, generic_dataset: GenericDataset, config: InputsConfig) -> Dict
 
             passage_entry = passages_data.rows[query_index]
 
+            passages: Union[List[str], List[Dict[str, str]]]
+            payload: Dict[str, Any]
```
matthewkotila, line 69: Just curious, why list the types here? Can't someone reading the code just jump to a definition to find that? That way python stays pythonic/simple?

genai-perf/tests/test_cli.py
```diff
@@ -772,9 +772,7 @@ def test_goodput_args_warning(self, monkeypatch, args, expected_error):
     def test_inferred_prompt_source(
         self, monkeypatch, mocker, args, expected_prompt_source
     ):
-        _ = mocker.patch("builtins.open", mocker.mock_open(read_data="data"))
-        _ = mocker.patch("os.path.isfile", return_value=True)
-        _ = mocker.patch("os.path.isdir", return_value=True)
+        _ = mocker.patch.object(Path, "is_file", return_value=True)
```
matthewkotila, line 775: 1. why is this an assignment
2. why did the other mocks in the before section get removed?

genai-perf/tests/test_file_input_retriever.py
```diff
@@ -143,32 +167,41 @@ def test_get_input_file_multiple_prompts(self, mock_file, mock_exists):
...
+        )
 
         assert file_data is not None
         assert file_data.filename == "multiple_prompts.jsonl"
         assert len(file_data.rows) == 3
         expected_prompts = [
             "What is the capital of France?",
             "Who wrote 1984?",
-            "What is quantum computing?"
+            "What is quantum computing?",
```
matthewkotila, line 180: Makin use of that trailing comma python 3.10 feature ;)

matthewkotila:
Looks good overall, just some small questions.

.github/workflows/python-package-genai.yml
dyastremsky: What do you mean? Usually, pre-commit is turned on, so the formatting would happen for every commit. Since it was turned off as part of this development work, re-enabling it immediately ran all the hooks. And then it warned me about types, other static errors, etc., so I fixed those manually, but it wouldn't let me commit without fixing everything.

I suppose I could have let the auto-formatting run, then uninstalled pre-commit, pushed a commit, re-installed pre-commit, and made any manual changes in a separate commit if that's in line with what you're asking for?

genai-perf/genai_perf/inputs/converters/rankings_converter.py
dyastremsky: Mypy complains when statically checking types, because the if/else branches have different types for the same variables. This was the fix. I could use different variable names for passages (though I think that's more confusing), but payload will raise an error either way since the dictionary values are different types.

genai-perf/tests/test_cli.py
dyastremsky: I'll remove the assignment. The other mocks were not necessary. I think what had happened was that we have a `[path.is_file() or path.is_dir](https://github.com/triton-inference-server/perf_analyzer/blob/8c0c09b79e613c809c5d6aded7ea14544e449de8/genai-perf/genai_perf/parser.py#L380)` check... and the way Python works, is_dir was an address and always evaluated to true. The checker caught that, so I had to make it `path.is_dir()` to properly be a function. 

Fixing the static error broke the test... and in fixing it, I realized that I had to mock Path directly (because of how we use it) and the other mocked functions no longer get called.

dyastremsky: Removed the assignment: https://github.com/triton-inference-server/perf_analyzer/pull/143/commits/9ffff7d5ca301cfe3a14f7cfb3b0a1219c39ef8c

.github/workflows/python-package-genai.yml
matthewkotila: I'm not sure how to achieve it, but I mean literally what I said. I had to read everything way closer because I couldn't tell if it was going to be a formatting related change or some other kind of change (e.g. using a python 3.10 feature or something).

It would be much easier to review if a single commit contained anything related to auto-formatting. Like run precommit without any feature-related changes for the first commit, then future commits are the non-formatting changes, like adding a trialing comma or whatever.

matthewkotila (approved):

.github/workflows/python-package-genai.yml
dyastremsky: Totally fair. I had originally had it as a separate commit and will do so next time.

Just a heads up that I didn't make the aforementioned change. The trialing comma was actually pre-commit formatting. The only things I did were fixing the errors pre-commit returned, then fixing the 2-3 CLI unit tests that broke when the behavior was fixed. I think you're right though - I can run it just with auto-format and save that as a commit so that you don't need to review that commit.

Thanks again for reviewing, Matt!

PR #147, opened 2024-10-22
Add support for nvclip endpoint

dyastremsky:
Add support for NVClip NIM endpoint.

Benchmarking NVCLip:
image

### DIFF
--- a/genai-perf/genai_perf/inputs/converters/base_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/base_converter.py
@@ -64,3 +64,9 @@ class BaseConverter:
             raise GenAIPerfException(
                 f"Model selection strategy '{config.model_selection_strategy}' is unsupported"
             )
+
+    def _add_request_params(
+        self, payload: Dict[Any, Any], config: InputsConfig
+    ) -> None:
+        for key, value in config.extra_inputs.items():
+            payload[key] = value
--- a/genai-perf/genai_perf/inputs/input_constants.py
+++ b/genai-perf/genai_perf/inputs/input_constants.py
@@ -39,14 +39,21 @@ class PromptSource(Enum):
 
 
 class OutputFormat(Enum):
+    ################################################################
+    # Triton backends
+    ################################################################
+    TENSORRTLLM = auto()
+    VLLM = auto()
+    ################################################################
+    # Other output formats
+    ################################################################
+    IMAGE_RETRIEVAL = auto()
+    NVCLIP = auto()
     OPENAI_CHAT_COMPLETIONS = auto()
     OPENAI_COMPLETIONS = auto()
     OPENAI_EMBEDDINGS = auto()
     OPENAI_VISION = auto()
     RANKINGS = auto()
-    IMAGE_RETRIEVAL = auto()
-    TENSORRTLLM = auto()
-    VLLM = auto()
     TENSORRTLLM_ENGINE = auto()
 
     def to_lowercase(self):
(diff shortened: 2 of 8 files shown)


genai-perf/genai_perf/inputs/converters/base_converter.py
```diff
@@ -64,3 +64,9 @@ def _select_model_name(self, config: InputsConfig, index: int) -> str:
             raise GenAIPerfException(
                 f"Model selection strategy '{config.model_selection_strategy}' is unsupported"
             )
+
+    def _add_request_params(
```
dyastremsky, line 68: This gets reused among converters, so I moved it to the base converter class. Other converts can reimplement it for custom logic.

genai-perf/genai_perf/inputs/input_constants.py
```diff
@@ -39,14 +39,21 @@ class PromptSource(Enum):
 
 
 class OutputFormat(Enum):
+    ################################################################
+    # Triton backends
+    ################################################################
+    TENSORRTLLM = auto()
```
dyastremsky, line 45: It looks like the logic added [here](https://github.com/triton-inference-server/perf_analyzer/pull/147/files#diff-531d67ceb8ef974e32746d3e8b7b2cd958bac761568f74220ef436b018cacf99R641) expected the options to only list the Triton backends. However, because that was unclear when adding OutputFormats here, it got out of sync.

We might need a better approach long-term as its own ticket, but I did clean-up here to fix the issue for now. 
image

genai-perf/genai_perf/parser.py
```diff
@@ -634,7 +638,7 @@ def _add_endpoint_args(parser):
     endpoint_group.add_argument(
         "--backend",
         type=str,
-        choices=utils.get_enum_names(ic.OutputFormat)[2:],
+        choices=utils.get_enum_names(ic.OutputFormat)[0:2],
```
dyastremsky, line 641: We might need a better approach long-term as its own ticket, but I did clean-up here to fix the issue for now. See help prompt with the correct backends displayed below.
image

nicolasnoble (approved):
This looks good to me, I haven't seen anything wrong with this PR.

One general comment however, which isn't related to this PR in particular, but I am starting to wish there was a bit more docstrings throughout the codebase, to better describe and document the APIs that we create, not only for the external users, but also for ourselves, when we have to maintain code in the future.

Don't take this as a request to document this one PR thoroughly, but this may be something to consider for future work / refactoring.

As an example, I feel I haven't documented enough my own python script that I submitted in third_party, tho I'd defend against my own argument saying this is a temporary fix until abseil's fix trickled down to us:

https://github.com/triton-inference-server/third_party/blob/main/tools/patch.py

dyastremsky:
> This looks good to me, I haven't seen anything wrong with this PR.
> 
> One general comment however, which isn't related to this PR in particular, but I am starting to wish there was a bit more docstrings throughout the codebase, to better describe and document the APIs that we create, not only for the external users, but also for ourselves, when we have to maintain code in the future.
> 
> Don't take this as a request to document this one PR thoroughly, but this may be something to consider for future work / refactoring.
> 
> As an example, I feel I haven't documented enough my own python script that I submitted in third_party, tho I'd defend against my own argument saying this is a temporary fix until abseil's fix trickled down to us:
> 
> https://github.com/triton-inference-server/third_party/blob/main/tools/patch.py

Thanks for reviewing and providing feedback! I'm happy to talk about how to best do this for GenAI-Perf and our codebase in general sometime. The [base converter class](https://github.com/triton-inference-server/perf_analyzer/blob/main/genai-perf/genai_perf/inputs/converters/base_converter.py) has some docstrings to document converter functions, though we can probably do more. Finding clearer ways to document is always good.

PR #150, opened 2024-10-23
Update docs to no longer require apt installing libb64

matthewkotila:


genai-perf/README.md
matthewkotila, line 1: @nicolasnoble @ganeshku1 we can merge before checks finish--this is just a docs update.

dyastremsky (approved):

PR #152, opened 2024-10-24
Convert cuda runtime dependency from dynamic linking to dynamic loading

matthewkotila:
Currently, if a user tried to run PA on a system without CUDA Runtime library installed, they see this:

```bash
$ perf_analzyer
perf_analyzer: error while loading shared libraries: libcudart.so.12: cannot open shared object file: No such file or directory
```

But PA really only needs the CUDA Runtime library when user uses `--shared-memory=cuda` option.

This PR removes the always-necessary requirement (dynamic linking) of the CUDA Runtime library (libcudart.so).

Before:

```bash
$ ldd $(which perf_analyzer)
...
        libcudart.so.12 => /usr/local/cuda/targets/x86_64-linux/lib/libcudart.so.12 (0x00007cea38600000)
...
```

After:

```bash
$ ldd $(which perf_analyzer)
...
        # not libcudart.so
...
```

Instead, PA will dynamically load libcudart.so via dlopen, only when users use `--shared-memory=cuda`.

Which means a single released PA binary can work on user systems _with_ and _without_ cuda installed.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -155,6 +155,7 @@ set(
   profile_data_exporter.cc
   periodic_concurrency_manager.cc
   periodic_concurrency_worker.cc
+  cuda_runtime_library_manager.cc
 )
 
 set(
@@ -198,6 +199,7 @@ set(
   periodic_concurrency_manager.h
   periodic_concurrency_worker.h
   thread_config.h
+  cuda_runtime_library_manager.h
 )
 
 add_executable(
@@ -263,10 +265,9 @@ if(TRITON_ENABLE_GPU)
     PUBLIC TRITON_ENABLE_GPU=1
   )
 
-  target_link_libraries(
-    perf_analyzer
-    PRIVATE CUDA::cudart
-  )
+  target_include_directories(perf_analyzer
+                             PUBLIC ${CUDAToolkit_INCLUDE_DIRS})
+
 endif() # TRITON_ENABLE_GPU
 
 if(TRITON_ENABLE_PERF_ANALYZER_C_API)
--- a/src/cuda_runtime_library_manager.cc
+++ b/src/cuda_runtime_library_manager.cc
@@ -30,7 +30,7 @@ namespace triton::perfanalyzer {
 
 CUDARuntimeLibraryManager::CUDARuntimeLibraryManager()
 {
-  handle_ = dlopen("libcudart.so", RTLD_LAZY);
+  handle_ = dlopen("libcudart.so", RTLD_NOW);
   if (handle_ == nullptr) {
     throw std::runtime_error(
         std::string("Cannot open CUDA runtime library: ") + dlerror() + "\n");
--- a/src/cuda_runtime_library_manager.cc
+++ b/src/cuda_runtime_library_manager.cc
@@ -1,0 +1,117 @@
+// Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions
+// are met:
+//  * Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
+//    documentation and/or other materials provided with the distribution.
+//  * Neither the name of NVIDIA CORPORATION nor the names of its
+//    contributors may be used to endorse or promote products derived
+//    from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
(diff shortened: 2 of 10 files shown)


src/cuda_runtime_library_manager.h
```diff
@@ -0,0 +1,68 @@
...
+  const char* (*cuda_get_error_string_func_)(cudaError_t){nullptr};
+  cudaError_t (*cuda_ipc_get_mem_handle_func_)(cudaIpcMemHandle_t*, void*){
+      nullptr};
+  cudaError_t (*cuda_memcpy_func_)(void*, const void*, size_t, cudaMemcpyKind){
+      nullptr};
+  cudaError_t (*cuda_set_device_func_)(int){nullptr};
+
+  void* handle_{nullptr};
+
+  void LoadFunctions();
```
nicolasnoble, line 65: This feels like this should be a private helper member, no?

src/cuda_runtime_library_manager.cc
```diff
@@ -0,0 +1,117 @@
...
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "cuda_runtime_library_manager.h"
+
+namespace triton::perfanalyzer {
+
+CUDARuntimeLibraryManager::CUDARuntimeLibraryManager()
+{
+  handle_ = dlopen("libcudart.so", RTLD_LAZY);
```
nicolasnoble, line 33: Hmmm... would you mind trying with `RTLD_NOW` instead? `RTLD_LAZY` can be dangerous with third party libraries we don't directly maintain, even though, well, cuda is technically Nvidia's :) Sorry for the extra work. If `RTLD_NOW` works, I'd rather use it instead. The difference is that if `libcudart.so` has other transitive dependencies, they'll be all resolved immediately, and the call will fail at that time if they're not found, instead of silently failing later.

nicolasnoble (approved):
Couple of very minor nits, but great job! Thanks for doing this.

src/cuda_runtime_library_manager.cc
matthewkotila: `LoadFunctions()` which happens directly after this, calls dlsym, and checks for null returns--isn't that good enough? Perhaps even better since it doesn't load unnecessary symbols?

src/cuda_runtime_library_manager.h
matthewkotila: It is in the private section

src/infer_data_manager_shm.h
```diff
@@ -159,6 +162,9 @@ class InferDataManagerShm : public InferDataManagerBase {
   size_t output_shm_size_;
   // Map from shared memory key to its starting address and size
   std::unordered_map<std::string, SharedMemoryData> shared_memory_regions_;
+
+ private:
+  CUDARuntimeLibraryManager cuda_runtime_library_manager_{};
```
matthewkotila, line 167: Actually, I might want to change this to a singleton, because I think this class gets instantiated more than once throughout the process lifetime.

matthewkotila: Nevermind, this class is only created once per process lifetime.

src/cuda_runtime_library_manager.cc
nicolasnoble: The issue is it would just fail silently *at usage time* and crash, if there was, in fact, a problem loading dependencies, which is kind of counter intuitive from the nicer error message you have already in place, as it'd be completely side stepped.

src/cuda_runtime_library_manager.h
nicolasnoble: Derp. I didn't spot there was a lone private statement cutting off a bunch of imports. Don't mind me.

src/cuda_runtime_library_manager.cc
matthewkotila: Wait you're saying dlsym can return a non-null pointer and it could crash (silently? what does a silent crash look like) when the pointer actually gets used?

dyastremsky:
Out of curiosity, have you timed the runtime of PA before versus after the change? I assume the time to dynamically load the dependencies is minor, but wanted to confirm since this would run every time PA is run with CUDA shared memory.

src/cuda_runtime_library_manager.cc
nicolasnoble: So, the full context:

`dlopen` with LAZY will always work if the file simply exists and is a valid file. Any of its own dependencies which aren't there will not be loaded, and it won't trigger an error, assuming they may be loaded and resolved later with another `dlopen`. Then, `dlsym` will only return pointers to functions, if they exist in the lookup table.

What this means is that any unresolved import inside of the library itself will just be a null pointer, as the LAZY mode only fills them in when they are loaded by other means, and the crash will be a typical uncaught segfault due to trying to access a null pointer when running code from within the library, as there is no mechanism at this point to tell if the library might be missing imports. The goal of the LAZY mechanism is to resolve circular dependencies, or solve situations where libraries are loaded on demand, like being streamed from a non-filesystem network location, but it results in a fragile environment.

On the countrary, `dlopen` with NOW means that any of the import inside of the loaded library MUST be resolved at open time, and it'll return an error if that's not the case, guaranteeing no dangling null pointer in the loaded library. It means that all of the transitives dependencies must be present already in memory, or be loadable from the filesystem.

src/CMakeLists.txt
matthewkotila, line 1: > @dyastremsky: Out of curiosity, have you timed the runtime of PA before versus after the change? I assume the time to dynamically load the dependencies is minor, but wanted to confirm since this would run every time PA is run with CUDA shared memory.

(Responding in file so a thread can be created)

I did not time the difference, but since dlopen is only called once, there will generally be no significant performance difference between that and dynamically linking the library.

Either the dynamic linker loads the libraries at the start of the executable (implicit linking/load-time dynamic linking)

Or the library is loaded via dlopen (explicit linking/run-time dynamic linking).

nicolasnoble: Both are at the start of the executable anyway, right? The dynamic case is going to be behind an option, which can't be turned on dynamically at runtime, so it wouldn't even cause a stutter during usage.

matthewkotila: Yeah, _relatively_ at the "start". And yes, the optional nature is what we really want with dlopen.

src/cuda_runtime_library_manager.cc
matthewkotila: Done

nicolasnoble (approved):

PR #153, opened 2024-10-24
Fix inference response access resulting in segfault

lkomali:
While using Triton CApi, the inference response is deleted even before it's copied which causes segmentation fault.
This PR fixes the issue by deep copying the output as soon as the CApi receives it.


lkomali:
@indrajit96 @statiraju Tagging you on the PR.

src/client_backend/triton_c_api/triton_loader.cc
```diff
@@ -1041,10 +1041,30 @@ TritonLoader::GetOutputs(
             &byte_size, &memory_type, &memory_type_id, &userp),
         "inference_response_output_fn_ error");
 
+    std::string data_type{datatype};
```
nicolasnoble, line 1044: This feels like this should be an std::string_view, but that's not really a big deal, all things considered, in this specific scenario.

src/infer_context.cc
```diff
@@ -214,19 +217,11 @@ InferContext::GetOutputs(const cb::InferResult& infer_result)
   RequestRecord::ResponseOutput output{};
   for (const auto& requested_output : infer_data_.outputs_) {
     std::string data_type{requested_output->Datatype()};
-    const uint8_t* buf{nullptr};
-    size_t byte_size{0};
-    infer_result.RawData(requested_output->Name(), &buf, &byte_size);
+    std::vector<uint8_t> buf{};
+    infer_result.RawData(requested_output->Name(), buf);
 
-    // The first 4 bytes of BYTES data is a 32-bit integer to indicate the size
```
nicolasnoble, line 221: So this change here feels a bit clumsy, overall. What is the motivation to change the behavior of this function, and push it down the workflow, aside from trying to reduce copies, maybe?

src/profile_data_exporter.cc
```diff
@@ -162,34 +162,40 @@ ProfileDataExporter::AddResponseTimestamps(
...
 ProfileDataExporter::SetValueToJSON(
-    rapidjson::Value& json, const size_t index, const uint8_t* buf,
-    const size_t byte_size, const std::string& data_type)
+    rapidjson::Value& json, const size_t index, const std::vector<uint8_t>& buf,
+    const std::string& data_type)
 {
+  auto data_type_size_opt = GetDataTypeSize(data_type);
+  if (!data_type_size_opt || (index * *data_type_size_opt >= buf.size())) {
+    std::cerr << "Index out of bounds." << std::endl;
+    return;  // Handle error appropriately
```
nicolasnoble, line 171: Did you mean TODO: in this comment?

nicolasnoble (commented):
So I have a couple of minor comments, and just one question, but otherwise this looks good to me.

We probably still should heavily refactor this, but...

src/profile_data_exporter.cc
lkomali: This comment I added for me to track when I was trying different things resulting in errors. I will remove it.

src/infer_context.cc
lkomali: Yeah, so the overhead was because there was a resize here and for larger outputs, it's a big operation.  To avoid the additional overhead, I moved this condition down in the workflow so that we don't need to resize the vector again.

lkomali: When I removed this resize operation, the additional overhead reduced.

nicolasnoble: Yep, okay, makes sense, thanks for the confirmation of my suspicion here.

This is still a sort of API change, but I'd leave this for a later refactor...

src/client_backend/client_backend.h
```diff
@@ -652,8 +652,7 @@ class InferResult {
   /// \return Error object indicating the success or failure of the
```
matthewkotila, line 652: I'm getting a build error on a clean rebuild:

```
In file included from /root/dev/perf_analyzer/src/client_backend/torchserve/torchserve_client_backend.cc:27:
/root/dev/perf_analyzer/src/client_backend/torchserve/torchserve_client_backend.h:104:9: error: 'triton::perfanalyzer::clientbackend::Error triton::perfanalyzer::clientbackend::torchserve::TorchServeInferResult::RawData(const string&, const uint8_t**, size_t*) const' marked 'override', but does not override
  104 |   Error RawData(
      |         ^~~~~~~
```

lkomali: Let me take a look, looks like some of the changes were not saved

lkomali: Yeah, fixed it, the build was successful in the latest pipeline.

nicolasnoble (approved):

PR #156, opened 2024-10-25
Remove RHEL-specific logic

fpetrini15:
**Goal:** Remove unnecessary RHEL flags. They are no longer needed as a result of this PR; https://github.com/triton-inference-server/perf_analyzer/pull/148


matthewkotila (approved):
🎉

PR #158, opened 2024-10-28
Update versions in PA for Release 24.10

pvijayakrish:
Updates the README file for the 24.10 Triton release.

### DIFF
--- a/templates/template_vars.yaml
+++ b/templates/template_vars.yaml
@@ -1,7 +1,7 @@
 General:
-  release: 24.09
+  release: 24.10
   triton_cli_version: 0.0.11
-  genai_perf_version: 0.0.7dev
+  genai_perf_version: 0.0.7
 
 README:
   filename: README.md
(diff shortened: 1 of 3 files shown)


templates/template_vars.yaml
```diff
@@ -1,7 +1,7 @@
 General:
-  release: 24.09
+  release: 24.10
   triton_cli_version: 0.0.11
-  genai_perf_version: 0.0.7dev
+  genai_perf_version: 0.0.7
```
mc-nv, line 4: Aren't version should be:
```suggestion
  genai_perf_version: 0.0.8dev
```

fpetrini15: This is being merged into a release branch, so I believe `0.0.7` is the correct version.

fpetrini15 (approved):

templates/template_vars.yaml
mc-nv: @ganeshku1 , could you please advise which version of "genai-perf" will be used for 24.11 release?

pvijayakrish: @mc-nv **0.0.7** is the correct version to be used for genai-perf. Please refer https://github.com/triton-inference-server/perf_analyzer/blob/r24.10/genai-perf/genai_perf/__init__.py

PR #164, opened 2024-11-04
Add ISL/OSL CLI arg aliases

dyastremsky:
Input sequence length (ISL) and output sequence length (OSL) are becoming very common paths for users, so add them as aliases to shorten the args --synthetic-input-tokens-mean and --output-tokens-mean.

We plan to simplify the CLI args when GenAI-Perf goes GA. We may also switch to config files. For now, this will aid usability for users who prefer using these aliases.


lkomali:
LGTM

lkomali (approved):

PR #166, opened 2024-11-06
Turn statistics into GAP Records

nv-braf:
I've gone through and changed all the Records imported from MA to match the naming in GAP and added/removed things as necessary. I now believe that everything in Records aligns with the names output by the statistics methods.

**This is the bulk of the changes and do not warrant more than a cursory review.**

The new code is in `statistics.py` - adding a method to create and list of Records based on the statistics gathered. This will be used to populate the RunConfigMeasurement each time PA is called to gather measurements for trial of sweep.


genai-perf/genai_perf/record/record.py
```diff
@@ -90,16 +91,7 @@ class Record(metaclass=RecordType):
     records
     """
 
-    def __init__(self, value: Union[float, int], timestamp: int):
-        """
```
nv-hwoo, line 94: Why are we getting rid of the entire docstring?

genai-perf/genai_perf/record/types/cpu_available_ram.py
nv-hwoo, line 1: Are these files no longer used in the MA?

genai-perf/genai_perf/record/types/gpu_power_usage.py
```diff
@@ -20,86 +20,33 @@
...
 class GPUPowerUsage(DecreasingGPURecord):
     """
-    GPU Power Usage
+    GPU's power usage metric
     """
 
     tag = "gpu_power_usage"
 
     def __init__(self, value, device_uuid=None, timestamp=0):
-        """
```
nv-hwoo, line 29: Same question as above: why not keep the docstring?

genai-perf/genai_perf/types.py
```diff
@@ -38,6 +38,7 @@
 GpuRecords: TypeAlias = Dict[GpuId, TelemetryRecords]
 PerfRecords: TypeAlias = Dict[str, "Record"]  # type: ignore
 PerfMetricName: TypeAlias = str
+RecordValue = Union[float, int]
```
nv-hwoo, line 41: Could we make this to be called something more generic (e.g. `Number`). Asking since I could see this type being useful outside of the `Record` class.

genai-perf/genai_perf/record/record.py
nv-braf: Because they're redundant when variables are typed and well named

genai-perf/genai_perf/record/types/cpu_available_ram.py
nv-braf: They are still used in MA - but GAP does not collect CPU telemetry at this time.

genai-perf/genai_perf/record/types/gpu_power_usage.py
nv-braf: Same answer :-).

genai-perf/genai_perf/types.py
nv-braf: There's no restriction on having multiple types alias to the same thing. In general, I feel that it defeats the purpose of typing to make them generic. If something else (that's not a record) also has this type they should create their own type. This both serves as documentation and can prevent bugs from being introduced.

nv-hwoo (approved):

PR #173, opened 2024-11-08
Remove unnecessary filename field in FileData

dyastremsky:
Remove the field in FileData that holds a reference to the file's name. It is not needed, since GenericDatastructure holds a dictionary of filenames to file data and is the way that file data are referenced. The field results in duplication, which could mask bugs.

### DIFF
--- a/genai-perf/genai_perf/inputs/retrievers/generic_dataset.py
+++ b/genai-perf/genai_perf/inputs/retrievers/generic_dataset.py
@@ -52,14 +52,12 @@ class FileData:
 
     def to_list(self) -> List[DataRowDict]:
         """
-        Converts the FileData object to a dictionary.
+        Converts the FileData object to a list.
         Output format example for two payloads from a file:
-        {
-            'file_0': [
-                {'texts': ['text1', 'text2'], 'images': ['image1', 'image2']},
-                {'texts': ['text3', 'text4'], 'images': ['image3', 'image4']}
-            ]
-        }
+        [
+            {'texts': ['text1', 'text2'], 'images': ['image1', 'image2']},
+            {'texts': ['text3', 'text4'], 'images': ['image3', 'image4']}
+        ]
         """
         return [row.to_dict() for row in self.rows]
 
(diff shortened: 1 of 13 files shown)


genai-perf/genai_perf/inputs/retrievers/generic_dataset.py
```diff
@@ -48,10 +48,9 @@ def to_dict(self) -> DataRowDict:
 
 @dataclass
 class FileData:
-    filename: str
     rows: List[DataRow]
 
-    def to_dict(self) -> Dict[Filename, List[DataRowDict]]:
+    def to_list(self) -> List[DataRowDict]:
         """
         Converts the FileData object to a dictionary.
```
nv-hwoo, line 55: Need to update the docstring.

genai-perf/genai_perf/inputs/retrievers/generic_dataset.py
```diff
@@ -81,6 +80,6 @@ def to_dict(self) -> GenericDatasetDict:
         }
```
nv-hwoo, line 80: is the example in the docstring still correct?

genai-perf/genai_perf/inputs/retrievers/generic_dataset.py
dyastremsky: Good catch, fixed!

genai-perf/genai_perf/inputs/retrievers/generic_dataset.py
dyastremsky: Thanks for checking. This docstring is still correct. The output format wasn't changed.

lkomali:
LGTM. Would wait for @nv-hwoo to take a look and approve/comment accordingly.

nv-hwoo (dismissed):
Just one nit suggestion to update the docstring. Otherwise, looks great 👍

dyastremsky:
> Just one nit suggestion to update the docstring. Otherwise, looks great 👍

Thanks, done!

nv-hwoo (approved):
Nice, LGTM!

PR #176, opened 2024-11-13
Handle sse comments and non-data events

nv-hwoo:
- Seems like some SSE services send empty responses (e.g. `:`) as a keep alive mechanism. Handle these responses by removing them in the preprocessor function.
- Some bug fix in the code and readability improvement


genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
```diff
@@ -182,26 +183,44 @@ def _preprocess_response(
...
-                responses = response.strip().split("\n\n")
+                response = res_outputs[i]["response"].strip()
+
+                # Ignore keepalive SSE response
+                if response == ":":
+                    res_outputs[i]["response"] = ""
+                    continue
+
+                responses = response.split("\n\n")
                 if len(responses) > 1:
```
GuanLuo, line 224: one liner to split and filter only the data events
```suggestion
                response = res_outputs[i]["response"]
                responses = [data for data in response.strip().split("\n\n") if data.startswith("data:")]
                if len(responses) == 0:
                    response = ""
                # always overwrite response as it may have removed non-data event
                else:
```

dyastremsky (approved):

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
```diff
@@ -182,26 +183,44 @@ def _preprocess_response(
...
             # This forces a merge with the previous chunk if error detected.
             if len(res_outputs) > 1:
-                for i in reversed(range(len(res_outputs))):
+                for i in reversed(range(1, len(res_outputs))):
                     response = res_outputs[i]["response"]
+
+                    # Ignore keepalive SSE response
+                    if response.strip() == ":":
+                        res_outputs[i]["response"] = ""
+                        continue
```
GuanLuo, line 192: Probably not intended side affect but I don't think you even need to change this part, you can keep the logic of stitching chunks until "data:" is found and it should still work in combination of the suggested multi-response change below..
i.e. if you have two chunk `"data: {}\n\n"`, `":\n\n"`, existing logic will merge them to `"data: {}\n\n:\n\n"`, and later multi-response processing will split it back to `["data: {}", ":"]` and only `"data: {}"` will remain after the filtering.

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
nv-hwoo: > i.e. if you have two chunk "data: {}\n\n", ":\n\n", existing logic will merge them to "data: {}\n\n:\n\n", and later multi-response processing will split it back to ["data: {}", ":"] and only "data: {}" will remain after the filtering.

That's actually a great insight that multi-response handling will automatically handle the keepalive responses. And with few modifications to the suggestion, this indeed works. But one issue with the suggestion is that it breaks when the the preprocessor method is run in non-streaming case because non-streaming case does not have `data:` prefix. Of course, we can workaround by checking if the current run is streaming or not... which should be quite simple, one would think! But surprisingly, the only way right now to check is by parsing the actual response (Why? TLDR; because of genai-perf `compare` subcommand functionality). So we would need to do something like
```python
res_str = res_outputs[i]["response"].strip()
res = load_json_str(remove_sse_prefix(res_str))
if res["choices"][0]["object"] == "chat.completion.chunk":
    ...
```
but having this check in the middle of the code block is probably far from ideal. I think at the core of the problem is that this class needs some major refactoring (which we all know/agree), but that would take much larger changes given the urgency of the fix.

rmccorm4: Similar to Guan's comments on filtering non-data events, it is possible to get error events over SSE as well - which could be handled more gracefully:

Not 100% sure on the schema, but here's an example output I got:
```
[DEBUG][before] res_outputs[i]['response']='data: {"id":"cmpl-18456e82-a156-11ef-a4bd-04d4c4933ecf","choices":[{"index":0,"finish_reason":null,"delta":{"content":" appears"}}],"created":1731457661,"model":"llama-3.1-8b-instruct","object":"chat.completion.chunk","usage":null,"system_fingerprint":null}\n\nevent: error\n: expected either a str or dict for enum'
[DEBUG][before] responses=['data: {"id":"cmpl-18456e82-a156-11ef-a4bd-04d4c4933ecf","choices":[{"index":0,"finish_reason":null,"delta":{"content":" appears"}}],"created":1731457661,"model":"llama-3.1-8b-instruct","object":"chat.completion.chunk","usage":null,"system_fingerprint":null}', 'event: error\n: expected either a str or dict for enum']
[DEBUG][before] len(responses)=2
2024-11-13 00:27 [ERROR] genai_perf.utils:82 - Failed to parse JSON string: 'event: error
: expected either a str or dict for enum'
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/genai_perf/main.py", line 214, in main
    run()
  File "/usr/local/lib/python3.10/dist-packages/genai_perf/main.py", line 206, in run
    data_parser = calculate_metrics(args, tokenizer)
  File "/usr/local/lib/python3.10/dist-packages/genai_perf/main.py", line 142, in calculate_metrics
    return LLMProfileDataParser(
  File "/usr/local/lib/python3.10/dist-packages/genai_perf/profile_data_parser/llm_profile_data_parser.py", line 75, in __init__
    super().__init__(filename, goodput_constraints)
  File "/usr/local/lib/python3.10/dist-packages/genai_perf/profile_data_parser/profile_data_parser.py", line 62, in __init__
    self._parse_profile_data(data)
  File "/usr/local/lib/python3.10/dist-packages/genai_perf/profile_data_parser/profile_data_parser.py", line 125, in _parse_profile_data
    metrics = self._parse_requests(requests)
  File "/usr/local/lib/python3.10/dist-packages/genai_perf/profile_data_parser/llm_profile_data_parser.py", line 94, in _parse_requests
    self._preprocess_response(res_timestamps, res_outputs)
  File "/usr/local/lib/python3.10/dist-packages/genai_perf/profile_data_parser/llm_profile_data_parser.py", line 225, in _preprocess_response
    text = self._extract_openai_text_output(r)
  File "/usr/local/lib/python3.10/dist-packages/genai_perf/profile_data_parser/llm_profile_data_parser.py", line 329, in _extract_openai_text_output
    data = load_json_str(response)
  File "/usr/local/lib/python3.10/dist-packages/genai_perf/utils.py", line 79, in load_json_str
    return json.loads(json_str)
```

GuanLuo: I don't know if it is a good idea to have a single code path to handle cleaning of two different response schemes, but I think it is doable, with one important assumption: "\n\n" is guarantee to be a differentiator of the schemes. Then your preprocessor actually consist of two parts:

1. stitching message split across responses, which naively move and concatenate the string until it sees a message end indicator ("\n\n"), and then continue for the next message. "\n\n" is convenient because it is indeed message end in streaming (SSE), and there is no message end for non-streaming so it simply stitch all responses together, which is desired.
2. (streaming only) further process the messages in responses. Now that we have make sure every message starts and ends in one response, we can examine the message easily:
    1. firstly we check if "\n\n" has even been seen in 1., if not, then we don't need to do anything
    2. continue with split and filter and group...

nv-hwoo: Sure, let me try applying your suggestions. I expect some overall changes to the method since `res_timestamps` needs to be updated accordingly, but probably doable as well.

nv-hwoo: @GuanLuo removed the manual checks and maintained the original logic to handle SSE comments.

@rmccorm4 Unfortunately, the current code does not handle other SSE fields like event :( Until now, this was closer to a basic WAR to handle openai streaming responses. And it probably means that we want more proper support for SSE responses.  This is likely a larger effort than the current PR, so I'll create a followup ticket to get started on this.

dyastremsky:
@nv-hwoo, are you still trying to get this in? I feel like you added this functionality in a different PR.

nv-hwoo:
@dyastremsky I don't remember the other PR but yes I believe this is still needed in the near future. @rmccorm4 mentioned that this fix will be needed with the recent change in the triton distributed.

dyastremsky:
> @dyastremsky I don't remember the other PR but yes I believe this is still needed in the near future. @rmccorm4 mentioned that this fix will be needed with the recent change in the triton distributed.

Got it, thanks for clarifying Hyunjae!

rmccorm4:
In the latest versions of everything we're using in Triton Distributed - we don't currently have any issues with genai-perf, but I will follow up as soon as we run into any issue. 

We're just starting to get everything hooked up together behind the new OpenAI frontend and starting to test and benchmark it, so will know more next week likely.

piotrm-nvidia:
This issue blocks high concurrency benchmarks in Triton Distributed:

https://github.com/triton-inference-server/triton_distributed/issues/264

rmccorm4 (approved):

PR #179, opened 2024-11-13
Fix statistics unit bug for request_goodput

nv-hwoo:
Fix a bug in `_add_unit` method in `Statistics` class. The a bug in the check for `request_goodput` unit was causing the other irrelevant metrics to have a wrong unit (e.g. requests/sec).

### DIFF
--- a/genai-perf/tests/test_statistics.py
+++ b/genai-perf/tests/test_statistics.py
@@ -1,0 +1,82 @@
+# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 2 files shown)


genai-perf/tests/test_statistics.py
lkomali, line 1: The different tests are doing the same for different types of metrics which is checking if the units are correctly assigned. So, I think you can implement this as a parametrized test something like `test_metric_units` that has different metric classes as parameters using` @pytest.mark.parametrize `.
The assertions can maybe be put inside a for loop? 
Something like
```
for metric, unit in dict.items():
        assert stats[metric]["unit"] == unit
```

dyastremsky (approved):
Great work! Could you make the pull request title more descriptive (e.g. "Fix metric unit bug for request_goodput")?

genai-perf/tests/test_statistics.py
nv-hwoo: sure, I have updated the test to use pytest parameters

nv-hwoo:
@dyastremsky done!

lkomali (approved):

PR #181, opened 2024-11-14
Fixing IWYU problem, causing a build issue.

nicolasnoble:
We probably had `string` being included transitively, which is basically UB.

See https://gitlab-master.nvidia.com/dl/triton/perf_analyzer_ci/-/jobs/122143789#L4216


dyastremsky (approved):

nicolasnoble:
I think @fpetrini15 has a similar PR with even more fixes pending. Should I close this one, @fpetrini15 ?

PR #182, opened 2024-11-15
Add support for the current CLI to PA Config Generator

nv-braf:
As a stopgap until we get a config file support in place I've added support for integrating the current CLI into the PA config generator class.

We now take in all the user CLI settings, and use these to create the command line (along with the current parameter, like concurrency, from the sweep generator). We also store the relevant CLI information in the checkpoint file.

This replaces some of the functionality that is currently found in `parser.py` - changes to that file (along with main and wrapper) will be checked in at a later time once we are closer to releasing the branch.

Along with the unit tests, I have a live test running a concurrency sweep successfully!


genai-perf/genai_perf/config/generate/perf_analyzer_config.py
```diff
@@ -63,14 +273,33 @@ def get_parameters(self) -> Dict[str, Any]:
         """
         return self._parameters
 
+    def get_obj_args(self) -> Namespace:
```
nv-hwoo, line 276: Is the new method used in the code?

genai-perf/genai_perf/config/generate/perf_analyzer_config.py
```diff
@@ -34,14 +89,36 @@ def __init__(
...
+        if not hasattr(self._args, "subcommand"):
+            return
+
+        self._cli_args += self._add_required_args(args)
+        self._cli_args += Profiler.add_protocol_args(args)
+        self._cli_args += Profiler.add_inference_load_args(args)
+        self._cli_args += self._add_misc_args(args)
+
+        if extra_args:
+            self._cli_args += self._add_extra_args(extra_args)
```
nv-hwoo, line 120: The check if `extra_args` is empty/null is done on both before and within the call to `_add_extra_args`, I think we can safely remove this check (or vice versa).

genai-perf/genai_perf/config/generate/perf_analyzer_config.py
nv-hwoo, line 1: A lot of new functions introduced here seems to be from the `parser.py`. Are the future PRs going to merge the duplicate codes so that we don't need to keep track of two places to change?

nv-braf: Yes - that's the hope, but in order to make that happen we'd have to update the other subcommands (profile/compare) to use these libraries.

genai-perf/genai_perf/config/generate/perf_analyzer_config.py
nv-braf: Not in this code, but it is used in the main code that runs analyze live.

genai-perf/genai_perf/config/generate/perf_analyzer_config.py
nv-braf: Good catch. I've removed this redundancy.

genai-perf/genai_perf/config/generate/perf_analyzer_config.py
nv-braf: The overarching goal with the dev branch (analyze_subcommand_phase1)  is to change as little as possible with the existing subcommands/main code, so that it merges cleanly and doesn't break any existing functionality.

nv-hwoo (approved):

PR #183, opened 2024-11-15
Add function to set end_id and apply chat template for GAP Triton In-Process

tedzhouhk:
Add two functions for GAP Triton In-Process:
1. Pass EOS token as end_id so that output can stop at EOS.
1. Apply chat template so that OSL is accurate.
1. A small bug fix for empty first/last response in triton engine.

### DIFF
--- a/genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
@@ -87,4 +87,5 @@ class TensorRTLLMEngineConverter(BaseConverter):
         for key, value in config.extra_inputs.items():
             if key == "triton_converter_set_end_id" and value:
                 payload["end_id"] = [config.tokenizer._tokenizer.eos_token_id]
-            payload[key] = [value]
+            else:
+                payload[key] = [value]
--- a/genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
@@ -55,7 +55,7 @@ class TensorRTLLMEngineConverter(BaseConverter):
                 if not config.apply_chat_template:
                     token_ids = config.tokenizer.encode(row.texts[0])
                 else:
-                    token_ids = config.tokenizer.apply_chat_template([{"role": "user", "content": row.texts[0]}])
+                    token_ids = config.tokenizer.apply_chat_template(row.texts[0])
                 payload = {
                     "input_ids": {
                         "content": token_ids,
--- a/genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
@@ -80,6 +80,8 @@ class TensorRTLLMEngineConverter(BaseConverter):
             payload["request_output_len"] = [num_tokens]
             if config.output_tokens_deterministic:
                 payload["min_length"] = [num_tokens]
+        if config.set_end_id:
+            payload["end_id"] = config.tokenizer._tokenizer.eos_token_id
 
         for key, value in config.extra_inputs.items():
             payload[key] = [value]
--- a/genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
@@ -81,7 +81,7 @@ class TensorRTLLMEngineConverter(BaseConverter):
             if config.output_tokens_deterministic:
                 payload["min_length"] = [num_tokens]
         if config.set_end_id:
-            payload["end_id"] = config.tokenizer._tokenizer.eos_token_id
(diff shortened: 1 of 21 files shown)


genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
```diff
@@ -80,6 +83,8 @@ def _add_request_params(self, payload: Dict, config: InputsConfig) -> None:
             payload["request_output_len"] = [num_tokens]
             if config.output_tokens_deterministic:
                 payload["min_length"] = [num_tokens]
+        if config.set_end_id:
+            payload["end_id"] = [config.tokenizer._tokenizer.eos_token_id]
```
nv-hwoo, line 87: I am a bit concerned about adding a new CLI option for this specific use cases. Since we try to avoid adding too much options to the tool, and this can be achieved through the `--extra-inputs <name>:<value>` as well.
cc @dyastremsky

genai-perf/genai_perf/parser.py
```diff
@@ -571,6 +571,21 @@ def _add_image_input_args(parser):
...
+    input_group.add_argument(
+        "--triton-converter-set-end-id",
+        action="store_true",
+        required=False,
+        help="If specified, the input to trtllm engines in triton server will "
+        "contain end_id set to EOS token."
+    )
+
+    input_group.add_argument(
+        "--triton-converter-apply-chat-template",
```
nv-hwoo, line 583: `apply_chat_template` I think could be a useful cli option to add: https://huggingface.co/docs/transformers/main/en/chat_templating

But I would go for adding it in a more generic way to support chat_template for any use cases, not just for triton in-process.

tedzhouhk: Please correct me if I'm wrong. Is chat template only necessary to benchmark raw engines? Benchmarking through api endpoints does not need to add chat template in GAP, right?

nv-hwoo: Yes, you're right. I guess what I wanted to say was we want to make it more generic so that the option is not limited to only triton in-process although the current use case is just triton in-process. This way we have a stronger reason to add this to our cli options.

tedzhouhk: Got it. Does GAP have any other route to benchmark raw engines that might need chat templates? I am not familiar with the codebase and it would be great if someone from your team has some BW to help.

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
matthewkotila: Yeah if this can be done via [--extra-inputs](https://github.com/triton-inference-server/perf_analyzer/tree/main/genai-perf#--extra-inputs-str), let's skip this change.

tedzhouhk: It is moved to --extra-inputs. however, code change is still necessary unless use directly provide eos token id (instead of fetching from tokenizer). Please let me know if current approach looks good. Thanks.

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
```diff
@@ -282,8 +282,11 @@ def _get_tensorrtllm_engine_token_counts(
         for r in res_outputs:
             if isinstance(r["output_ids"], list):
                 token_ids += r["output_ids"]
-            else:
+            elif isinstance(r["output_ids"], int):
                 token_ids.append(r["output_ids"])
+            else:
+                # for the empty first/last responses
+                token_ids.append(0)
```
tedzhouhk, line 289: @matthewkotila added a bug fix for empty first/last response (it will be and empty string and cause error).

dyastremsky (commented):
Good start to this PR. I like the idea of making the chat template generic for use with any/all endpoints.

We should also add unit testing to support these changes.

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
```diff
@@ -82,4 +85,7 @@ def _add_request_params(self, payload: Dict, config: InputsConfig) -> None:
                 payload["min_length"] = [num_tokens]
 
         for key, value in config.extra_inputs.items():
-            payload[key] = [value]
+            if key == "triton_converter_set_end_id" and value:
```
dyastremsky, line 88: You could probably shorten this key, if you wanted. e.g. "set_end_id"

genai-perf/genai_perf/inputs/inputs_config.py
```diff
@@ -142,3 +142,9 @@ class InputsConfig:
 
     # Seed used to generate random values
     random_seed: int = DEFAULT_RANDOM_SEED
+
+    # whether to set end_id in triton converter
+    set_end_id: bool = False
```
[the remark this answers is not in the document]
tedzhouhk: good catch, i'll remove this

dyastremsky:
Closing, as this functionality was incorporated into [another pull request](https://github.com/triton-inference-server/perf_analyzer/pull/253), which has been merged.

PR #184, opened 2024-11-15
Fix L0_CApi segmentation fault when using cuda

lkomali:
Fix segmentation fault  in Triton CAPI mode, when shared memory is set to cuda.

### DIFF
--- a/src/client_backend/triton_c_api/triton_loader.cc
+++ b/src/client_backend/triton_c_api/triton_loader.cc
@@ -42,6 +42,11 @@
 #include "c_api_infer_results.h"
 #include "scoped_defer.h"
 
+#ifdef TRITON_ENABLE_GPU
+#include "../../cuda_runtime_library_manager.h"
+#endif
+
+
 namespace triton { namespace perfanalyzer { namespace clientbackend {
 namespace tritoncapi {
 namespace {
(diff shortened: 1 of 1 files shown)


src/client_backend/triton_c_api/triton_loader.cc
```diff
@@ -1043,22 +1048,25 @@ TritonLoader::GetOutputs(
...
-    } else {
-      data_copy.reserve(byte_size);
-      std::copy(
-          static_cast<const uint8_t*>(base),
-          static_cast<const uint8_t*>(base) + byte_size,
-          std::back_inserter(data_copy));
+      base = static_cast<const uint8_t*>(base) + 4;
+      byte_size -= 4;
+    }
+#ifdef TRITON_ENABLE_GPU
```
matthewkotila, line 1055: I don't think the existence of this macro means that the user was requesting cuda shared memory. You'll need to check somehow that `--shared-memory=cuda` was passed in the call to PA.

src/client_backend/triton_c_api/triton_loader.cc
```diff
@@ -1043,22 +1048,25 @@ TritonLoader::GetOutputs(
...
+    CUDARuntimeLibraryManager cuda_manager;
```
matthewkotila, line 1056: Instantiating the library object like this will call dlopen each time. We don't want the library loaded each time GetOutputs() is called. Can you make this a data member like how it's used in InferDataManagerShm?:

https://github.com/triton-inference-server/perf_analyzer/blob/30ccf3201c450b7108faae71403e409e0b91481a/src/infer_data_manager_shm.h#L167

nicolasnoble: So, dlopen won't open the same .so file twice if it hasn't been unloaded: it's reference counted. The waste isn't as bad as you think that is.

src/client_backend/triton_c_api/triton_loader.cc
```diff
@@ -1043,17 +1048,20 @@ TritonLoader::GetOutputs(
...
     if (data_type == "BYTES" && byte_size >= 4) {
-      data_copy.reserve(byte_size - 4);
-      std::copy(
-          static_cast<const uint8_t*>(base) + 4,
-          static_cast<const uint8_t*>(base) + byte_size,
-          std::back_inserter(data_copy));
+      base = static_cast<const uint8_t*>(base) + 4;
+      byte_size -= 4;
+    }
+    if (memory_type == TRITONSERVER_MEMORY_GPU) {
```
matthewkotila, line 1055:
```suggestion

    if (memory_type == TRITONSERVER_MEMORY_GPU) {
```

matthewkotila (approved):

PR #187, opened 2024-11-21
Update README.md

matthewkotila:


genai-perf/README.md
```diff
@@ -99,6 +99,8 @@ you'll need to install the Perf Analyzer binary:
 
 ```bash
 pip install tritonclient
+
+sudo apt update && sudo apt install -y --no-install-recommends libb64-0d
```
ganeshku1, line 103: Is the sudo expected here?

matthewkotila: Yes, sudo is generally needed when running apt as non-root user.

ganeshku1: As far as i recall we did not had that in the past version of the doc.
Approving this change as i anticipating we will be mostly reverting this change after we release 24.11.

ganeshku1 (approved):

genai-perf/README.md
matthewkotila: We might have left it off accidentally. See here from the official ubuntu documentation showing that sudo is needed for apt:

https://ubuntu.com/server/docs/package-management#installing-packages

PR #188, opened 2024-11-21
Fix L0_CApi segmentation fault when using cuda (#184)

mc-nv:
* fix segfault while copying data to vector when using cuda memory

* Fix comments

* Update src/client_backend/triton_c_api/triton_loader.cc

---------

### DIFF
--- a/src/client_backend/triton_c_api/triton_loader.cc
+++ b/src/client_backend/triton_c_api/triton_loader.cc
@@ -42,6 +42,11 @@
 #include "c_api_infer_results.h"
 #include "scoped_defer.h"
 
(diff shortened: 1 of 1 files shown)


krishung5:
Could you please run the pre-commit hook to pass the check?

yinggeh:
Please fix pre-commit error. LGTM otherwise.

mc-nv:
I won't run any pre-commit hooks, it's a cherry-pick.

krishung5:
@mc-nv I think it's fine if not fixed for now, but will need to address the pre-commit hook later when advancing main to 24.11 changes later.

krishung5 (approved):

PR #194, opened 2024-11-23
Disable TRT-LLM echo by default

dyastremsky:
Now that TRT-LLM supports disabling echo as a request-level parameter, disable it by default for parity with Triton + vLLM benchmarking.

### DIFF
--- a/genai-perf/README.md
+++ b/genai-perf/README.md
@@ -388,9 +388,8 @@ a request in order. Random means that assignment is uniformly random
 
 ##### `--backend {tensorrtllm,vllm}`
 
-When using the "triton" service-kind, this is the backend of the model. For the
-TRT-LLM backend, you currently must set `exclude_input_in_output` to true in the
-model config to not echo the input tokens in the output. (default: tensorrtllm)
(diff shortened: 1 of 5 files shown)


genai-perf/README.md
matthewkotila, line 1: Could you give a brief refresher on what "echoing" means in this context?

dyastremsky: Absolutely, great question. TRT-LLM by default would echo whatever was in the request in the output. For example:
1. Request: "Hello, what is my name?"
2. Response: "Hello, what is my name? My name is Matt."

So the token counts were not what people were expecting. TRT-LLM did not have a way to configure that setting at the request level, only the model level (in the model configuration at load time). This threw off the metrics, including the output token length, and we asked them to add "exclude_input_in_output" as a request-level parameter, just like vLLM does. They added it a few months ago, so we can finally send the parameter to default to not receive this "echo" of input in the output. So if someone requests an output sequence length of 100 tokens, they get 100 tokens rather than 100 tokens + all the input tokens.

matthewkotila (approved):

nv-braf (approved):

PR #196, opened 2024-11-23
Add request-count and warmup-request-count

dyastremsky:
Add request count and warmup-request-count as native GenAI-Perf arguments. Also, small clean up of some of our documentation and conditional checks.


genai-perf/genai_perf/parser.py
```diff
@@ -434,26 +439,42 @@ def _add_input_args(parser):
...
+        "'image' field 'in JSONL format. Example: {\"text\":"
+        ' "Your prompt here"}\'. To use synthetic files for a converter that '
+        "needs multiple files, prefix the path with 'synthetic:', followed "
+        "by a comma-separated list of filenames. The synthetic filenames "
+        "should not have extensions. For example, "
+        "'synthetic:queries,passages'. ",
     )
 
     input_group.add_argument(
+        "--num-payloads",
```
matthewkotila, line 471: Thoughts on `--num-dataset-entries`?

templates/genai-perf-templates/README_template
```diff
@@ -476,7 +480,13 @@ when `--output-tokens-mean` is provided. (default: `0`)
 
 The seed used to generate random values. (default: `0`)
 
+##### `--request-count <int>`
+
+The number of requests to use for measurement. By default, the measurement
+will continue until stabilization is detected. (default: 0)
```
matthewkotila, line 486: To me, this reads confusingly.

```suggestion
The number of requests to use for measurement. By default, the benchmark does not
terminate based on a request count--instead it continues until stabilization is
detected. (default: 0)
```

genai-perf/genai_perf/parser.py
dyastremsky: I'd prefer the first, since this is used for the input file approach too.

@debermudez @nv-hwoo, any thoughts on the best name for this arg?

nv-hwoo: Actually, do we want to use this for both synthetic and input file approaches? When the option is used for synthetic, it sets the total number of input data being generated, whereas in input file approach, it defines how many data we want to use/sample from the given input file. Feels like we need a separate option for each.. (although we are trying to cut-down the number of options 😂 )

Anyways to your question, I think I like
- `--input-data-size`
- `--input-sample-size`

genai-perf/genai_perf/parser.py
```diff
@@ -434,26 +439,42 @@ def _add_input_args(parser):
...
         default=None,
         required=False,
         help="The input file or directory containing the content to use for "
-        "profiling. To use synthetic files for a converter that needs "
-        "multiple files, prefix the path with 'synthetic:', followed by a "
-        "comma-separated list of filenames. The synthetic filenames should "
-        "not have extensions. For example, 'synthetic:queries,passages'. "
-        "Each line should be a JSON object with a 'text' or 'image' field "
-        'in JSONL format. Example: {"text": "Your prompt here"}',
+        "profiling. Each line should be a JSON object with a 'text' or "
```
nv-hwoo, line 461: Should we also update the CLI doc for `--input-file` as well?

genai-perf/README.md
```diff
@@ -479,6 +484,11 @@ when `--output-tokens-mean` is provided. (default: `0`)
 
 The seed used to generate random values. (default: `0`)
 
+##### `--request-count <int>`
+
+The number of requests to use for measurement. By default, the measurement
+will continue until stabilization is detected. (default: 0)
```
nv-hwoo, line 490:
```suggestion
will continue until stabilization is detected. (default: `0`)
```

genai-perf/genai_perf/parser.py
matthewkotila: @nv-hwoo I actually think it makes sense to only have one option. They're both functionally doing the same thing. Picking how many dataset entries to generate/use for the purposes of what each request's inputs are, round-robin style.

nv-hwoo: NooOOoOoOOooOoo!! I AM RIGHT 😤 
Yeah they're functionally the same, I agree.

dyastremsky: Thank you both for the feedback! I addressed all your comments and updated the code. It is currently running through CI.

genai-perf/genai_perf/inputs/inputs_config.py
```diff
@@ -131,8 +131,8 @@ class InputsConfig:
     # Synthetic Prompt Generation Parameters
     ########################################
 
-    # The number of synthetic output prompts to generate
-    num_prompts: int = DEFAULT_NUM_PROMPTS
+    # The number of dataset entries to generate and use as the payload pool
```
matthewkotila, line 134: Payload pool--outstandingly clear description--nice!!

matthewkotila (dismissed):

dyastremsky:
Thanks for the approval, Matt!

One file did not get committed. If you have a chance whenever, a re-approval would be much appreciated!

matthewkotila (approved):

PR #197, opened 2024-11-25
Capture Telemetry Records in checkpoint

nv-braf:
Updated the telemetry records to reflect what is actually being captured in GAP and added a method (very similar to what is in statistics) to convert the telemetry metrics dict into Records that can be captured/checkpointed in the RunConfigMeasurement class.

Finally, I hooked this up and ran it live to confirm it is working as intended. Unlike with statistics, I couldn't find a unit test that I could leverage to test this, so we will rely on our CI to provide coverage for this new method.


lkomali:
Can you add a picture of what this PR is trying to add in the description so that it's easier for me to visualize?

lkomali (approved):
Looks good to me

PR #201, opened 2024-11-27
Handle missing data in console and csv exporter

nv-hwoo:
When some statistics are missing (for any reason), handle it by displaying "N/A" instead of throwing an error to the user.

The console output will contain "N/A" if some values are missing
image

The CSV output will also contain "N/A"
```
Metric,avg,min,max,p99,p95,p90,p75,p50,p25
Request Latency (ms),205.80,205.40,207.50,207.37,206.84,206.17,205.76,205.58,205.48
Output Sequence Length,N/A,11.00,13.00,13.00,13.00,13.00,12.00,11.00,11.00
Input Sequence Length,N/A,N/A,N/A,N/A,N/A,N/A,N/A,N/A,N/A

Metric,Value
Output Token Throughput (per sec),56.27
Request Throughput (per sec),4.85
```

The JSON exporter does not require any handling since it directly takes the statistics dictionary and saves it without any modification. So any missing data won't be in the final JSON file.

### DIFF
--- a/src/client_backend/openai/http_client.h
+++ b/src/client_backend/openai/http_client.h
@@ -33,8 +33,8 @@
 #include <map>
 #include <memory>
 #include <mutex>
-#include <thread>
 #include <string>
+#include <thread>
 
 namespace triton { namespace perfanalyzer { namespace clientbackend {
 namespace openai {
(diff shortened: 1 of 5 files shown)


src/client_backend/openai/http_client.h
nv-hwoo, line 1: For some reason, pre-commit was complaining about the order of the includes so I had to change it.

dyastremsky (approved):

dyastremsky:
Great work, Hyunjae! Looks good.

PR #204, opened 2024-12-03
Use ignore_eos when benchmarking Triton + vLLM deterministically

dyastremsky:
When using the --output-tokens-mean-deterministic flag, use the arg "ignore_eos" for Triton + vLLM, since it supports it and it could result in more accurate benchmarking.


debermudez (approved):

dyastremsky:
Closing, since it already works without ignore-eos after user testing. No need to send extra data if the current approach accomplishes the expected behavior.

PR #205, opened 2024-12-04
Unit test for CustomRequestScheduleManager

lkomali:
This PR
1. Creates Unit test for CustomRequestScheduleManager
2. Corrects the schedule used while creating workers (used schedule instead of scaled_schedule)
3. Fixes segmentation fault that is occurred when schedule.size() < 4.
4. Modifies the param variable passed, earlier was passing a shared_ptr to params, but now modified to passing params object to make the unit test work.
5. Update GenerateSchedule function to only take request_rate, as schedule is an attribute of the class and need not be passed.


src/custom_request_schedule_manager.h
```diff
@@ -55,7 +55,8 @@ class CustomRequestScheduleManager : public RequestRateManager {
   /// object
   ///
   static cb::Error Create(
-      const pa::PAParamsPtr& params, const std::shared_ptr<ModelParser>& parser,
+      const PerfAnalyzerParameters& params,
```
matthewkotila, line 58: I'm a fan of passing references over smart pointers, but why was this change needed?

lkomali: For the unit test, I don't remember quite correctly, but the unit test couldn't access the pointer.
I will take a look later to refresh my memory and correctly define what was going wrong.

src/test_custom_request_schedule_manager.cc
```diff
@@ -0,0 +1,143 @@
...
+          std::chrono::duration<float>(scaled_value));
+
+      expected_timestamps.push_back(timestamp);
+    }
+
+    size_t request_index = 0;
+    for (auto worker : workers_) {
+      auto timestamp = std::dynamic_pointer_cast<RequestRateWorker>(worker)
+                           ->GetNextTimestamp();
+      REQUIRE(timestamp.count() == expected_timestamps[request_index].count());
```
dyastremsky, line 75: Are the unit tests only testing that the correct number of timestamps are provided? Do the unit tests also need to make sure that the timestamps are the ones expected?

lkomali: As far as I understand, it's checking if the timestamps calculated in the test match the timestamps the workers are scheduled at. 
I have taken reference to this [check](https://github.com/triton-inference-server/perf_analyzer/blob/main/src/test_request_rate_manager.cc#L149) in **test_request_rate_manager.cc**
The .count() gives duration as an integer. 
Essentially, we are checking that the requests are scheduled at expected timestamps.

src/test_custom_request_schedule_manager.cc
```diff
@@ -0,0 +1,143 @@
...
+
+#include "command_line_parser.h"
+#include "custom_request_schedule_manager.h"
+#include "doctest.h"
+#include "request_rate_worker.h"
+#include "test_load_manager_base.h"
+
+namespace triton::perfanalyzer {
+
+/// Class to test CustomRequestScheduleManager
```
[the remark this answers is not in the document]
lkomali: I will remove this line before merging.

src/test_custom_request_schedule_manager.cc
dyastremsky: I see, thanks for explaining. I believe you can check if the timestamps (std::chrono::nanoseconds) are equal directly and I don't believe the .count() adds anything to this check, but I'm okay with us leaving it if that's preferred. It's not worth slowing down this getting in.

src/CMakeLists.txt
matthewkotila, line 1: PR LGTM, just waiting on other reviewers to suggest approval.

dyastremsky: Testing portion of the PR LGTM as well.

matthewkotila (approved):

PR #207, opened 2024-12-06
Analyze: Print CSV reports

nv-braf:
Will now write CSV reports to the artifacts directory (identical to what is done when we profile). JSON and screen are skipped.
A new summary report is created in the CWD - called `analyze_export_genai_perf.csv`.

Here is an example of it's output:

image

The first table is: config name + stimulus + perf_metrics (p99)
The second table is: config name + gpu_metrics


genai-perf/genai_perf/config/generate/perf_analyzer_config.py
```diff
@@ -311,6 +317,37 @@ def get_obj_args(self) -> Namespace:
...
+        """
+        infer_type = self.get_inference_type()
+        infer_cmd_option = (
+            "--concurrency-range"
+            if infer_type == InferenceType.CONCURRENCY
+            else "--request-rate-range"
+        )
+
+        cmd = self.create_command()
+        infer_value_index = cmd.index(infer_cmd_option) + 1
```
debermudez, line 346: This feels brittle.
Would a dictionary be safer as the code base grows?

genai-perf/genai_perf/record/record.py
```diff
@@ -15,13 +15,23 @@
...
 from genai_perf.exceptions import GenAIPerfException
 from genai_perf.types import RecordValue
 
 
+@dataclass(frozen=True)
+class ReductionFactor:
+    NS_TO_MS = 6
+    NJ_TO_MJ = 6
+    B_TO_GB = 9
+    PERCENTAGE = -2
```
debermudez, line 31: Is the negative intentional here?
It doesnt follow the pattern of the rest of them, so I figured i would check.

genai-perf/genai_perf/record/types/gpu_power_limit_base.py
```diff
@@ -24,6 +25,7 @@ class GPUPowerLimitBase(IncreasingGPURecord):
     """
 
     base_tag = "gpu_power_limit"
+    reduction_factor = ReductionFactor.NONE
```
debermudez, line 28: why import no reduction factor?
is the reduction handled in a base class?
or is it for uniformity?

debermudez (commented):
couple of comments, overall looks good

genai-perf/genai_perf/record/record.py
nv-braf: Yes, in this case we are going from something like 0.34 to 34% (so it's an anti-reduction!)

genai-perf/genai_perf/record/types/gpu_power_limit_base.py
nv-braf: This is the base class. Every base class needs a `base_tag` and a `reduction_factor`.  In the case of GPU Power Limit (as well as ISL/OSL, request throughput, etc...) we don't need to reduce the value. This both makes the function work and serves as documentation.

genai-perf/genai_perf/config/generate/perf_analyzer_config.py
debermudez: Then I think we have to come up with a better solution at some point but for now, we move forward. Nice work.

genai-perf/genai_perf/record/types/gpu_power_limit_base.py
debermudez: Gotcha. I missed that in the file name.

PR #209, opened 2024-12-06
Support input type payload for timing strategy

lkomali:
As part of timing strategy, this PR adds a new type for --input-file arg.
With this change, user can set --input-file=payload:"path-to-timing-file" to pass the file that has request level timing information.
In the input file, users can set the time at which request should be sent.
For example, {"timestamp":"1"} will send request at 1 second.

### DIFF
--- a/genai-perf/genai_perf/parser.py
+++ b/genai-perf/genai_perf/parser.py
@@ -351,16 +351,28 @@ def parse_goodput(values):
 
 
 def _infer_prompt_source(args: argparse.Namespace) -> argparse.Namespace:
-
     args.synthetic_input_files = None
+    args.payload_input_file = None
 
     if args.input_file:
-        if str(args.input_file).startswith("synthetic:"):
+        input_file_str = str(args.input_file)
+        if input_file_str.startswith("synthetic:"):
             args.prompt_source = ic.PromptSource.SYNTHETIC
-            synthetic_input_files_str = str(args.input_file).split(":", 1)[1]
+            synthetic_input_files_str = input_file_str.split(":", 1)[1]
             args.synthetic_input_files = synthetic_input_files_str.split(",")
             logger.debug(
                 f"Input source is synthetic data: {args.synthetic_input_files}"
+            )
+        elif input_file_str.startswith("payload:"):
+            args.prompt_source = ic.PromptSource.PAYLOAD
+            payload_input_file_str = input_file_str.split(":", 1)[1]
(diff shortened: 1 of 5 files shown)


dyastremsky:
Should this be targeting a feature branch or is this the whole feature? Also, could you update the branch off of main?

lkomali:
@dyastremsky Yep sorry, should base off of a feature branch. Thanks for catching that.

genai-perf/genai_perf/parser.py
```diff
@@ -351,17 +351,29 @@ def parse_goodput(values):
 
 
 def _infer_prompt_source(args: argparse.Namespace) -> argparse.Namespace:
-
     args.synthetic_input_files = None
+    args.payload_input_file = None
 
     if args.input_file:
+        input_file_str = str(args.input_file)
         if str(args.input_file).startswith("synthetic:"):
```
dyastremsky, line 359: You may want to reuse the arg above (input_file_str). 

Also, this function is starting to get long. I'm thinking we might want to refactor parser sometime soon (CC: @debermudez @nv-hwoo). This function is starting to feel long. I don't think we need to slow down development by breaking this into its own class with functions yet, but I do think we should take a look at how we can restructure parser.py as a whole and break it into a cleanly-organized directory.

dyastremsky (commented):
Changes generally look good! Thanks for providing great testing coverage.

genai-perf/genai_perf/parser.py
```diff
@@ -459,12 +471,17 @@ def _add_input_args(parser):
...
-        "'image' field 'in JSONL format. Example: {\"text\":"
-        ' "Your prompt here"}\'. To use synthetic files for a converter that '
+        "'image' field in JSONL format. Example: {\"text\": "
+        '"Your prompt here"}. To use synthetic files for a converter that '
         "needs multiple files, prefix the path with 'synthetic:', followed "
         "by a comma-separated list of filenames. The synthetic filenames "
         "should not have extensions. For example, "
-        "'synthetic:queries,passages'. ",
+        "'synthetic:queries,passages'. For payload data, prefix the path with 'payload:', "
+        "followed by a JSON string representing a payload object. The payload should "
```
debermudez, line 480: I dont think the rest of this description is clear.
We will need to fix this to reflect our final design

debermudez (approved):
I am not sure that comment is correct but I think it can go in.

PR #212, opened 2024-12-09
Analyze Subcommand

nv-braf:
Analyze is a new subcommand which allows the user to sweep over PA (concurrency, request rate) or GAP (ISL, dataset entries) stimulus ranges using the following new CLI options: `--sweep-type, --sweep-range, --sweep-list`.

In addition to allowing the user to profile multiple configurations with a single command, GAP now saves the results, and has the ability to restore its state via the checkpoint JSON. This provides several advantages: 

- if a crash occurs you can rerun the command and pickup where you left off
- provides an API to access the results which can be used by subsequent subcommands

Individual CSV reports are still generated in each of the artifacts directory locations as well as a top-level summary CSV combining all the profiled runs (showing p99 values only).


nv-hwoo:
@nv-braf I think you forgot to clean up the old commits after rebase.

nv-braf:
> @nv-braf I think you forgot to clean up the old commits after rebase.

I don't think I did. This is a merge of all the commits that I've made for the past several months (to the analyze_branch) onto main.

nv-hwoo:
My bad. I saw a few new commits mixed with old PR commits so I assumed this was another separate PR going into a feature branch.

debermudez (approved):
All of these changes were reviewed throughout the development process. The final changes are just merge conflicts that were solved. This is ready to go.

PR #206, opened 2024-12-05
feat: Add Generate Endpoint Support

nnshah1:
Adds support for endpoint-type 'generate' (https://docs.nvidia.com/deeplearning/triton-inference-server/user-guide/docs/protocol/extension_generate.html).

Adds new endpoint types:

1) generate
2) kserve
3) tritonllm_engine

PR #214, opened 2024-12-10
Add unit testing for Triton Generate converter

dyastremsky:
This converter was added in a [recent PR](https://github.com/triton-inference-server/perf_analyzer/pull/206/files#diff-531d67ceb8ef974e32746d3e8b7b2cd958bac761568f74220ef436b018cacf99R210). Add unit testing for this converter.

I also added a conditional check for --output-tokens-mean-deterministic, because the other Triton endpoints support it but this one cannot (because it supports both vLLM and TRT-LLM, which have different argument names for min_tokens/min_length).


genai-perf/tests/test_triton_generate_converter.py
```diff
@@ -0,0 +1,189 @@
...
+                        }
+                    ]
+                }
+            ]
+        }
+
+        assert result == expected_result
+
+    def test_convert_empty_dataset(self):
+        generic_dataset = GenericDataset(files_data={})
```
nv-hwoo, line 178: Is empty dataset possible with our current GAP? If so, should we let it pass without any warning/error?

dyastremsky: Theoretically, yes. The retrievers decide which data to provide, the converters turn it into the right format. If a retriever  provides no data, it'd be good if the converter doesn't fail spectacularly. I don't know if it's possible with our current retrievers (maybe if the input file is empty), but it could definitely be possible when we add a new retriever.

I was divided on whether to include this test. I think it's very optional, but we have it for a few other converters. We could always add a check if an empty dataset was retrieved before we call the converter to return an exception, but I don't think it's necessary now. If we want to do so, I'd prefer to push it into its own ticket/PR.

Would you prefer for this test to be removed?

dyastremsky: Sounds good, removed it!

nv-hwoo (approved):

PR #217, opened 2024-12-12
Add time-to-second-token metric

dyastremsky:
In streaming scenarios, time to second token can be a useful metric for benchmarking LLMs. It can provide an easy comparison for the prefill vs decode stages.

This PR adds this metric to GenAI-Perf.

Example output:
image


nv-braf (approved):
Looks good!

PR #218, opened 2024-12-12
Documentation for the Analyze Subcommand

nv-braf:
Documentation for the Analyze subcommand along with a bug fix to determining inference type that I found when creating the example CSV output.

### DIFF
--- a/genai-perf/docs/analyze.md
+++ b/genai-perf/docs/analyze.md
@@ -1,0 +1,240 @@
+<!--
+Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+ * Redistributions of source code must retain the above copyright
+   notice, this list of conditions and the following disclaimer.
+ * Redistributions in binary form must reproduce the above copyright
+   notice, this list of conditions and the following disclaimer in the
+   documentation and/or other materials provided with the distribution.
+ * Neither the name of NVIDIA CORPORATION nor the names of its
+   contributors may be used to endorse or promote products derived
+   from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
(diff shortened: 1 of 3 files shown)


genai-perf/genai_perf/subcommand/analyze.py
```diff
@@ -392,7 +392,9 @@ def _determine_infer_type(self):
         else:
             if self._args.concurrency:
                 infer_type = "concurrency"
-            else:
+            elif self._args.request_rate:
                 infer_type = "request_rate"
+            else:
+                infer_type = "concurrency"
```
debermudez, line 398: This could instead just check for request rate and set concurrency in its absence since it is the default mode.
Is there a reason that you explicitly check for both?

genai-perf/docs/analyze.md
```diff
@@ -0,0 +1,240 @@
...
+In the CWD a summary report CSV (`analyze_export_genai_perf.csv`) is created. In the first table, each row is a different scenario profiled while the columns show p99 perf metrics. In the second table, the rows are again scenario's profiled, while the columns are p99 GPU telemetry metrics.
+
+### Example Summary Report CSV
+```
+Config Name,Concurrency,ISL,Num Dataset Entries,p99 Time To First Token (ms),p99 Inter Token Latency (ms),p99 Request Latency (ms),p99 Output Sequence Length (tokens),Avg. Output Token Throughput (tokens/sec),Request Throughput (requests/sec)
+gpt2_run_config_2,1,201,200,33.54,7.16,779.75,132.10,149.63,1.32
+gpt2_run_config_1,1,201,150,33.13,7.29,778.62,126.16,147.93,1.32
+gpt2_run_config_0,1,201,100,82.02,7.53,879.20,124.55,145.93,1.30
+
+Config Name,GPU,p99 GPU Power Usage (W),p99 GPU Energy Consumption (MJ),p99 GPU Utilization (%),p99 GPU Memory Used (GB),Avg. GPU Power Limit (W),Avg. GPU Total Memory (GB)
```
debermudez, line 89: Historically, we have had users complain that our csv files are delivered like this.
They don't technically abide by the csv format so they break when trying to use standard libraries to import.
Would it be easy to make this two different csv files? OR was this intentionally designed this way and I dont know the reason?

debermudez (commented):
couple of questions. i like this document overall. nice work

genai-perf/genai_perf/subcommand/analyze.py
nv-braf: No reason other than I feel like this is more explicit in showing intent of the code.

genai-perf/docs/analyze.md
nv-braf: Intentionally designed this way to match the CSV created in the artifact directory.

debermudez: I worry this is going to make it hard for the TMEs to parse the csv data.
It may need to iterate...but they should also just use the checkpoint

nv-braf: It's easy enough to change - I have no problem modifying this to match whatever the TMEs want it to look like.

genai-perf/genai_perf/subcommand/analyze.py
debermudez: I worry this introduces a spot where we can have a bug.

nv-braf: How so? I feel like the code is safer/more future proof this way.

debermudez: There are two places that set concurrency. If we change it at one point, its two places to change.

debermudez: This is ultimately a method that is going to go away in the near future when we move to a different UX.

PR #220, opened 2024-12-16
Fix re-rankings payload schema for NIMs

dyastremsky:
During the inputs modularization work, the payload for benchmarking re-ranking NIMs was broken. Fix the payload schema.

After changes:
image


nv-hwoo (approved):
Thanks for the quick fix! 🚀

nv-braf (approved):
This looks fine. I like that there is unit testing for this, but I'm curious as to who is the "owner" of the formatting for this? How did you discover it was broken (as the unit test would not have failed)?
Is this something that should be tested live (L0)?

PR #221, opened 2024-12-16
Organize unit tests into directories

dyastremsky:
As GenAI-Perf has expanded, so too has its unit testing. There are now many unit tests.

For the unit tests that fall into common buckets, move them into their own directories. This makes it easier to, for example, find or run all of the converter tests. This will have no impact on how unit tests are run otherwise (i.e. running `pytest` will still run all of the tests).


debermudez (approved):
Assuming this is a move and the unit tests pass the CI, this is good to go. Thanks for this clean up @dyastremsky.

PR #225, opened 2024-12-18
Forward error messages from OpenAI client

dyastremsky:
Perf Analyzer returns the error code when there is an OpenAI error returned. It does not return the message. This makes debugging harder for users, since they do not know what went wrong.

This pull request makes it so that PA forwards the error message back to the user.

Before:
image

After:
image

### DIFF
--- a/src/client_backend/openai/openai_client.h
+++ b/src/client_backend/openai/openai_client.h
@@ -65,7 +65,8 @@ class ChatCompletionResult : public InferResult {
   {
     if ((http_code_ >= 400) && (http_code_ <= 599)) {
       return Error(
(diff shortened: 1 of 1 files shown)


debermudez (approved):
This has been a pain point for a long time. The solution was so simple.
Nice work!

rmccorm4 (approved):
Thank you for adding this! 🚀

src/client_backend/openai/openai_client.h
matthewkotila, line 1: Outstanding!!!

matthewkotila (approved):

PR #226, opened 2024-12-18
Add headers as a native arg in GenAI-Perf

dyastremsky:
Headers are increasingly used for authentication. These should be moved from superuser args to GenAI-Perf-native args. This pull request adds the flag `--header`/`-h` for users within GenAI-Perf.

This also adds documentation for Telemetry metrics.

Screenshot of generated PA command with headers:
image

### DIFF
--- a/genai-perf/README.md
+++ b/genai-perf/README.md
@@ -343,12 +343,10 @@ AUTHENTICATION
 
 GenAI-Perf can benchmark secure endpoints such as OpenAI, which require API
 key authentication. To do so, you must add your API key directly in the command.
-At the end of your command, append the below flags. Replace the key with your
-API key. The `--` flag allows arguments to pass directly into Perf Analyzer in
-superuser mode. The `-H` flag is used to add HTTP headers.
+Add the following flag to your command.
 
 ```bash
--- -H "Authorization: Bearer ${API_KEY}" -H "Accept: text/event-stream"
+-h "Authorization: Bearer ${API_KEY}" -H "Accept: text/event-stream"
 ```
 
 </br>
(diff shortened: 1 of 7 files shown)


genai-perf/README.md
```diff
@@ -377,6 +375,13 @@ the inference server.
 
 </br>
 
+### Telemetry Metrics
```
dyastremsky, line 378: Note that I did not provide example metrics by design. That would make the README much longer for this one section. 

If we want to provide example output, we should do it in its own tutorial that goes more in depth. We should also weigh doing so versus any extra maintenance required.

debermudez: I would say that we need that doc.
Not gating for this PR but we should make that ticket to make sure we follow up.

genai-perf/tests/test_exporters/test_json_exporter.py
```diff
@@ -253,6 +253,7 @@ class TestJsonExporter:
           "tokenizer_trust_remote_code": false,
           "verbose": false,
           "goodput": null,
+          "header": null,
```
debermudez, line 256: Does it make sense to make this null?
I can see this being used often, based on the same logic that we promoted it. We likely need to ensure its exporting correctly.
What do you think?

genai-perf/README.md
dyastremsky: @lkomali Would you be able to make that Jira ticket? If you could include the details about what a tutorial for Telemetry should include, that'd be helpful (e.g. if we're expanding it past Triton, is it different for Triton vs vLLM).

genai-perf/tests/test_exporters/test_json_exporter.py
dyastremsky: I think this is the default value. Are you asking for it to be exported differently or to set the header in the unit test? I don't expect it to be exported this way, just tested in parallel with the way we test the other fields.

In general, we might want to consider how to unit test this exporter better. Right now, we have a few unit tests to cover all of the args and we just sort of add the args to this expected JSON result every time we add an arg to the parser.

debermudez: like many things, this quickly outgrew its initial design.
im fine leaving this alone for now but we need to revisit once we move to a config file.

debermudez (approved):

genai-perf/README.md
dyastremsky: Created TPA-895 to track this work.

lkomali: Sorry, I didn't check your comment yesterday. Thanks for creating the ticket @dyastremsky.

PR #228, opened 2024-12-19
Add Option to Sweep PA batch size

nv-braf:
Adds an option to sweep PA batch size to `--sweep-type` when running the `analyze` subcommand.
This was an oversight on my part as the logic for this was already in place in the sweep objective generator.


debermudez (approved):
We will need a doc soon that covers how to do this for users and for us to learn.
This PR will ungate Nic quickly. Nice job!

PR #229, opened 2024-12-20
Modify PA to disallow user attempt to use concurrency, warmup with --schedule

lkomali:
Modify PA to disallow user attempt to use concurrency, warmup with schedule

### DIFF
--- a/src/command_line_parser.cc
+++ b/src/command_line_parser.cc
@@ -1679,6 +1679,7 @@ CLParser::ParseCommandLine(int argc, char** argv)
           }
           params_->schedule = schedule;
           params_->request_count = schedule.size();
+          params_->using_request_rate_range = true;
           break;
         }
         case 'v':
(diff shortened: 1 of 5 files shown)


src/command_line_parser.cc
matthewkotila, line 1: PR title is a bit misleading. Support is not being removed. This is fixing bugs where a user could attempt to combine options that are _already not_ currently supported.

src/inference_profiler.cc
```diff
@@ -623,12 +623,15 @@ InferenceProfiler::Profile(
   meets_threshold = true;
 
 
-  RETURN_IF_ERROR(dynamic_cast<RequestRateManager*>(manager_.get())
-                      ->PerformWarmup(request_rate, warmup_request_count));
+  if (dynamic_cast<CustomRequestScheduleManager*>(manager_.get()) == nullptr) {
+    std::cout << "Perform Warmup" << std::endl;
```
matthewkotila, line 627: This warmup print shouldn't be here

src/inference_profiler.cc
```diff
@@ -623,12 +623,15 @@ InferenceProfiler::Profile(
...
+    RETURN_IF_ERROR(dynamic_cast<RequestRateManager*>(manager_.get())
+                        ->PerformWarmup(request_rate, warmup_request_count));
+    std::cout << "Request Rate: " << request_rate
+              << " inference requests per second" << std::endl;
+  }
```
matthewkotila, line 632: Can this be rewritten to be more readable? I can't easily understand what's going on. Too many negations and special conditions and logic that happens regardless.

matthewkotila (approved):
Looks great!

PR #230, opened 2024-12-21
Update readme with new pip instructions and reorganize

debermudez:
Update the Readme to use the refined pip installation instructions.
Move the pip installation instructions to the top of the readme.


genai-perf/README.md
```diff
@@ -73,45 +73,34 @@ INSTALLATION
...
+### Install Perf Analyzer (Ubuntu 24.04, Python 3.10+)
 
 ```bash
-export RELEASE="24.10"
-
-docker run -it --net=host --gpus=all  nvcr.io/nvidia/tritonserver:${RELEASE}-py3-sdk
-
-# Check out genai_perf command inside the container:
-genai-perf --help
+pip install git+https://github.com/triton-inference-server/perf_analyzer.git#subdirectory=genai-perf
```
matthewkotila, line 80: I think we should be telling people to use the pypi.org option since that has existed since mid november

```bash
pip install genai-perf
```

https://pypi.org/project/genai-perf/

debermudez: oh definitely. ill update

debermudez: done

genai-perf/README.md
```diff
@@ -73,45 +73,34 @@ INSTALLATION
 
 ## Installation
 
-The easiest way to install GenAI-Perf is through
-[Triton Server SDK container](https://ngc.nvidia.com/catalog/containers/nvidia:tritonserver).
-Install the latest release using the following command:
+The easiest way to install GenAI-Perf is through pip.
+### Install Perf Analyzer (Ubuntu 24.04, Python 3.10+)
```
dyastremsky, line 77: I find some of this confusing. Do we need to talk about installing PA in the GAP instructions? If so, we should have sections at the same header level (e.g. H3 `###`) for each step in the install process. I think we could remove it though.

genai-perf/README.md
```diff
@@ -73,45 +73,34 @@ INSTALLATION
...
-
-You can also build Perf Analyzer [from source](../docs/install.md#build-from-source) as well.
+export RELEASE="24.10"
 
-### Install GenAI-Perf from source
+docker run -it --net=host --gpus=all  nvcr.io/nvidia/tritonserver:${RELEASE}-py3-sdk
 
-```bash
-pip install git+https://github.com/triton-inference-server/perf_analyzer.git#subdirectory=genai-perf
+# Check out genai_perf command inside the container:
```
dyastremsky, line 98:
```suggestion
# Validate the genai-perf command works inside the container:
```

genai-perf/README.md
```diff
@@ -73,45 +73,34 @@ INSTALLATION
...
-Since GenAI-Perf depends on Perf Analyzer,
-you'll need to install the Perf Analyzer binary:
+<summary>Alternatively, to install the container:</summary>
 
-### Install Perf Analyzer (Ubuntu, Python 3.10+)
+[Triton Server SDK container](https://ngc.nvidia.com/catalog/containers/nvidia:tritonserver)
 
-**NOTE**: you must already have CUDA 12 installed
-(checkout the [CUDA installation guide](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html)).
+Install the latest release using the following command:
```
dyastremsky, line 91: I think install may not be the right word when describing pulling a container. Maybe "pull" or "access"?

genai-perf/README.md
```diff
@@ -73,45 +73,34 @@ INSTALLATION
...
-### Install GenAI-Perf from source
+docker run -it --net=host --gpus=all  nvcr.io/nvidia/tritonserver:${RELEASE}-py3-sdk
 
-```bash
-pip install git+https://github.com/triton-inference-server/perf_analyzer.git#subdirectory=genai-perf
+# Check out genai_perf command inside the container:
+genai-perf --help
 ```
 
+You can also build Perf Analyzer [from source](../docs/install.md#build-from-source) to use alongside GenAI-Perf as well.
```
dyastremsky, line 102: This may be beyond the scope of this PR, but we should make sure we are clear when referring to PA. We could have a separate section describing it and describing why you might want to rebuild it. The casual user shouldn't need to know what PA is and how it works within GAP, so reading statements about it without context in the documentation will be more confusing than helpful IMO.

dyastremsky (commented):
I appreciate you making these changes! Improving the documentation, especially around install, is a great idea.

Do you also need to update the README template? I think we're also past 24.10 now.

genai-perf/README.md
debermudez: That is a good point but out of scope for this since we are trying to get this to a customer today.
In general, we need a pass on all our docs to make them cleaner.

genai-perf/README.md
debermudez: It should be GenAI-Perf and not PA here. Thank you.

genai-perf/README.md
debermudez: yup. makes sense.

genai-perf/README.md
debermudez: I like pull. Nice call out

dyastremsky (commented):
This looks good to me. Does the README template also need to be updated before approving/merging?

debermudez:
> This looks good to me. Does the README template also need to be updated before approving/merging?

I did a pass on the templates.
Thanks for calling this out.

dyastremsky (approved):

PR #232, opened 2024-12-25
Fix typing visibility for ExporterConfig

dyastremsky:
ExporterConfig was previously functioning as a dataclass without being one. In its init, all the types were set to None, even though the functions calling the config expected them to actually be set to a type. In other words, the type hint listed None but it was never actually supposed to be None. This resulted in Pylance errors everywhere we used this class's data objects and unclear visibility into the types of each object. This slowed development and was likely to lead to failures, albeit it made testing a bit easier (only the necessary data had to be set).

This pull request changes ExporterConfig to be a datatype and cleans up the type hints. It also adds a test util function to create a sample exporter config to keep the tests concise, create a single source of truth for changing the ExporterConfig in unit testing. That should make the unit testing more flexible and faster to update if this class changes in the future.

### DIFF
--- a/genai-perf/genai_perf/metrics/telemetry_statistics.py
+++ b/genai-perf/genai_perf/metrics/telemetry_statistics.py
@@ -48,11 +48,18 @@ class TelemetryStatistics:
         for attr, data in self._metrics.data.items():
             if self._should_skip(data):
                 continue
+
+            gpu_index = None
+            gpu_data = None
+
             for gpu_index, gpu_data in data.items():
                 self._stats_dict[attr][gpu_index]["avg"] = (
                     self._statistics._calculate_mean(gpu_data)
                 )
             if not self._is_constant_metric(attr):
+                if gpu_data is None or gpu_index is None:
+                    continue
+
                 percentile_results = self._statistics._calculate_percentiles(gpu_data)
                 for percentile_label, percentile_value in percentile_results.items():
                     self._stats_dict[attr][gpu_index][
@@ -121,5 +128,5 @@ class TelemetryStatistics:
         return attr in ["gpu_power_limit", "total_gpu_memory"]
 
     @property
-    def stats_dict(self) -> Dict:
+    def stats_dict(self) -> Dict[str, Any]:
         return self._stats_dict
--- a/genai-perf/tests/test_exporters/test_console_exporter.py
+++ b/genai-perf/tests/test_exporters/test_console_exporter.py
@@ -39,6 +39,7 @@ from genai_perf.metrics import (
     TelemetryMetrics,
     TelemetryStatistics,
 )
+from tests.test_utils import create_default_exporter_config
 
 
 class TestConsoleExporter:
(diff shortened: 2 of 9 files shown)


genai-perf/tests/test_exporters/test_console_exporter.py
```diff
@@ -124,10 +124,10 @@ def test_nonstreaming_llm_output(self, monkeypatch, capsys) -> None:
         )
         stats = Statistics(metrics=metrics)
 
-        config = ExporterConfig()
-        config.stats = stats.stats_dict
-        config.metrics = stats.metrics
-        config.args = args
+        assert isinstance(stats.metrics, Metrics)
```
dyastremsky, line 127: These asserts are necessary because the current architecture means that stats.metrics can be Metrics or TelemetryMetrics. It is beyond the scope of this PR to separate the two, but we do need asserts like this to check the correct type is used and to not result in type checking failures. It was previously opaque that we were using two types in the exporter config. We could always tackle it in a future cleanup effort if it causes issues, but making the typing transparent should help.

genai-perf/genai_perf/metrics/telemetry_statistics.py
```diff
@@ -48,11 +48,18 @@ def __init__(self, metrics: TelemetryMetrics):
...
+
+            gpu_index = None
+            gpu_data = None
+
             for gpu_index, gpu_data in data.items():
                 self._stats_dict[attr][gpu_index]["avg"] = (
                     self._statistics._calculate_mean(gpu_data)
                 )
             if not self._is_constant_metric(attr):
+                if gpu_data is None or gpu_index is None:
```
matthewkotila, line 60: Am I reading this correctly? When would gpu_data or gpu_index _not_ be `None`?

They're set to `None` at the start of this code block comment. My understanding is that the for loop variables of the same name are separate and shadow the initial ones, and disappear after the for loop.

dyastremsky: I have learned something new today! I thought Python kept the variable values; I hadn't heard of shadowing.

I'll take a look at a different way to do this. I think PyLance reported a potential static error here (e.g. a NoneType). I'll revert this and look for a better way to do this. Thanks, Matt! 🙇‍♂️

genai-perf/tests/test_utils.py
```diff
@@ -60,6 +64,27 @@ def check_statistics(s1: Statistics, s2: Statistics) -> None:
...
+    extra_inputs: Optional[Dict[str, Any]] = None,
+    artifact_dir: Optional[Path] = None,
+    telemetry_stats: Optional[Dict[str, Any]] = None,
+) -> ExporterConfig:
+    return ExporterConfig(
+        stats=stats or {},
+        metrics=metrics or Metrics(),
+        args=args or Namespace(),
+        extra_inputs=extra_inputs or {},
+        artifact_dir=artifact_dir or Path("."),
```
matthewkotila, line 83: Neat use of `or` operator! Never seen it used in assignments.

genai-perf/genai_perf/metrics/telemetry_statistics.py
dyastremsky: Matt and I discussed this offline. After some testing, Python does keep variable values when they go out of scope for a loop. This prevents an unbounded value for gpu_data or gpu_index if data is empty (for loop never iterates).

@matthewkotila, would you like me to leave this in? Alternatively, I can remove these changes and let @lkomali know about this pontential error since she's more familiar with this code.

matthewkotila: My understanding was incorrect--as you mentioned, variables defined in the python for loop persist after the loop with the last value they retained.

That means this code is valid--resolved.

matthewkotila (approved):

PR #234, opened 2024-12-30
Improve error visibility when printing metrics

dyastremsky:
When a metric error happens, it can be hard to debug whether something went wrong with the server, perf analyzer, or GenAI-Perf (e.g. edge condition, bug). These errors have been reported for different situations, and debugging them has been opaque.

This pull request catches common errors when accessing the statistics dictionary and logs them as errors. This allows the output to still be printed (with N/A) to not lose the data already gathered/processed but also alerts the user what the specific issue is.

_Example screenshot with fake "p10" statistic added to trigger error._
image


genai-perf/README.md
```diff
@@ -508,6 +508,7 @@ when `--output-tokens-mean` is provided. (default: `0`)
 The seed used to generate random values. (default: `0`)
 
 ##### `--request-count <int>`
+##### `--num-requests <int>`
```
dyastremsky, line 511: Added aliases. While PA uses request-count, all other GAP args that use counts start with --num (most notably, --num-prompts).

We can decide for GA what arg name to use (if only one), but having this consistency would make it less confusing to be adding args. I often would type out the wrong one, so I imagine some users might too.

genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
```diff
@@ -124,7 +124,7 @@ def _construct_telemetry_stats_subtable(
                     value = str(int(round(value)))
                 else:
                     value = f"{value:,.2f}"
-            row.append(value)
+            row.append(str(value))
```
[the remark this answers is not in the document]
dyastremsky: Removed this. I think Pylance is in the wrong on this, since all the branches are strings. In any case, I chose not to figure out how to silence the pre-existing warning in this PR.

genai-perf/README.md
debermudez: This was the previous name right?
An alias is fine, just having deja vu here.

debermudez (approved):

debermudez:
Nice work on this!

dyastremsky:
>This was the previous name right?
>An alias is fine, just having deja vu here.

@debermudez Not as far as I know. I wasn't involved in the initial request count work, so it's possible that was the case. This does seem like the more consistent name on the GenAI-Perf name, though I know we try to reuse the PA flag names in GenAI-Perf.

PR #236, opened 2025-01-04
Add Prometheus metric support

dyastremsky:
Support a Prometheus metric endpoint in GenAI-Perf. This is a proof of concept. With this design, Prometheus metrics can be enabled via the CLI. If they are enabled, the metrics are exported at that endpoint. This would only be supported for the `profile` subcommand to begin with, though it may be useful to expand it to `analyze`. To avoid breaking anything, that is out of scope for this proof of concept.

When Prometheus metrics are enabled, the user must terminate the program when they want GenAI-Perf to exit. This design is necessary because otherwise, the metrics would be up for a very short amount of time, since the endpoint is closed when GenAI-Perf exits.

TODO: Only system metrics are displayed right now, so add the other metrics. Also, it may be good to omit request_goodput if goodput is not used.

Example command:
`genai-perf profile -v -m gpt2 --service-kind openai --endpoint-type completions --num-requests 5 --num-prompts 5 --enable-prometheus --prometheus-port 8002`

Example metrics:
image


genai-perf/genai_perf/main.py
```diff
@@ -26,18 +26,22 @@
...
 from genai_perf import parser
 
+logger = logging.getLogger(__name__)
+
 
 # Separate function that can raise exceptions used for testing
 # to assert correct errors and messages.
 def run():
     # TMA-1900: refactor CLI handler
-    logging.init_logging()
```
nv-braf, line 40: This should not have been deleted, it is needed for the logger to function properly

genai-perf/genai_perf/main.py
```diff
@@ -46,6 +50,31 @@ def run():
     else:  # profile
         args.func(args, extra_args)
 
+    if getattr(args, "enable_prometheus", False):
+        # Fix: This doesn't actually get logged.
```
nv-braf, line 54: You can fix this by adding main to `logger.py`

genai-perf/genai_perf/inputs/input_constants.py
```diff
@@ -70,6 +70,7 @@ def to_lowercase(self):
 DEFAULT_SYNTHETIC_FILENAME = "synthetic_data.json"
 DEFAULT_WARMUP_REQUEST_COUNT = 0
 DEFAULT_BACKEND = "tensorrtllm"
+DEFAULT_PROMETHEUS_PORT = 8002
```
nv-braf, line 73: This is the default port for the Tritonserver metrics. I would change this to be 9090 which is what prometheus uses in its getting started configuration file.

genai-perf/genai_perf/main.py
dyastremsky: Reverted, thanks for catching this.

genai-perf/genai_perf/main.py
dyastremsky: Ah, I didn't know that. Thanks, Brian! Done.

genai-perf/genai_perf/inputs/input_constants.py
dyastremsky: Great feedback. Done.

genai-perf/genai_perf/export_data/prometheus_exporter.py
```diff
@@ -0,0 +1,83 @@
...
+    def __init__(self, config: ExporterConfig):
+        self._stats = config.stats
+        self._metrics = config.metrics
+        self._args = config.args
+        if not hasattr(self._args, "prometheus_port"):
+            raise ValueError("Prometheus port is not provided.")
+        self._port = (
+            self._args.prometheus_port
+            if hasattr(self._args, "prometheus_port")
+            else 8000
```
ajcasagrande, line 50: Shouldn't this be `ic.DEFAULT_PROMETHEUS_PORT`?

genai-perf/genai_perf/export_data/prometheus_exporter.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    Exports metrics in a Prometheus-compatible format via an HTTP endpoint.
+    """
+
+    def __init__(self, config: ExporterConfig):
+        self._stats = config.stats
+        self._metrics = config.metrics
+        self._args = config.args
+        if not hasattr(self._args, "prometheus_port"):
+            raise ValueError("Prometheus port is not provided.")
```
ajcasagrande, line 46: This check is not necessary, as the lines below (`47-51`) will assign `self._port` to the default value if it is not found in `self._args`. The way it is written now, the else condition on lines `49-50` will never be executed because a `ValueError` will be raised.

genai-perf/genai_perf/export_data/prometheus_exporter.py
dyastremsky: Great feedback, thanks! I'll fix this shortly. :)

Just a heads up that this PR is currently stalled on whether we want to use this approach for publishing Prometheus metrics. It's being reviewed as a proof of concept.

genai-perf/genai_perf/export_data/prometheus_exporter.py
dyastremsky: Great point, will fix this. Thanks Anthony!

PR #239, opened 2025-01-09
Fix GAP OSL issue due to PA not removing first 4 bytes from output

lkomali:
PA removes first 4 bytes from output when datatype is BYTES.

When fixing L0_CApi segfault, this piece of code was moved to another file, to reduce overhead by avoiding vector resizing. Hence, PA removes first 4 bytes only when using Triton CAPI mode, which caused OSL increase.

This PR fixes the above issue by placing back the code in the original file.

### DIFF
--- a/src/client_backend/triton_c_api/triton_loader.cc
+++ b/src/client_backend/triton_c_api/triton_loader.cc
@@ -1048,10 +1048,6 @@ TritonLoader::GetOutputs(
 
     std::string data_type{datatype};
     std::vector<uint8_t> data_copy;
-    if (data_type == "BYTES" && byte_size >= 4) {
-      base = static_cast<const uint8_t*>(base) + 4;
-      byte_size -= 4;
-    }
 
     if (memory_type == TRITONSERVER_MEMORY_GPU) {
       CUDARuntimeLibraryManager cuda_manager;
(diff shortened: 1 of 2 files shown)


matthewkotila (approved):

debermudez (approved):

dyastremsky (approved):

rmccorm4:
CC @kthui

rmccorm4:
Just curious - is there plans for a simple test with a dummy model that returns an output with a known number of expected tokens, and verification that the token count is correct? I think that would automate catching this class of bug.

src/client_backend/triton_c_api/triton_loader.cc
rmccorm4, line 1: Nice work Harshini!

Just curious - can we have a simple test with a dummy model that returns an output with a known number of expected tokens, and verification that the token count is correct? I think that would automate catching this class of bug.

dyastremsky: I drafted a CI test to add this verification. Still working on it.

If a unit test is possible as well, that'd be ideal.

lkomali: I don't know on top of my head the effort for a unit test but it's definitely worth looking into. 
I will create a follow-up ticket to investigate about the unit test and add it.

PR #240, opened 2025-01-11
Fix OSL numbers for streaming case

dyastremsky:
Instead of concatenating before joining the response texts, encode the response texts as they are received. Then, sum the encodings. That will make GenAI-Perf calculate output tokens the same way the server is doing so.

Before:
image

After:
image

_Confirmed with GPT2 and Meta-Llama-3-8B-Instruct_

### DIFF
--- a/genai-perf/tests/test_data_parser/test_llm_profile_data_parser.py
+++ b/genai-perf/tests/test_data_parser/test_llm_profile_data_parser.py
@@ -135,13 +135,13 @@ class TestLLMProfileDataParser:
                     "request_throughputs": [1 / ns_to_sec(5)],
                     "time_to_first_tokens": [2, 2],
                     "time_to_second_tokens": [2, 3],
-                    "inter_token_latencies": [2, 1],
+                    "inter_token_latencies": [1, 1],
                     "output_token_throughputs_per_request": [
-                        3 / ns_to_sec(7),
-                        2 / ns_to_sec(3),
+                        5 / ns_to_sec(7),
+                        8 / ns_to_sec(9),
                     ],
-                    "output_token_throughputs": [9 / ns_to_sec(10)],
-                    "output_sequence_lengths": [3, 6],
+                    "output_token_throughputs": [13 / ns_to_sec(10)],
+                    "output_sequence_lengths": [5, 8],
                     "input_sequence_lengths": [3, 4],
                 },
             ),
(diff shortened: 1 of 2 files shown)


genai-perf/tests/test_data_parser/test_llm_profile_data_parser.py
```diff
@@ -841,7 +841,7 @@ def test_openai_output_token_counts(self, mock_json) -> None:
             res_outputs
         )
         assert output_token_counts == [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]  # total 10
-        assert total_output_token == 9
+        assert total_output_token == 19
```
dyastremsky, line 844: @nv-hwoo I want to call out that the old way of tokenizing has a big difference here. This seems specific to the unit test though. This is likely because a lot of the responses aren't actually a typical response with a set token: they have a space in them (which is usually its own token, so the response gets tokenized as two tokens). My initial guess is that the llama tokenizer trick we use somehow gets rid of the space.

This won't matter for OSL statistics, but I wanted to mention this.

nv-hwoo: It seems like this is due to the tokenizer we are using in the test. The test works as expected when I use `gpt2` tokenizer (`total_output_token` equals 10). I think this may be because the the texts in `output_texts` is generated by `gpt2` tokenizer but the test is using the default tokenizer.

One workaround I can think of is to use the texts/tokens generated by the given tokenizer to avoid any mismatch:
```python
text = "what is deep learning?"
tokenizer = get_tokenizer(DEFAULT_TOKENIZER)
output_texts = [tokenizer.decode(txt) for txt in tokenizer.encode(text)]
# ... get token counts ...
assert output_token_counts == [1, 1, 1, 1, 1]
assert total_output_token == 5
```

But I feel like we need to do some e2e tests to properly test the token counts.

dyastremsky: Yeah, I'm adding e2e tests. I didn't change the tokenizer - it's interesting that the change affected that. Thanks for identifying the above.

I will say that if we need to update these tests often, it'd be good to think if there is another way to update these unit tests. It took me 2-3 hours to understand and properly update these three unit tests because there are a lot of calculations that have to be understood, performed, and manually updated in both the comments and the code.

I can update the unit tests if we think it's worthwhile to do so to avoid the above issue in the future, but it might be worth saving our effort and applying it elsewhere, since we already know what token counts the unit tests should get.

nv-hwoo: Yeah I think we can leave it as is for now. There are a few factors that contribute to the complexity of the above unit tests and it all boils down to the fact that the data parser class is very complex and needs some well-thought-out refactoring.

nv-hwoo (approved):

dyastremsky:
It seems like this solution is more exact for the completions endpoint, but it results in much worse counts for the chat endpoint. I'll close the pull request for now.

PR #241, opened 2025-01-13
Basic Config File Parsing

nv-braf:


genai-perf/genai_perf/config/generate/search_parameters.py
```diff
@@ -64,21 +47,20 @@ def __init__(
...
-            self._config = config.optimize
-        elif subcommand == Subcommand.ANALYZE:
-            self._config = config.analyze  # type: ignore
 
-        # TODO: OPTIMIZE
-        # self._supports_model_batch_size = model.supports_batching()
+        if subcommand == Subcommand.ANALYZE:
+            self._config = config.analyze
+        else:
+            raise GenAIPerfException(f"Subcommand {subcommand} not supported")
```
debermudez, line 54: I am assuming this will never be hit in practice?
This is just here to guard against it and throw a warning for us to debug. Is that correct?

genai-perf/tests/test_config_command.py
```diff
@@ -0,0 +1,272 @@
...
+from unittest.mock import patch
+
+# Skip type checking to avoid mypy error
+# Issue: https://github.com/python/mypy/issues/10632
+import yaml  # type: ignore
+from genai_perf.config.input.config_command import ConfigCommand, Range
+from genai_perf.inputs.input_constants import ModelSelectionStrategy, OutputFormat
+
+
+class TestConfigCommand(unittest.TestCase):
```
debermudez, line 25: I have a concern about using unittest here. Pytest and unittest are compatible but the structure of the testing has different patterns now.
Was unittest used here for a specific functionality that is harder in pytest? OR is it what you are more familiar with?

genai-perf/tests/test_config_command.py
```diff
@@ -0,0 +1,272 @@
...
+        """
+        Test that the perf analyzer configuration is parsed correctly
+        """
+        # yapf: disable
+        yaml_str = ("""
+            model_name: gpt2
+
+            perf_analyzer:
+                path: test_path
+                stimulus:
```
debermudez, line 131: what does stimulus translate to in PA options?

genai-perf/tests/test_optuna_objective_generator.py
```diff
@@ -1,227 +0,0 @@
-# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
debermudez, line 1: why was this deleted?

genai-perf/genai_perf/config/input/config_command.py
```diff
@@ -241,3 +376,305 @@ def get_min(self, config_value: ConfigRangeOrList) -> int:
...
+            elif key == "service_kind":
+                self.endpoint.service_kind = value
+            elif key == "streaming":
+                self.endpoint.streaming = value
+            elif key == "server_metrics_url":
+                self.endpoint.server_metrics_url = value
+            elif key == "url":
+                self.endpoint.url = value
+            else:
+                raise ValueError(
```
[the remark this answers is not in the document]
nv-braf: Good idea! I've created a ticket for this.

genai-perf/tests/test_optuna_objective_generator.py
nv-braf: It would only be used by optimize and it relies heavily on config options that do not exist. If/When we decide to implement optimize we can pull these files back in to the code base.

genai-perf/tests/test_config_command.py
nv-braf: Good catch! This is either concurrency or request rate (replaces --concurrency/request-rate). I've fixed the default (should be concurrency: 1) and updated the unit testing to make this more clear.

genai-perf/tests/test_config_command.py
nv-braf: I just don't know pytest and unittest is fully compatible in a pytest environment. I'm not doing anything fancy here, so there's no reason to use unit test other than it's what I'm familiar with.

FYI, all the code w/ analyze I added is done in unittest so unfortunately we already have a mix of testing in GAP.

genai-perf/genai_perf/inputs/input_constants.py
```diff
@@ -29,33 +29,33 @@
...
+    FILE = "FILE"
 
 
 class OutputFormat(Enum):
     ################################################################
     # Triton backends
     ################################################################
-    TENSORRTLLM = auto()
-    VLLM = auto()
+    TENSORRTLLM = "TENSORRTLLM"
```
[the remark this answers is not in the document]
nv-braf: It's so you can assign to the enum with string like this:
`self.endpoint.backend = OutputFormat(value.upper())`

otherwise I'd need to have a switch statement which seemed more tedious.

genai-perf/genai_perf/config/generate/search_parameters.py
nv-braf: Yes, this can never be hit. I suppose this could also be an assert to ensure that the subcommand is analyze. What's your preference?

genai-perf/genai_perf/config/input/config_command.py
```diff
@@ -241,3 +376,305 @@ def get_min(self, config_value: ConfigRangeOrList) -> int:
...
+        elif sweep_type == "request_rate":
+            return AnalyzeDefaults.MAX_REQUEST_RATE
+        elif sweep_type == "num_dataset_entries":
+            return AnalyzeDefaults.MAX_NUM_DATASET_ENTRIES
+        elif sweep_type == "input_sequence_length":
+            return AnalyzeDefaults.MAX_INPUT_SEQUENCE_LENGTH
+        else:
+            raise ValueError(f"User Config: {sweep_type} is not a valid sweep type")
+
+    def _get_default_step(self, sweep_type: str) -> int:
```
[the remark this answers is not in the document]
nv-braf: It should be. I've added this as a default.  Good catch.

genai-perf/tests/test_config_command.py
debermudez: okay. perhaps not ideal but its the most extensive testing we have in place so im not gonna complain about it. Thanks for chatting thru it with me.

genai-perf/genai_perf/config/generate/search_parameters.py
debermudez: This is fine. They are functionally equal

debermudez (approved):

PR #243, opened 2025-01-16
Starting to work on coroutines.

nicolasnoble:
This introduces a Coroutine class in the pa namespace, which we will then be able to use for a lightweight scheduler.


src/CMakeLists.txt
```diff
@@ -361,6 +363,7 @@ target_compile_options(
   perf_analyzer_unit_tests
   PRIVATE
     -Wno-write-strings
+    -fcoroutines
```
matthewkotila, line 366: Why is this necessary? I thought using `--std=c++20` would be enough

nicolasnoble: Ha ha ha nope. For some reason, the C++ steering committee decided to put this behind a gate. Both are needed.

matthewkotila: A gotcha. Well let's not forget to add -fvector -fstring -vunordered_map -fmemory -futility

Oh wait, what's that? Those aren't needed? There's different patterns used within the same language?

/s

src/coroutine.h
```diff
@@ -0,0 +1,265 @@
...
+    void unhandled_exception() {}
+    void return_void()
+    {
+      auto awaitingCoroutine = awaitingCoroutine_;
+      if (awaitingCoroutine) {
+        awaitingCoroutine_ = nullptr;
+        std::coroutine_handle<>::from_address(awaitingCoroutine).resume();
+      }
+    }
+    [[no_unique_address]] std::monostate value_{};
```
matthewkotila, line 90: Is this necessary for this?:

https://github.com/triton-inference-server/perf_analyzer/pull/243/files#diff-053471f1e45d31573787f7b9b3b13d623b2d2cbff9026b8e8d43bf3d4c95f2fdR122-R124

In other words if the structs didn't have the same list of data member identifiers, the typedef line wouldn't work?

nicolasnoble: It's for the await_resume at the end, and the various other generic places where we haul the result of the promise.

src/test_coroutines.cc
```diff
@@ -0,0 +1,136 @@
+// Copyright 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
matthewkotila, line 1: nit: rename `test_coroutine.cc`

src/coroutine.h
matthewkotila: Why not just one promise_type struct that has both return_void and return_value?

src/test_coroutines.cc
matthewkotila, line 1: Can you give examples in the context of PA's current objectives?

One of the main features implemented in this PR is manual control over coroutines (which kinda seems counterintuitive at first). Can you give an example with inference load generation (concurrency, request rate, etc) that would rely on being able to perform Resume(), Done(), etc manually on the coroutine?

matthewkotila, line 1: Another main feature I've noticed is support for returning both void and non-void.

Can you give examples within the context of inference load generation where both void and non-void returns aid the main objectives of PA?

src/coroutine.h
```diff
@@ -0,0 +1,265 @@
...
+  };
+
+  /**
+   * @brief Resumes the coroutine.
+   *
+   * @details This method resumes the coroutine if it is suspended using the
+   * Awaiter mechanism above, or for its initial execution. If the coroutine is
+   * not suspended, it sets a flag to resume it early when it is next suspended.
+   */
+  void Resume()
```
matthewkotila, line 186: Should this function be thread-safe?

src/test_coroutines.cc
```diff
@@ -0,0 +1,136 @@
...
+  auto coroutine = CoroutineVoidTest();
+
+  unsigned rounds = 0;
+  while (!coroutine.Done()) {
+    coroutine.Resume();
+    rounds++;
+  }
+
+  CHECK(rounds == 2);
+  CHECK(coroutine.Done());
```
matthewkotila, line 83: Should you check that value is std::monostate here? Just to validate that part of the implementation.

src/test_coroutines.cc
```diff
@@ -0,0 +1,136 @@
...
+
+TEST_CASE("coroutine:testing the Coroutine class with void")
+{
+  auto coroutine = CoroutineVoidTest();
+
+  unsigned rounds = 0;
+  while (!coroutine.Done()) {
+    coroutine.Resume();
+    rounds++;
+  }
```
matthewkotila, line 80: Took me a few minutes of reading code to understand why rounds was 2. I think a test like this would have been quicker to gleam the understanding of while still validating the same behavior.

```suggestion
REQUIRE(!coroutine.Done());

coroutine.Resume(); // because coroutine is initially suspended
coroutine.Resume(); // because coroutine suspends at end

REQUIRE(coroutine.Done());
```

src/coroutine.h
nicolasnoble: So this is a bit on the C++ design here. There's a sort of flaw: when the language encounters `co_return`, it looks up the contents of the promise object associated with the coroutine, and makes a decision based on the presence of either `return_value` and `return_void`. To solve this connundrum, we create two promise types, and then we erase one of the two using `std::conditional` to pick the proper one. See https://en.cppreference.com/w/cpp/language/coroutines for some more details on the execution flow.

Basically, having both creates this ambiguity you shouldn't have in normal C++:
```c++
int someFunction() {
    if (someErrorCondition) {
        return;
    }
    return 42;
}
```

src/coroutine.h
```diff
@@ -0,0 +1,265 @@
...
+    constexpr bool await_suspend(std::coroutine_handle<U> h)
+    {
+      auto& promise = h.promise();
+      bool ret = promise.earlyResume_;
+      promise.earlyResume_ = false;
+      if (!ret) {
+        promise.suspended_ = true;
+      }
+      return !ret;
+    }
```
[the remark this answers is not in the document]
nicolasnoble: Right, this Awaiter class here is a generic helper, and as such, it can't be hoisted inside the actual logic of what's being awaited on. A more specialized awaiter would use whichever business logic appropriate, while hoisting itself into the actual logic of the asynchronous code.

For example:
```c++
Coroutine<> someCoroutine {
    Coroutine<>::Awaiter awaiter;
    someAsynchronousFunction([coro]{ coro.Resume(); })
    co_await awaiter;
}
```

In this case, we can't really know if `someAsynchronousFunction` terminates its work immediately before returning, which can happen in some API. And so without some state management, this throw off the state of the coroutine, with `co_await` here in this case waiting forever.

Now, as the documentation I've added on the class says, this is a generic helper which should only be used as a last resort. It's best to have specialized awaiters which will know the actual state of the asynchronous operation happening, and using await_ready returning the proper value upfront. But both await_ready and await_suspend have the same power in terms of resuming the coroutine immediately, just one has extra knowledge compared to the other. Not exactly sure why the C++ design committee chose to create two instead of just one. It feels futile.

src/coroutine.h
nicolasnoble: Short answer: no.

Similar to the rest of the C++ language, low level building blocks shouldn't be threadsafe, unless their purpose is to actually be threads building blocks. And in the case of co-routines, as they are supposed to _replace_ threads, their purpose is to avoid threads altogether. The scheduler which is what will be responsible for resuming a coroutine will be either single-threaded, or thread-safe itself. It shouldn't be the responsability of the low level building block to make that decision, as it'd be a cost that's paid regardless of the intention: single-threaded or not.

src/test_coroutines.cc
nicolasnoble: Ack.

nicolasnoble: So, first and foremost: only the scheduler is supposed to handle these calls like done or resume. I'm planning to write a more fleshed out scheduler next, but I wanted to avoid doing so in a single PR, to keep things in separate building blocks (a scheduler implementation isn't necessarily tied to a specific coroutine implementation, and a coroutine implementation isn't tied to a specific coroutine implementation).

In terms of goal, we absolutely want to reduce our thread count. A thread is extremely expensive, both CPU and memory wise. It implies delegating to the kernel all things scheduling, while we actually have some sort of scheduler implace already, which we'd replace with a coroutine-based one here. Using many threads also implies allocating a thread every time, which is a giant, several megabytes blob. And it finally implies having proper locks and atomics usage, which comes with their own complexity and CPU contention.

By rewriting the scheduler we currently have to leverage coroutines instead of threads, we can coalese a bunch of our threads in a single one. And this will also ultimately simplify the business logic writing of scenarios, which wouldn't need to care much about thread safety. I know it's a bit of a "trust me" situation, but that's the best I can offer without writing everything to demonstrate what I'm talking about :)

nicolasnoble: Well, this is the same as with any normal business logic writing. At the end of the day, a coroutine is just a normal function, just running asynchronously. You want to be able to write your business logic using void and non-void coroutines the same way you want to be able to write your business logic using void and non-void functions.

```c++
Coroutine<> someTopLevelCoroutine(int someArgument) {
    co_await someOtherCoroutine(someArgument);
    int x = co_await someCoroutineReturningSomething();
    co_await someFinalCoroutine(someArgument * x);
}
```

In the case of PA, a scenario would most likely be a void coroutine, since I can't really think of why it'd need to return anything when it's done running, but it still should be split into more manageable business logic building blocks, which may have to return values.

src/test_coroutines.cc
nicolasnoble: I don't mind changing this, though it takes away a bit of the original goal of showcasing how a proto-scheduler works. Basically, the point of a scheduler is to run a coroutine until it's done, but there's a more complex example below anyway, so, sure.

src/test_coroutines.cc
nicolasnoble: `std::monostate` is a type, not a value we can test against at runtime, but I can add a compile-time check here, sure.

src/test_coroutines.cc
matthewkotila: Sure feel free to keep it as a loop, just offering ideas where there's agreement

src/coroutine.h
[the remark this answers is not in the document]
matthewkotila: Didn't quite understand but thanks for making the explanation.

matthewkotila (dismissed):
I don't comfortably understand coroutines yet but let's just see how this can be used.

nicolasnoble:
I'm trying hard not to go "trust me bro" here, but we'll see with the next few changes as I'll work on the new scheduler next :)

PR #244, opened 2025-01-16
Update readme and version

mc-nv:
Related changes: 
- https://github.com/triton-inference-server/triton_cli/pull/99
- https://github.com/triton-inference-server/server/pull/7936


templates/template_vars.yaml
```diff
@@ -1,7 +1,7 @@
 General:
-  release: 24.12
-  triton_cli_version: 0.0.11
-  genai_perf_version: 0.0.10dev
+  release: 25.01
+  triton_cli_version: 0.1.4
```
debermudez, line 3: this is not the usual single increment.
Assuming this was intentional, this is good to go. Just wanted to call it out.

debermudez (approved):

templates/template_vars.yaml
rmccorm4: Triton CLI doesn't have a bump to 0.1.4 available yet / isn't ready to upgrade. You could try bumping to 0.1.1 (https://github.com/triton-inference-server/triton_cli) if tested to make sure it works with genai-perf, but I think @dyastremsky was running into some issue recently.

rmccorm4: But I think help from genai-perf folks would be needed to test or reject the upgrade before merge - I don't think @mc-nv should need to

mc-nv: This is a current version and we can apply it if needed.
```suggestion
  triton_cli_version: 0.1.1
```

dyastremsky: I think it's okay to bump. I was having issues with running a model with Triton CLI and need to try doing so again.

If there's an issue, it would have been caught by the Triton CLI CI, so it's likely user error. :)

mc-nv: I reverted that change and will leave it to the team.

dyastremsky (approved):

dyastremsky:
Approved. Was reverting all of the template vars correct? https://github.com/triton-inference-server/perf_analyzer/pull/244/commits/07d8c6f679dae96443b62eaf64b482bd583b81b3

This will still be 24.12 and 0.0.10dev then.

PR #246, opened 2025-01-17
Improved developer experience somewhat.

nicolasnoble:
- Added a devcontainer.json.
- Added a launch.json for the unit tests.
- Cleaned up .gitignore file.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,7 +1,7 @@
 {
   "name": "perf_analyzer Dev Container",
   "image": "gitlab-master.nvidia.com:5005/dl/dgx/tritonserver:master-py3-base",
-  "postCreateCommand": "apt update && apt install -y --no-install-recommends clang-format-16 cmake gdb rapidjson-dev libssl-dev",
+  "postCreateCommand": "apt update && apt install -y --no-install-recommends clang-format-16 cmake gdb rapidjson-dev libssl-dev && update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-16 100",
   "runArgs": ["--gpus=all", "--net=host", "--pull=always"],
   "customizations": {
     "vscode": {
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,6 +1,6 @@
 {
   "name": "perf_analyzer Dev Container",
-  "image": "gitlab-master.nvidia.com:5005/dl/dgx/tritonserver:master-py3-devel",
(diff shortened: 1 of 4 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "perf_analyzer Dev Container",
+  "image": "gitlab-master.nvidia.com:5005/dl/dgx/tritonserver:master-py3-devel",
```
matthewkotila, line 3:
```suggestion
  "image": "gitlab-master.nvidia.com:5005/dl/dgx/tritonserver:master-py3-base",
```

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,11 @@
...
+  "postCreateCommand": "apt update && apt install -y --no-install-recommends clang-format-15 cmake gdb libb64-dev rapidjson-dev",
```
matthewkotila, line 4:
```suggestion
  "postCreateCommand": "apt update && apt install -y --no-install-recommends clang-format-16 cmake gdb rapidjson-dev",
```

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,11 @@
...
+  "runArgs": ["--gpus=all", "--net=host"],
```
matthewkotila, line 5:
```suggestion
  "runArgs": ["--gpus=all", "--net=host", "--pull=always"],
```

.devcontainer/devcontainer.json
nv-hwoo: When did we start using the base container for PA build? I was using devel until now 😅

nicolasnoble: Ask @matthewkotila - originally I had the -devel too :D

matthewkotila: You might have different needs than me, but I use base because it still has triton and is smaller than devel (faster to pull).

nv-hwoo: No particular needs. I was mainly using it because our onboarding doc uses it. Good to know since I would like to keep things more minimal 👍

matthewkotila (dismissed):

.gitignore
```diff
@@ -1,17 +1,5 @@
...
-/.vscode
-
 #C++
 /build
-/builddir
-*.so
-
-#Python
-__pycache__/
-*.pyc
```
nv-hwoo, line 11: What is he reason behind not ignoring these python bytecodes? Is it for faster startup time?

nicolasnoble: Oh I meant to move that to the genai directory which has its own gitignore file.

matthewkotila (approved):

PR #247, opened 2025-01-19
Creating new boilerplate to load dynamic libraries.

nicolasnoble:
This boilerplate is intending to replace the shared_library.{h,cc} code, and provide a better mechanism to load shared libraries while adding encapsulation and type safety.


matthewkotila:
Closing due to changing priorities.

PR #249, opened 2025-01-22
Add ConfigField Library

nv-braf:
Adds a new set of classes ConfigField and BaseConfig.
ConfigField stores information information about the field and does bounds/choice checking.
BaseConfig stores a dict of ConfigField along with the necessary dunder methods to allow clean read/write of the fields.

I've updated ConfigCommand to use this new class and have filled in the default values and bounds/choices (templating will come in a later story). And, I've updated the unit testing in ConfigCommand to reflect these changes.

I've also updated the PA and GAP config and SearchParameter classes to reflect these changes. 
All unit tests are passing and live testing will begin along with changes to the top-level/parser in a follow-on story.


genai-perf/tests/test_config_command.py
```diff
@@ -47,6 +49,24 @@ def test_model_name(self):
...
+        self.assertEqual(config.get_field("model_names").required, True)
+
+    def test_multi_model_name(self):
+        """
+        Test that a configuration with multiple model names is parsed correctly
+        """
+        # yapf: disable
+        yaml_str = ("""
+            model_names:
+                - gpt2
```
debermudez, line 61: if your system has 2 instance of a model, how would we handle that here? do we care? do we have to list the model with 2 instances twice?

genai-perf/tests/test_config_command.py
```diff
@@ -267,6 +284,45 @@ def test_tokenizer(self):
         self.assertEqual(config.tokenizer.revision, "test_revision")
         self.assertEqual(config.tokenizer.trust_remote_code, True)
 
+    def test_deepcopy_config(self):
+        """
+        Test that the configuration can be deepcopied
+        """
+        # yapf: disable
+        yaml_str = ("""
+            model_name: gpt2
```
debermudez, line 293: why do you list model names as `model_name: gpt2` here but above in `test_multi_model_name` you have a list with each item on a separate line? Do they both work the same way?

genai-perf/genai_perf/constants.py
```diff
@@ -47,14 +47,9 @@
...
-    "instance_count",
-    "max_queue_delay",
-]
 
 runtime_pa_parameters = ["runtime_batch_size", "concurrency", "request_rate"]
 
 runtime_gap_parameters = ["num_dataset_entries", "input_sequence_length"]
 
-all_parameters = model_parameters + runtime_pa_parameters + runtime_gap_parameters
+all_parameters = runtime_pa_parameters + runtime_gap_parameters
```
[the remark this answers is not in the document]
nv-braf: model parameters are only a part of optimize. This code should have been removed in the previous story

genai-perf/tests/test_config_command.py
nv-braf: Number of instances of a model is a "model parameter" and shouldn't be relevant here. I don't think we care in GAP/PA

debermudez (approved):

PR #250, opened 2025-01-22
Trying to prevent people's local launch.json from being squashed.

nicolasnoble:


.vscode/settings.json
matthewkotila, line 1: This will overwrite people's local settings, too 😂😅

nicolasnoble: Right, something that multi-root workspaces addresses too.

matthewkotila (approved):

PR #251, opened 2025-01-23
Don't change the casing of model parameters.

nicolasnoble:
cc @debermudez 

I have very low confidence we won't break anything with this change. @nv-hwoo do you remember why this feature from 2 years ago needed to have lower case parameters?

### DIFF
--- a/src/command_line_parser.cc
+++ b/src/command_line_parser.cc
@@ -1648,7 +1648,6 @@ CLParser::ParseCommandLine(int argc, char** argv)
                 "<name:value:type>.");
           }
 
-          std::for_each(values.begin(), values.end(), ToLowerCase);
           std::string name{values[0]};
           std::string value{values[1]};
           std::string type{values[2]};
--- a/src/command_line_parser.cc
+++ b/src/command_line_parser.cc
@@ -1651,6 +1651,7 @@ CLParser::ParseCommandLine(int argc, char** argv)
           std::string name{values[0]};
           std::string value{values[1]};
           std::string type{values[2]};
+          ToLowerCase(type);  // to string-match the type
 
           cb::RequestParameter param;
           param.name = name;


src/command_line_parser.cc
```diff
@@ -1648,7 +1648,6 @@ CLParser::ParseCommandLine(int argc, char** argv)
                 "<name:value:type>.");
           }
 
-          std::for_each(values.begin(), values.end(), ToLowerCase);
           std::string name{values[0]};
           std::string value{values[1]};
           std::string type{values[2]};
```
nv-hwoo, line 1653: It seems like we used to string match the `type` value but that part has been [removed](https://github.com/triton-inference-server/client/pull/405/files#diff-ccd651e544af02068cd4f24962c8cb245ad5e0c3285993594a8e7d94cdbec6eeL1604) from the command_line_parser.cc and moved to other places in [client repo](https://github.com/triton-inference-server/client/blob/3d9424956ba14660f9447575928601d449919abd/src/c%2B%2B/library/grpc_client.cc#L1455) so I think the lowercasing is still needed. But I think we only need to lowercase the `type` value (not sure why I was applying it to other ones):

```suggestion
          std::string type{values[2]};
          ToLowerCase(type);  // to string-match the type
```

debermudez:
This is the type of change that scares me. 
So many assumptions can come from the precondition that we lowercase before checking.
If it passes CI, we can roll with it but lets keep an eye on this in case others come up with issues.

src/command_line_parser.cc
nv-hwoo: After giving it some more thoughts, I think another main issue here is that the place where the lowercasing is actually being used is now very far away. So I'm leaning into just deleting the lowercase entirely here and adding it to the client repo just before the string matching. @nicolasnoble any thoughts?

nv-hwoo: Not sure if I understood your point. Could you clarify the suggestion/point?

debermudez: I am just saying that it is wise to move the modification of the data as close to where you need the data changed.
Basically, i agree with you.

nv-hwoo (approved):

PR #252, opened 2025-01-23
Telemetry support for multiple metric endpoints

dyastremsky:
Add support for users to pass multiple metrics endpoints. Aggregate the metrics and provide them with the same behavior as a single endpoint.

This will be valuable for cases where more distributed serving is used.

Example:
image

### DIFF
--- a/genai-perf/README.md
+++ b/genai-perf/README.md
@@ -1,5 +1,5 @@
 <!--
-Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+Copyright (c) 2024-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions
--- a/genai-perf/README.md
+++ b/genai-perf/README.md
@@ -415,6 +415,13 @@ Set a custom endpoint that differs from the OpenAI defaults. (default: `None`)
 The endpoint-type to send requests to on the server. This is only used with the
 `openai` service-kind. (default: `None`)
 
+##### `--server-metrics-url <list>`
+
+The list of Triton server metrics URLs. These are used for Telemetry metric
+reporting with the Triton service-kind. Example usage: --server-metrics-url
+http://server1:8002/metrics http://server2:8002/metrics.
+(default: `http://localhost:8002/metrics`)
+
 ##### `--service-kind {triton,openai}`
 
 The kind of service perf_analyzer will generate load for. In order to use
--- a/genai-perf/genai_perf/export_data/exporter_config.py
+++ b/genai-perf/genai_perf/export_data/exporter_config.py
@@ -1,4 +1,4 @@
-# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2024-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
--- a/genai-perf/genai_perf/export_data/exporter_config.py
+++ b/genai-perf/genai_perf/export_data/exporter_config.py
@@ -26,9 +26,7 @@
 
 
 import argparse as args
-from dataclasses import dataclass
-from dataclasses import field
-from dataclasses import field as default_factory
+from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Dict
 
--- a/genai-perf/genai_perf/export_data/exporter_config.py
+++ b/genai-perf/genai_perf/export_data/exporter_config.py
@@ -27,8 +27,10 @@
 
 import argparse as args
 from dataclasses import dataclass
+from dataclasses import field
+from dataclasses import field as default_factory
 from pathlib import Path
-from typing import Any, Dict, Optional
+from typing import Any, Dict
 
 from genai_perf.metrics import Metrics
 
@@ -40,4 +42,4 @@ class ExporterConfig:
     args: args.Namespace
     extra_inputs: Dict[str, Any]
     artifact_dir: Path
-    telemetry_stats: Optional[Dict[str, Any]] = None
+    telemetry_stats: Dict[str, Any] = field(default_factory=dict)
--- a/genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
+++ b/genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
@@ -1,4 +1,4 @@
-# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2024-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
--- a/genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
+++ b/genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
@@ -60,7 +60,14 @@ TELEMETRY_GROUPS = {
 
 
 def merge_telemetry_stats_json(telemetry_stats: Dict, stats_and_args: Dict) -> None:
-    stats_and_args.update({"telemetry_stats": telemetry_stats})
+    filtered_telemetry_stats = {
+        metric_name: metric_data
+        for metric_name, metric_data in telemetry_stats.items()
+        if any(gpu_data for gpu_id, gpu_data in metric_data.items() if gpu_id != "unit")
+    }
+
+    if filtered_telemetry_stats:
+        stats_and_args.update({"telemetry_stats": filtered_telemetry_stats})
 
 
 def export_telemetry_stats_csv(telemetry_stats: Dict, csv_writer) -> None:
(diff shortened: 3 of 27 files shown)


genai-perf/genai_perf/parser.py
```diff
@@ -637,11 +638,12 @@ def _add_endpoint_args(parser):
     endpoint_group.add_argument(
         "--server-metrics-url",
         type=str,
-        default=None,
+        nargs="+",
```
lkomali, line 641: When we implemented the first time, we expected that there would be only 1 url so a new arg --server-metrics-url was the best solution then. But I think when aggregating from multiple endpoints, I can see this becoming cluttered. We should move to a config file to read the metrics urls ( maybe another better approach to deal with a long list of urls), maybe not in this PR but in a following PR or whenever possible.

genai-perf/genai_perf/subcommand/common.py
```diff
@@ -192,11 +200,11 @@ def run_perf_analyzer(
...
     extra_args: Optional[List[str]] = None,
     perf_analyzer_config: Optional[PerfAnalyzerConfig] = None,
-    telemetry_data_collector: Optional[TelemetryDataCollector] = None,
+    telemetry_data_collectors: List[TelemetryDataCollector] = [],
 ) -> None:
     try:
-        if telemetry_data_collector is not None:
-            telemetry_data_collector.start()
+        for collector in telemetry_data_collectors:
+            collector.start()
```
lkomali, line 207: When I implemented the collector,  I was dealing with one metric url that will only run when PA subcommand is running, so I didn't think thoroughly about the resource usage. But now, we are starting one collector for one endpoint. I think this will quickly increase resource usage. 
We should go away from threads and maybe use a better way (asyncio or multiprocessing), I'm not sure of the best alternative. Again, this is something I'm suggesting that we should do quickly in the near future, it doesn't need to be in this PR.

dyastremsky: Good point! For now, it might be okay since this should only be used for a small number of endpoints and PA is a separate process, but asyncio would be a great follow-on approach soon. I could add it as an enhancement ticket after this one.

genai-perf/genai_perf/parser.py
dyastremsky: Very true. I think we're moving toward config files soon. As a more generic change though, so it should hopefully not be needed specifically for this feature.

lkomali:
Since you are adding more support to telemetry data, can you also move all the calculations inside the for loop [here](https://github.com/triton-inference-server/perf_analyzer/blob/9e2035bb169592e93c492d74e7f0ca1690e3dba6/genai-perf/genai_perf/metrics/telemetry_statistics.py#L55-L73)? This will ensure correct calculations.
I have a ticket for it but it might take some time to get to it. 
I will add the unit test for the telemetry_statistics in a new PR.  I should probably add a CI too for this.

genai-perf/tests/test_wrapper.py
```diff
@@ -157,7 +157,7 @@ def test_stdout_verbose(self, mock_telemetry_collector, mock_subprocess_run):
         run_perf_analyzer(
             args=args,
             extra_args=None,
-            telemetry_data_collector=telemetry_data_collector,
+            telemetry_data_collectors=[telemetry_data_collector],
```
lkomali, line 160: These tests related to telemetry collectors should ideally be out of this file because the current wrapper doesn't have any code related to telemetry collectors. 
We probably should have a `test_common.py` and move these two unit tests there.

nv-braf:
I don't see this causing a problem in analyze, but it's an interface change that I'll need to be aware of when making my config changes. Thanks for the heads up.

genai-perf/genai_perf/subcommand/common.py
zoramt: Just from our use cases, I don't see us needing more than a handful, say 2-10, of metrics url endpoints.

genai-perf/genai_perf/metrics/telemetry_metrics.py
```diff
@@ -58,19 +58,19 @@ class TelemetryMetrics:
...
-        self.total_gpu_memory = total_gpu_memory
-        self.gpu_memory_used = gpu_memory_used
+        gpu_power_usage: Optional[Dict[str, List[float]]] = None,
+        gpu_power_limit: Optional[Dict[str, List[float]]] = None,
+        energy_consumption: Optional[Dict[str, List[float]]] = None,
+        gpu_utilization: Optional[Dict[str, List[float]]] = None,
+        total_gpu_memory: Optional[Dict[str, List[float]]] = None,
+        gpu_memory_used: Optional[Dict[str, List[float]]] = None,
+    ):
+        self.gpu_power_usage = defaultdict(list, gpu_power_usage or {})
```
dyastremsky, line 68: Using default dictionaries in initialization parameters is dangerous if they later get updated. The state is shared among default dictionaries/lsits.

The new unit tests revealed this. One unit test used the default value and appended to the dictionary in the Telemetry metrics, then the next unexpectedly had metric values when initialized with default values. This is the more standard approach when a parameter can have a default dictionary to avoid sharing state.

genai-perf/genai_perf/export_data/exporter_config.py
```diff
@@ -40,4 +40,4 @@ class ExporterConfig:
     args: args.Namespace
     extra_inputs: Dict[str, Any]
     artifact_dir: Path
-    telemetry_stats: Optional[Dict[str, Any]] = None
```
dyastremsky, line 43: Telemetry_stats should never actually be None, so changing type checking to be more precise.

genai-perf/genai_perf/export_data/telemetry_data_exporter_util.py
```diff
@@ -59,35 +59,47 @@
 }
 
 
-def merge_telemetry_stats_json(
-    telemetry_stats: Optional[Dict], stats_and_args: Dict
-) -> None:
-    if telemetry_stats is not None:
-        stats_and_args.update({"telemetry_stats": telemetry_stats})
+def merge_telemetry_stats_json(telemetry_stats: Dict, stats_and_args: Dict) -> None:
+    filtered_telemetry_stats = {
```
dyastremsky, line 63: Checks if there are any Telemetry statistics collected by filtering for any actual data other than the unit name (which is there by default).

genai-perf/genai_perf/metrics/telemetry_statistics.py
```diff
@@ -56,21 +56,26 @@ def __init__(self, metrics: TelemetryMetrics):
...
-                    self._stats_dict[attr][gpu_index][
-                        percentile_label
-                    ] = percentile_value
-                min, max = self._statistics._calculate_minmax(gpu_data)
-                self._stats_dict[attr][gpu_index]["min"] = min
-                self._stats_dict[attr][gpu_index]["max"] = max
-                self._stats_dict[attr][gpu_index]["std"] = (
-                    self._statistics._calculate_std(gpu_data)
-                )
+                if not self._is_constant_metric(attr):
```
dyastremsky, line 59: Fixing this. It was meant to be inside the loop but was outside the loop, so only the last iteration of Telemetry metrics were captured.

genai-perf/tests/test_common.py
```diff
@@ -0,0 +1,54 @@
...
+# limitations under the License.
+
+import subprocess
+from unittest.mock import MagicMock, patch
+
+from genai_perf.subcommand.common import run_perf_analyzer
+
+
+class TestCommon:
+    @patch("genai_perf.subcommand.common.subprocess.run")
```
dyastremsky, line 22: Per Harshini's request, moved these into a test_common.py folder because run_perf_analyzer is a shared function in common.py.

genai-perf/README.md
```diff
@@ -415,6 +415,13 @@ Set a custom endpoint that differs from the OpenAI defaults. (default: `None`)
 The endpoint-type to send requests to on the server. This is only used with the
 `openai` service-kind. (default: `None`)
 
+##### `--server-metrics-url <list>`
```
dyastremsky, line 418: Added this missing section

lkomali:
Just one nit, but otherwise the PR looks good to me. I'll wait for others to review and share any suggestions before approving.

genai-perf/genai_perf/metrics/telemetry_statistics.py
debermudez: Thats a nefarious little bug.
I feel like robust testing should catch this type of bug.

dyastremsky:
> Just one nit, but otherwise the PR looks good to me. I'll wait for others to review and share any suggestions before approving.

@lkomali Thanks! Do you mean your nits above that were resolved? If not, I don't see it.

genai-perf/genai_perf/parser.py
debermudez: That is the plan in the mid term.
This is a great extension until we are ready for that.

debermudez (dismissed):

genai-perf/genai_perf/parser.py
```diff
@@ -637,11 +650,13 @@ def _add_endpoint_args(parser):
     endpoint_group.add_argument(
         "--server-metrics-url",
```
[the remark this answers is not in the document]
lkomali: I forgot to add this comment lol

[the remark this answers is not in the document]
dyastremsky: Good point! I was thinking about this too previously. I put it in our GA release tracker. We can't remove this name until our GA release because it'd be a breaking change. I'll make a follow-up ticket to add this alias to avoid delaying the merge.

debermudez (approved):

PR #253, opened 2025-01-24
Add chat template option for TRT-LLM engine

debermudez:
Allow the TRT-LLM engine to use extra inputs to receive an arg to apply the tokenizer's chat template, so that it can be used in generating tokens. This is only necessary for TRT-LLM engine because that is the only path that uses token IDs directly.


genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
```diff
@@ -82,4 +82,35 @@ def _add_request_params(self, payload: Dict, config: InputsConfig) -> None:
...
-            payload[key] = [value]
+            if key == "set_end_id":
+                payload["end_id"] = [config.tokenizer._tokenizer.eos_token_id]
+            elif key == "apply_chat_template":
+                pass
+            else:
+                payload[key] = [value]
+
+    def _encode_tokens(self, prompt: str, config: InputsConfig) -> List[int]:
+        if config.extra_inputs.__contains__("apply_chat_template"):
```
nv-hwoo, line 93:
```suggestion
        if "apply_chat_template" in config.extra_inputs:
```
Is this a viable alternative?

nicolasnoble: I was about to suggest the same :)

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
```diff
@@ -295,8 +295,11 @@ def _get_tensorrtllm_engine_token_counts(
         for r in res_outputs:
             if isinstance(r["output_ids"], list):
                 token_ids += r["output_ids"]
-            else:
+            elif isinstance(r["output_ids"], int):
                 token_ids.append(r["output_ids"])
+            else:
+                # for the empty first/last responses
+                token_ids.append(0)
```
nicolasnoble, line 302: We _probably_ want this as a constant somewhere instead of hardcoding 0 here?

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
```diff
@@ -82,4 +82,35 @@ def _add_request_params(self, payload: Dict, config: InputsConfig) -> None:
...
+            token_ids = self._encode_with_chat_template(prompt, config)
+        else:
+            token_ids = config.tokenizer.encode(prompt, add_special_tokens=False)
+        return token_ids
+
+    def _encode_with_chat_template(
```
matthewkotila, line 99: This is only doing art-llm specific logic, right? should the function name reflect that?

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
matthewkotila, line 93: Similar here, if we're detecting for this specific extra input and we're doing trt-llm specific logic, should the input name be trt-specific?

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
```diff
@@ -82,4 +82,35 @@ def _add_request_params(self, payload: Dict, config: InputsConfig) -> None:
...
+        """
+        Apply the default TRT-LLM engine chat template to the prompt
+        """
+        try:
+            import jinja2
+        except ImportError:
+            raise ImportError(
+                "Jinja2 is required for using TRT-LLM with chat template processing. Install it using: pip install jinja2."
+            )
+        default_template = [{"role": "user", "content": prompt}]
```
nicolasnoble, line 111: Shouldn't we move this to a more global location for easier overriding of the default template if we want to? This being so local here feels a bit wrong.

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
nv-hwoo: This method lives inside `TensorRTLLMEngineConverter` class only so I think function name is fine.

dyastremsky: This is in the TRT-LLM engine converter, so that might be redundant. Each of the converters' logic is specific to its endpoints.

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
dyastremsky: Same as above. I don't think we currently include the converter name in its functions.

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
dyastremsky: Yeah, I was unsure about this. It's specific to this converter. I can move it to the top of the class if that's helpful. Let me do that now.

dyastremsky: This is the only place this can go, because the prompt is a local value. Do you prefer a different approach?

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
dyastremsky: Great suggestion, fixed!

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
matthewkotila: Good catch--didn't see that high up the chain lol.

The perils of reviewing code so deep in something I know very little about

matthewkotila:
LGTM--deferring to others

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
nicolasnoble: Maybe potentially a method to construct it so it's more clear? I'm not too hung out about this one.

dyastremsky: How's this? https://github.com/triton-inference-server/perf_analyzer/pull/253/commits/d693c583d7f93c45e283ca7b230406e2fb3c03e7

nicolasnoble (approved):

nv-hwoo (approved):

PR #255, opened 2025-01-27
Add pre-commit hook to auto-update copyright year

dyastremsky:
Modeled after the server repo's copyright [pre-commit hook script](https://github.com/triton-inference-server/server/blob/f057b04b62cf8863ecd3b8b730e4fc275efc5f80/tools/add_copyright.py), add a copyright pre-commit hook to the PA repo. The script is pretty much the same except for some type fixes and the removal of the tests within the script.

I also updated the copyright, since the previous one stated the copyright belongs to Triton Inference Server instead of NVIDIA, which is what we use everywhere including in the files in this repo.

This hook auto-updates the end range of the copyright year for every file we touch, when needed (e.g. file being updated in 2025 with a copyright of 2021-2024 will be updated to 2021-2025).

See this example:
image

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -34,5 +34,12 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
     - uses: actions/checkout@v3
+      with:
+        fetch-depth: 2
+    - name: Get modified files
+      id: modified-files
+      run: echo "modified_files=$(git diff --name-only -r HEAD^1 HEAD | xargs)" >> $GITHUB_OUTPUT
     - uses: actions/setup-python@v3
     - uses: pre-commit/action@v3.0.0
+      with:
+        extra_args: --files ${{ steps.modified-files.outputs.modified_files }}
--- a/LICENSE
+++ b/LICENSE
@@ -1,28 +1,25 @@
-BSD 3-Clause License
-
-Copyright (c) 2024, Triton Inference Server
+Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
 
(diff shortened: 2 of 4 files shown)


LICENSE
```diff
@@ -1,28 +1,25 @@
-BSD 3-Clause License
-
-Copyright (c) 2024, Triton Inference Server
+Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
```
debermudez, line 1: There is a copyright here but do we not need a License type?

.github/workflows/pre-commit.yml
```diff
@@ -34,5 +34,12 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
     - uses: actions/checkout@v3
+      with:
```
dyastremsky, line 37: This updates the pre-commit hooks to only run on files that were modified as part of this pull request.

LICENSE
debermudez: got it. thanks for double checking

debermudez (approved):

.github/workflows/pre-commit.yml
matthewkotila, line 1: Does this include both python and c++ source files?

dyastremsky: It does! There are multiple handlers. The C++ one is here: https://github.com/triton-inference-server/perf_analyzer/pull/255/files#diff-427b28d805a7a2ab10e3b0f0f50dc513a03f308234ce0e9b06adca0aec358483R235

We might not have all of those file types. Those are the ones used by the server repo. If we find other extensions we need to use this for, we can easily add handlers.

PR #259, opened 2025-01-29
Dynamic gRPC client support

nv-hwoo:
This is an initial effort to support streaming NIMs using gRPC protocol. The PR introduces a new gRPC client and client backend that enables PA to send requests to gRPC server dynamically without requiring PA to pre-compile the protobuf to generate codes for gRPC client stub.

The current version of the client only supports bidirectional streaming RPC and will support other RPC types in the future.

#### Brief Example
1. User writes a short script (with language of their choice) that creates protobuf messages to send and writes it to a file stream:
```python
# input_stream.py
import  sys

def generate_msgs():
    # generate protobuf messages
    yield ...
    yield ...
    ...

for msg in generate_msgs():
    serialized = msg.SerializeToString()
    # encode msg size
    sys.stdout.buffer.write(len(serialized).to_bytes(4, byteorder=sys.byteorder))
    # encode actual msg
    sys.stdout.buffer.write(serialized)
    sys.stdout.buffer.flush()
```
2. Create JSON file to specify the command to run the script (e.g. `inputs.json`):
```json
{
  "data": [
    {
      "message_generator": "python3 input_stream.py"
    }
  ]
}
```
4. Run Perf Analyzer:
```
perf_analyzer -u localhost:8004 --service-kind dynamic_grpc --input-data inputs.json --rpc <full grpc method name>
```

PA will execute the user provided command, read/store the messages generated by user script, and send them over to server (in the same order as the user sent them).


CMakeLists.txt
matthewkotila, line 1: @nv-hwoo can you periodically rebase onto the base branch ([main](https://github.com/triton-inference-server/perf_analyzer/tree/main))?

And can you look into the CI failures? (I know L0_https is a known failure but there were some GAP failures too)

nv-hwoo: Thanks for reminder. I merged main into the branch. I checked the CI for GAP failures, and they were all cuda compute compatibility issue that we recently fixed. I will run the CI again.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
matthewkotila, line 1: I think some auto-copyright stuff messed up. We want to preserve the original year of creation for the file.

nicolasnoble:
So I'm not 100% certain the `ReadDataFromPipe` system does the right thing, as I think it means it's a one-shot stream at the moment, but we can alter this later without too much problem, I think.

What I'd like to see documented maybe into the .md file would be the schema of the json file you're adding? I don't think I've seen it fully documented yet.

docs/benchmarking.md
```diff
@@ -149,14 +149,98 @@ Inferences/Second vs. Client Average Batch Latency
...
+    yield ...   # Add more messages as required
+
+for msg in generate_msgs():
+    serialized = msg.SerializeToString()
+    # Write the message length as 4-byte integer (using system byte order)
+    sys.stdout.buffer.write(len(serialized).to_bytes(4, byteorder=sys.byteorder))
+    # Write the serialized message itself
+    sys.stdout.buffer.write(serialized)
+    sys.stdout.buffer.flush()
+```
```
matthewkotila, line 200: Mentioned this on calls, but I do think a more complete example that is copy-and-paste-able would help a lot here.

docs/benchmarking.md
```diff
@@ -149,14 +149,98 @@ Inferences/Second vs. Client Average Batch Latency
...
+```json
+{
+  "data": [
+    {
+      "ipc_stream": "python3 input_stream.py"
+    }
+  ]
+}
+```
+(Note: The "ipc_stream" key holds the command that will be executed to generate the messages.)
```
matthewkotila, line 214: Was this method of supplying GAP/PA with the grpc messages approved by the stakeholders?

docs/benchmarking.md
```diff
@@ -149,14 +149,98 @@ Inferences/Second vs. Client Average Batch Latency
...
+
+With both the generator script and the JSON configuration in place,
+run Perf Analyzer using the Dynamic gRPC service kind.
+Replace `<URL>`, `<package>`, `<service>`, and `<method>` with the appropriate values for your gRPC service:
+
+```bash
+perf_analyzer --service-kind=dynamic_grpc -u=<URL> --input-data=inputs.json  --rpc=<package>.<service>/<method> --streaming
```
matthewkotila, line 221: Not sure where we left off on the call from last week, but I'm still not a fan of requiring the user to supply `--streaming` when the lack of it is invalid.

I think that variable should be ignored and not required if `--service-kind=dynamic_grpc`.

docs/cli.md
```diff
@@ -470,6 +472,18 @@ Only supported when gRPC protocol is being used.
...
+Specifies a fully-qualified gRPC method name in `<package>.<service>/<method>` format.
+The option is only supported with `dynamic_grpc` service kind and is used to identify
+the RPC to use when sending requests to the server.
+
+#### `--proto <string>`
+
+Specifies the path to the protobuf file that defines all the gRPC service and RPC methods.
+The options is only supported with `dynamic_grpc` service kind for dynamically parsing
+the protobuf at runtime.
+
```
matthewkotila, line 486: As mentioned, since this is unused, better to remove it.

docs/benchmarking.md
matthewkotila, line 221: Also, regarding the `--rpc` option name, it's referring to the fully qualified method name, right?

Should the option name reflect that more clearly? Is "rpc" a community-common shorthand for a fully qualified method name? Maybe `--grpc-method`?

src/infer_context.cc
```diff
@@ -145,36 +167,45 @@ InferContext::SendRequest(
...
-        infer_data_.outputs_);
+    thread_stat_->idle_timer.Start();
+    auto start_time_sync = std::chrono::system_clock::now();
+    if (streaming_) {
+      thread_stat_->status_ = infer_backend_->StreamInfer(
+          &results, *(infer_data_.options_), infer_data_.valid_inputs_,
+          infer_data_.outputs_);
+    } else {
+      thread_stat_->status_ = infer_backend_->Infer(
+          &results, *(infer_data_.options_), infer_data_.valid_inputs_,
```
matthewkotila, line 179: I don't think `StreamInfer` needs to exist. Especially considering all of the streaming is encapsulated below this point.

It takes the same parameters as Infer. Infer can just be used here and within the dgrpc backend, it can implement Infer the same way it currently implements StreamInfer.

src/client_backend/client_backend.h
```diff
@@ -28,6 +28,7 @@
 #include <rapidjson/document.h>
 #include <rapidjson/rapidjson.h>
 
+#include <chrono>
```
matthewkotila, line 31: We don't have time for this!

image

src/infer_context.cc
```diff
@@ -46,13 +46,34 @@ InferContext::Init()
...
+    return;
+  }
+
+  // Dynamic grpc client requires to restart the stream before every new request
+  // because the for each request, the stream is half-closed from the client
+  // side due to calling WritesDone.
+  if (streaming_ && inference_started_) {
+    infer_backend_->StopStream();
+    infer_backend_->StartStream(
+        async_callback_func_, (!parser_->IsDecoupled()));
```
matthewkotila, line 65: It looks like PrepareRequest is called for every request sent.

If the `DynamicGrpcClient::StartStream` and `DynamicGrpcClient::StopStream` really need to be called at the start and end of every request, then let's just encapsulate that logic inside of Infer so that we don't need to change more out here.

docs/benchmarking.md
matthewkotila, line 214: One call to `python3 input_stream.py` provides all of the messages for sending _one_ request, right?

If only one `ipc_stream` is provided in the input data JSON, those same messages get reused for all requests sent, right?

So if someone wants to send 100 unique requests, they'd have to create 100 entries in the input data json with presumably unique scripts (or unique calls to the same script)?

According to the stakeholders, is it common that they would want to send many requests all with different inputs? Because if so, perhaps this design should be rethought?

One off-the-cuff alternative would be to somehow allow a single `python3 input_stream.py` to supplied all messages for _all_ requests? (via a delimiter or something).

src/constants.h
```diff
@@ -43,6 +43,10 @@ constexpr static const uint32_t OPTION_ERROR = 3;
 constexpr static const uint32_t GENERIC_ERROR = 99;
 constexpr static const size_t DEFAULT_MAX_THREADS = 16;
 
+/// Size of the data being read from stream (e.g. pipe) when reading input data
+/// from user-provided data stream. Defaults to 4 bytes.
+constexpr uint32_t DEFAULT_STREAM_DATA_SIZE = 4;
```
matthewkotila, line 48: I don't like how this piece of information needs to be known in two distinct locations in the code, but I can't think of a clean way to encapsulate it...

Perhaps this will change anyway if DataLoader needs to be modified to provide more genuine/piecewise piping.

src/infer_context.cc
```diff
@@ -145,36 +167,45 @@ InferContext::SendRequest(
...
+
     if (results != nullptr) {
       if (thread_stat_->status_.IsOk()) {
         response_outputs = GetOutputs(*results);
         thread_stat_->status_ = ValidateOutputs(results);
+
+        // When streaming, use the timestamps collected from client
+        if (streaming_) {
+          results->ResponseTimestamps(&response_timestamps);
+        }
```
matthewkotila, line 196: I still feel a bit strange about how each received message of a response (i.e. calls to CompletionQueue::Next() until `ok` is false) has its own timestamp, but all of the messages sent for an outgoing request (i.e. calls to Write()) don't have their own timestamps.

Did the stakeholders express what kind of benchmark metrics/statistics they want?

matthewkotila (commented):
Done with my first review. Thanks for doing all of this work, Hyunjae. I know it's not easy to add something into PA.

src/infer_context.cc
nv-hwoo: Removed the dependency to streaming boolean.

src/infer_context.cc
nv-hwoo: Removed this part since stakeholders currently only care about the request latency and throughput (not the fine-grained collection of timestamps of messages received in a response).

docs/benchmarking.md
nv-hwoo: Done 👍

src/model_parser.h
```diff
@@ -68,9 +68,8 @@ class ModelParser {
     ENSEMBLE_SEQUENCE
   };
 
-  explicit ModelParser(cb::BackendKind backend_kind)
```
nv-hwoo, line 71: context: removed the parameter because it was being used nowhere in the model parser.

src/infer_context.cc
matthewkotila: But how do they want latency and throughout defined? Do they want the time between the first outgoing message of a single complete inference and the last incoming message of the same single inference to be considered one latency value in the benchmark?

docs/benchmarking.md
matthewkotila: Is there perhaps a simple server that can be pip installed as a package or something to skip us needing to write code for the server here?

src/infer_context.cc
nv-hwoo: They were fine with whatever the default way GAP/PA was measuring for request latency and throughput.

matthewkotila: I guess my question is more for us. Can we confirm how GAP/PA is defining latency for this bidirectional streaming feature?

nv-hwoo: The main reason why I added timestamps for response was because async stream case was doing it and I thought it would be a good to have parity. But it was not a requirement 😅 I was going a bit of ahead of myself.

matthewkotila: Oh so did you remove the set of timestamps associated with each incoming message of a single inference?

It's now just one timestamp for when the last message of the single inference came in?

CMakeLists.txt
nv-hwoo: Updated the copyrights

src/infer_context.cc
nv-hwoo: Now it's just the same as how we used to do for plain synchronous inference call: before & after the InferContext::Infer call (https://github.com/triton-inference-server/perf_analyzer/blob/f53136ae2a2c9be5c0eba4b1ab0fe81f953182df/src/infer_context.cc#L150-L156).

docs/input_data.md
```diff
@@ -304,3 +304,63 @@ your application. Use
...
+          "ipc_stream"
+        ]
+      }
+    }
+  },
+  "required": [
+    "data"
+  ]
+}
+```
```
matthewkotila, line 366: I might have missed the reason for this--is it just to show users how to use this feature?

The documentation with a simple example of `python3 input_stream.py` seems clear enough.

src/client_backend/dynamic_grpc/dynamic_grpc_client.h
```diff
@@ -161,32 +154,20 @@ class DynamicGrpcClient {
...
-  // The inference statistic of the current client
-  InferStat infer_stat_;
-
- private:
   // Generic bi-directional stream using dynamic protobuf message.
   std::unique_ptr<grpc::GenericClientAsyncReaderWriter> bidi_stream_;
   std::unique_ptr<grpc::ClientContext> grpc_context_;
   std::unique_ptr<grpc::CompletionQueue> completion_queue_;
   bool stream_started_{false};
+  bool writesdone_called_{false};
```
matthewkotila, line 170: Could this variable be renamed to something more semantic?

What does writes done being called actually mean?

Does that mean a single inference has had all of its messages sent?

Maybe a name like `inference_sent_` or `inference_messages_sent_` or something like that might be more educative?

src/command_line_parser.cc
```diff
@@ -2149,10 +2152,6 @@ CLParser::VerifyOptions()
         "Must specify --collect-metrics when using the --metrics-interval "
         "option.");
   }
-  if (params_->streaming && (params_->protocol != cb::ProtocolType::GRPC &&
-                             params_->kind != cb::BackendKind::TRITON_C_API)) {
-    Usage("Streaming is only allowed with gRPC protocol and Triton C API.");
-  }
```
matthewkotila: Does this if still need to be moved above?

src/data_loader.cc
```diff
@@ -268,10 +268,72 @@ DataLoader::ParseData(
...
+              std::to_string(data_size) +
+              ", bytes read = " + std::to_string(bytes_read),
+          pa::GENERIC_ERROR);
+    }
+
+    // Store data size and data, respectively, into input data
+    std::memcpy(size.data(), &data_size, sizeof(data_size));
+    std::copy(size.begin(), size.end(), std::back_inserter(it->second));
+    std::copy(buffer.begin(), buffer.end(), std::back_inserter(it->second));
+  }
```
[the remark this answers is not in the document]
matthewkotila: @nv-hwoo what was the feedback from the stakeholders for this feature? Are they ok with the current constraint that all unique benchmark inputs must fit in system memory?

src/constants.h
matthewkotila: Since I couldn't come up with a better way to encapsulate this, I'll just resolve the conversation.

docs/input_data.md
nv-hwoo: @nicolasnoble asked for json schema (unless I misunderstood his ask).

matthewkotila: Yeah I mean I guess regardless of who asked for it, I'm just trying to understand the value-add.

src/client_backend/dynamic_grpc/dynamic_grpc_client.h
nv-hwoo: Sure, calling [WritesDone()](https://grpc.github.io/grpc/cpp/classgrpc_1_1_client_reader_writer_interface.html#a52f4e5d5ac7fe0e4995cb337aa0ecfc8) means half-closing the stream from the client side, so maybe `stream_half_closed_`? This name will make it clear why we are restarting the stream in the beginning of the inference.

src/command_line_parser.cc
nv-hwoo: I think you might be looking at the old changes.

src/data_loader.cc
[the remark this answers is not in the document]
nv-hwoo: I've shared the constraint with the stakeholders, and I'm still waiting for their ballpark figure on the size of their dataset but I did get their confirmation that their usage will not be significant enough to exceed system memory.

src/command_line_parser.cc
```diff
@@ -187,6 +187,8 @@ CLParser::Usage(const std::string& msg)
   std::cerr << "\t--streaming" << std::endl;
   std::cerr << "\t--grpc-compression-algorithm <compression_algorithm>"
             << std::endl;
+  std::cerr << "\t--proto" << std::endl;
```
matthewkotila, line 190: This should be removed, yeah?

src/command_line_parser.cc
matthewkotila: Ah, it was originally up above, and when I filtered the "changes" to only show your recent commits, it was you undoing the change you did earlier where you moved it from up above to down below.

docs/input_data.md
nv-hwoo: I think it's possible that our input JSON usage seems confusing for anyone who's not familiar with PA. So the value add for providing schema for dynamic grpc case would be to help the users see the full spec of the JSON file and remove any question marks. At least, that's why I think @nicolasnoble asked for it (correct me if I am reading your mind wrong).

src/client_backend/dynamic_grpc/dynamic_grpc_client.h
```diff
@@ -0,0 +1,178 @@
...
+  /// \param options The options for inference request.
+  /// \param inputs The vector of InferInput describing the model inputs.
+  /// \param outputs Optional vector of InferRequestedOutput describing how the
+  /// output must be returned. If not provided then all the outputs in the model
+  /// config will be returned as default settings.
+  /// \param headers Optional map specifying additional HTTP headers to include
+  /// in the metadata of gRPC request.
+  /// \return Error object indicating success or failure of the
+  /// request.
+  Error BidiStreamRPC(
```
matthewkotila, line 124: I think including "inference" in the function name might be more understandable for a new reader.

The fact that it's using a bidirectional stream RPC is an implementation detail, but this class is intended to send inference requests (messages), essentially.

src/client_backend/dynamic_grpc/dynamic_grpc_client.h
matthewkotila: I think incorporating the concept of inference in this variable name would help a new reader understand its role within the context of the entire class, whose purpose is to send inference requests (messages).

matthewkotila: Idk, maybe it's fine to leave it as it is. It's just that when I read "writes done called" it doesn't help me understand what this variable is gonna be used for.

src/data_loader.cc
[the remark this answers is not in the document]
matthewkotila: We don't need to know any absolute sizes, just that they're ok with the constraint that whatever inputs they give cannot exceed whatever system memory they're running on.

src/client_backend/dynamic_grpc/dynamic_grpc_client.h
nv-hwoo: I would consider "inference" as a concept that exists higher above the dynamic grpc client, and dynamic grpc client doesn't really know/care how it's being used or called. The goal of dynamic grpc client is to support all of the four grpc method types in a generic manner:
- Unary RPC
- Client Streaming RPC
- Server Streaming RPC
- Bidirectional Streaming RPC

For the current phase, we are only introducing the bidirectional one since that's what the stakeholders need, but I foresee the dynamic grpc client to support the rest through separate APIs (in theory, BidiStreamRPC could handle other ones but some stakeholders may not want to establish stream for simple RPC calls).

src/client_backend/dynamic_grpc/dynamic_grpc_client.h
nv-hwoo: I hear your concern. I think I'm just trying to keep the grpc client to contain logics that makes sense in the context of grpc communication. I'll change to `stream_half_closed_` and make a small comment that WriteDone closes the stream. How's that?

docs/benchmarking.md
nv-hwoo: @matthewkotila would you like me to add this to the doc?

src/client_backend/dynamic_grpc/dynamic_grpc_client.h
matthewkotila: Sounds good 😄

src/client_backend/dynamic_grpc/dynamic_grpc_client.h
matthewkotila: Hmm, I guess that's fair. But then shouldn't ever mention of inference concept be scrubbed from the class? There's still mentions of inference in documentation and UpdateInferStat.

I'll resolve this thread, but something to think about for future, what levels of abstraction we want to have exposed to this level.

matthewkotila (approved):

docs/benchmarking.md
nv-hwoo: Updated the doc to include the instructions for setting up mock grpc server.

src/command_line_parser.h
```diff
@@ -144,8 +144,6 @@ struct PerfAnalyzerParameters {
   size_t warmup_request_count{0};
 
   size_t session_concurrency{0};
-  SessionConcurrencyMode session_concurrency_mode{
-      SessionConcurrencyMode::Disabled};
```
matthewkotila: Why is this removed?

nv-hwoo: There was a merge conflict but accepting the change still contained this part, so I had to remove it manually to match with the main. There's no more SessionConcurrencyMode in main branch.

matthewkotila (approved):

nv-hwoo:
Cancelled pipeline for the merge commit. Passing CI ID: 23990942

PR #260, opened 2025-01-29
Remove extra printing for copyright hook

dyastremsky:
Make the copyright hook not print out extra statements after running

Updated:
image

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -84,6 +84,4 @@ repos:
     name: Add License
     entry: python tools/add_copyright.py
     language: python
-    stages: [pre-commit]
-    verbose: true
     require_serial: true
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -84,6 +84,4 @@ repos:
     name: Add License
     entry: python tools/add_copyright.py
     language: python
-    stages: [pre-commit]
```
matthewkotila, line 87: What does removing this line do? Don't we want it to happen in the pre-commit still?

dyastremsky: I think this is the default value. Changing it did not change behavior when I tried committing again. None of the other hooks specify this stage, so I removed this for parity and brevity.

matthewkotila (approved):

PR #261, opened 2025-01-30
Add session args and timing args for multi-turn chat

lkomali:
This PR adds support for following args required for multi-turn chat feature
`--num-sessions:` default 0 which means every request is individual and it's not multi turn
`--turns-per-session-mean`: Mean turns per session, default 1
`--turns-per-session-stddev`: The standard deviation of the number of turns per session, default 0
`--session-turn-delay-mean`: Mean delay per turn in a session, default 1000ms
`--session-turn-delay-stddev`: The standard deviation of the delay between turns in a " "session, default 200ms
`-session-concurrency`: The number of concurrent sessions to simulate. It's the default mode when user doesn't parse any value for session-concurrency or session-rate and num-sessions > 1, default concurrency is 1
`--session-rate`: Rate at which sessions are started

Added a method `_check_session_args` to validate if the above args are correctly used.

Updated unit test `test_cli.py` to test different scenarios.

Additionally, updated `Readme`, `README_template`, `test_json_exporter.py` as needed.


dyastremsky:
> How to fix this pre-commit failure? I used the command pinned to slack channel and it passed locally.

Pre-commit should run automatically whenever you commit files. You shouldn't use the top-of-thread command. If you read inside that thread, I had details for what to do if you use that command in your workflow. You need to revert the changes. You updated the copyrights for every single file, including ones untouched by this pull request. That conversation has a way to undo this if done by accident before committing, but you already committed and pushed those changes. You'd need to update the git diff in the script to get the diff against feature-multiturn-chat or undo this manually.

As to your original question, it sounds like there's a discrepancy between your local pre-commit and the one on the GitHub side, though I don't know why that would be. I would re-run `pre-commit install` or try `pre-commit autoupdate`. You can also go into the failing hook and copy the changes pre-commit made as a temporary fix.

CC: @debermudez, we might want to unpin that conversation to avoid confusion if someone copies that command without reading the thread. We want people to let pre-commit run automatically on changed files in a commit rather than manually run it on all files.

lkomali:
I've has issues with pre-commit like it doesn't run automatically for me for some reason. That's the reason why i thought i should probably run the command. I read the thread but wasn't sure, if it will run automatically. I should've probably not ran it manually and then decide.

genai-perf/genai_perf/inputs/input_constants.py
```diff
@@ -87,6 +87,11 @@ def to_lowercase(self):
 DEFAULT_NUM_DATASET_ENTRIES = 100
 DEFAULT_NUM_PREFIX_PROMPTS = 0
 DEFAULT_PREFIX_PROMPT_LENGTH = 100
+DEFAULT_NUM_SESSIONS = 0
+DEFAULT_TURNS_PER_SESSION_MEAN = 1
+DEFAULT_TURNS_PER_SESSION_STDDEV = 0
+DEFAULT_SESSION_TURN_DELAY_MEAN_MS = 1000
+DEFAULT_SESSION_TURN_DELAY_STDDEV_MS = 200
```
lkomali, line 94: I thought that if someone wants to benchmark multi turn without providing turn_delay, they should be able to do with default values. I asked perplexity like what values would better simulate real-time chat and it suggested delay_stddev 200ms and delay_mean 1000s.
Let me know if you think these shouldn't have default values and should always be given by the user.

dyastremsky:
> I've has issues with pre-commit like it doesn't run automatically for me for some reason. That's the reason why i thought i should probably run the command. I read the thread but wasn't sure, if it will run automatically. I should've probably not ran it manually and then decide.

Nice job reverting the changes. Let me know if you want to talk offline. I'd be curious how long it has not been running automatically for you. We can probably find the issue and fix it.

lkomali:
> > I've has issues with pre-commit like it doesn't run automatically for me for some reason. That's the reason why i thought i should probably run the command. I read the thread but wasn't sure, if it will run automatically. I should've probably not ran it manually and then decide.
> 
> Nice job reverting the changes. Let me know if you want to talk offline. I'd be curious how long it has not been running automatically for you. We can probably find the issue and fix it.

It's running autoatically now with the new update. I will reach out to you if I face the problem again.

genai-perf/tests/test_cli.py
```diff
@@ -932,6 +942,145 @@ def test_positive_image_input_args(self, monkeypatch, args):
         with pytest.raises(SystemExit) as excinfo:
             parser.parse_args()
 
+    def test_defaults_applied_when_sessions_enabled(self, monkeypatch):
```
dyastremsky, line 945: Nit: For naming, I'd prefer if the specific feature set you're testing (session) is first so that it's easier to identify the feature and sort by it. It'd also be good if you add a comment header to say that these are the session tests.

[Beyond the scope of this PR] Long-term, I think our team should break this test file into a directory of CLI test files, broken by major feature. This file is getting too lengthy and unwieldy. CC: @debermudez

genai-perf/genai_perf/parser.py
```diff
@@ -274,6 +274,54 @@ def _check_goodput_args(args):
...
 
+def _check_session_args(
+    parser: argparse.ArgumentParser, args: argparse.Namespace
+) -> argparse.Namespace:
+    """
+    Check if session args are used correctly
+    """
+    if args.num_sessions > 0:
+        if not args.session_concurrency and not args.session_rate:
+            args.session_concurrency = 1
```
dyastremsky, line 285: I suggest using a constant here.

genai-perf/tests/test_cli.py
```diff
@@ -932,6 +942,145 @@ def test_positive_image_input_args(self, monkeypatch, args):
...
+            ["--turns-per-session-stddev", "5"],
+            ["--session-turn-delay-mean", "1"],
+            ["--session-turn-delay-stddev", "0.2"],
+        ],
+    )
+    def test_error_on_disallowed_args_with_zero_sessions(self, monkeypatch, args):
+        combined_args = ["genai-perf", "profile", "--num-sessions", "0"] + args
+        monkeypatch.setattr("sys.argv", combined_args)
+
+        with pytest.raises(SystemExit):
```
dyastremsky, line 1006: A more specific error/exception assertion here could be good.

genai-perf/tests/test_cli.py
```diff
@@ -932,6 +942,145 @@ def test_positive_image_input_args(self, monkeypatch, args):
...
+            "--num-sessions",
+            "5",
+            "--session-concurrency",
+            "2",
+            "--session-rate",
+            "3",
+        ]
+        monkeypatch.setattr("sys.argv", combined_args)
+
+        with pytest.raises(SystemExit):
```
dyastremsky, line 988: A more specific error/exception assertion here could be good.

genai-perf/tests/test_cli.py
```diff
@@ -932,6 +942,145 @@ def test_positive_image_input_args(self, monkeypatch, args):
...
+            "-m",
+            "test_model",
+            "--num-sessions",
+            "2",
+            "--session-rate",
+            "3",
+        ]
+        monkeypatch.setattr("sys.argv", combined_args)
+
+        with pytest.raises(SystemExit):
```
dyastremsky, line 1038: A more specific error/exception test here could be good.

genai-perf/tests/test_cli.py
```diff
@@ -932,6 +942,145 @@ def test_positive_image_input_args(self, monkeypatch, args):
...
+            parser.parse_args()
+
+    def test_valid_session_args(self, monkeypatch):
+        NUM_SESSIONS = 5
```
dyastremsky, line 1042: I don't think you should be defining constants within a function in Python. I don't think we do it in other tests or that it's considered best practice. Let me know if that's not true.

If it's not, making these lowercase would be good.

templates/genai-perf-templates/README_template
```diff
@@ -589,6 +589,39 @@ exported to `<profile_export_file>_genai_perf.json` and
...
+##### `--session-turn-delay-stddev`
+The standard deviation of the delay (in milliseconds) between turns in a session.
+(default: `200`)
+
+##### `--session-concurrency`
+The number of concurrent sessions to be simulated. This is the default mode for simulating sessions.
+(default: `1`)
+
+##### `--session-rate`
+The rate at which sessions are started.
```
dyastremsky, line 623: Does this have a default value?

templates/genai-perf-templates/README_template
```diff
@@ -589,6 +589,39 @@ exported to `<profile_export_file>_genai_perf.json` and
...
+This is only used when --num-sessions is provided.
+(default: `1`)
+
+##### `--turns-per-session-stddev`
+
+The standard deviation of the number of turns per session.
+This is only used when --turns-per-session-mean is provided.
+(default: `0`)
+
+##### `--session-turn-delay-mean`
```
dyastremsky, line 610: I think this and the args after it need an extra newline for consistency.

genai-perf/genai_perf/parser.py
```diff
@@ -274,6 +274,54 @@ def _check_goodput_args(args):
...
+        if invalid_options:
+            parser.error(
+                f"The following arguments require --num-sessions > 0: {', '.join(invalid_options)}."
+            )
+
+    if args.session_concurrency and args.session_concurrency > args.num_sessions:
+        parser.error("--session-concurrency cannot be greater than --num-sessions. ")
+
+    if args.session_rate and args.session_rate > args.num_sessions:
+        parser.error("--session-rate cannot be greater than --num-sessions. ")
```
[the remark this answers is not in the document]
lkomali: I think session-rate doesn't have to be < num-sessions. It makes sense to be higher.

genai-perf/tests/test_cli.py
lkomali: Yeah, i kind of forgot it. Will change.

templates/genai-perf-templates/README_template
lkomali: This doesn't have a default value. The default mode is session-concurrency and default value is 1, so if someone wants to use it have to set it explicitly, just like how we do for concurrency and request-rate today.

dyastremsky: Can you make the default None to be consistent with the way we document request-rate?

genai-perf/genai_perf/parser.py
```diff
@@ -274,6 +274,54 @@ def _check_goodput_args(args):
...
 
 
+def _check_session_args(
+    parser: argparse.ArgumentParser, args: argparse.Namespace
+) -> argparse.Namespace:
+    """
+    Check if session args are used correctly
+    """
+    if args.num_sessions > 0:
+        if not args.session_concurrency and not args.session_rate:
```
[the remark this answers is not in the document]
debermudez: This would be in addition to @dyastremsky's comment. Use the defaults but check if num sessions is not set but other args are.

genai-perf/tests/test_cli.py
```diff
@@ -932,6 +942,125 @@ def test_positive_image_input_args(self, monkeypatch, args):
         with pytest.raises(SystemExit) as excinfo:
             parser.parse_args()
 
+    def test_multiturn_defaults_applied(self, monkeypatch):
```
debermudez, line 945: I would add a test around num_sessions not being set but the other session args are and lock down the expected behavior.

templates/genai-perf-templates/README_template
```diff
@@ -589,6 +590,43 @@ exported to `<profile_export_file>_genai_perf.json` and
 export file is `profile_export.json`, the genai-perf file will be exported to
 `profile_export_genai_perf.csv`. (default: `profile_export.json`)
 
+##### `--num-sessions`
```
debermudez, line 593: Alphabetize

templates/genai-perf-templates/README_template
```diff
@@ -589,6 +590,43 @@ exported to `<profile_export_file>_genai_perf.json` and
...
+
+##### `--turns-per-session-stddev`
+
+The standard deviation of the number of turns per session.
+This is only used when --turns-per-session-mean is provided.
+(default: `0`)
+
+##### `--session-turn-delay-mean`
+
+The mean delay (in milliseconds) between turns in a session.
```
debermudez, line 613: What happens if the response isnt back before the delay is up? Do we need to outline that behavior here?

genai-perf/genai_perf/parser.py
```diff
@@ -274,6 +274,54 @@ def _check_goodput_args(args):
...
+            args.session_turn_delay_mean = ic.DEFAULT_SESSION_TURN_DELAY_MEAN_MS
+        if not args.session_turn_delay_stddev:
+            args.session_turn_delay_stddev = ic.DEFAULT_SESSION_TURN_DELAY_STDDEV_MS
+
+    if args.session_concurrency and args.session_rate:
+        parser.error(
+            "--session-concurrency and --session-rate cannot be used together."
+        )
+
+    if args.num_sessions == 0:
```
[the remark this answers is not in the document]
debermudez: I think i highlighted the opposite behavior in another comment. I don't love silent decisions and an early warning can save hours on a profile and meaningful $$$ if the server being benchmarked is on a cloud instance. 
You are right, we cant test the full matrix everywhere, but its a balance.

templates/genai-perf-templates/README_template
lkomali: I don't know how PA handles when the response isn't back before delay. But it isn't GAP's duty to decide what to do if the response isn't back before delay. It's the PA's work, and maybe should be explained there?

genai-perf/tests/test_cli.py
lkomali: i had that test earlier, but when i remove not throwing an error, i removed the test too.

genai-perf/tests/test_exporters/test_json_exporter.py
```diff
@@ -232,6 +232,7 @@ class TestJsonExporter:
           "output_tokens_stddev": 0,
           "random_seed": 0,
           "request_count": 0,
+          "session_concurrency": 1,
```
dyastremsky, line 235: Why is this getting set? I thought this would only get if there is a session concurrency specified.

genai-perf/README.md
```diff
@@ -581,6 +581,13 @@ stable. The measurement is considered as stable if the ratio of max / min from
 the recent 3 measurements is within (stability percentage) in terms of both
 infer per second and latency. (default: `999`)
 
+### Session Options
+
+##### `--session-concurrency`
+
+The number of concurrent sessions to be simulated. This is the default mode for simulating sessions.
+(default: `1`)
```
dyastremsky, line 589: It looked like the default is None? Could we specify that this mode will only be used if X (whatever X is... since this isn't used by default if the benchmark mode isn't sessions).

genai-perf/tests/test_exporters/test_json_exporter.py
lkomali: Due to me setting session-concurrency 1 in add_session_args.
Earlier, I was checking if `num_session > 0` and then setting `session_concurrency 1`, but since i removed num-sessions 1 condition check, this part is being added by default.
We don't have a way to figure if someone wants to use multiturn file path, so I left it here and will update condition once that is figured.

genai-perf/README.md
dyastremsky: I see, you set it in check_session_args. I find that a bit misleading. What do you think about setting the default value in the arg itself (under `_add_session_args(parser)`) like we do for other args?

lkomali: Should we set default or not have any default and user has to pass in session-concurrency? 
Because if we set default, it's added even when not using multiturn path.

dyastremsky: I need to think about this. I do definitely think it needs to be coded + documented that this is only used when session IDs are passed in.

genai-perf/tests/test_exporters/test_json_exporter.py
dyastremsky: Will we have add that checking before sending to PA? Do you have ideas for what the design would look like?

We can leave it as is here but it sounds like this will break things until we find a way to confirm that sessions are being used.

lkomali: Right, so the problem is we have `--input-file payload:<path to file>` which can be for fixed schedule or multiturn as of now. We don't know which one user is providing. Until we figure out how to differentiate both, it's difficult to figure which path user wants to take. 
We could use `--input multiturn:<path to file>` for this path. That seems like one way to go or maybe ask GAP user to pass `--multiturn` option or split input-file option into two options (discussion in slack). I'm not sure, which is a better way and this requires some discussion.

dyastremsky: Okay, so I think we should have the default be None. If the user passes in a multi-turn case without specifying a session concurrency or session rate, PA can handle returning an error or using whatever its default value is.

The documentation should make it clear that it's only used when sessions are specified.

lkomali: sounds good, i will remove _check_session_args for now, can be added later if required and update docs that it should be added only benchmarking sessions.

dyastremsky (approved):

PR #262, opened 2025-01-30
Allow extra-inputs to support keys without values

dyastremsky:
In some case, extra-inputs will just be used to enable certain functionalities (e.g. in the TRT-LLM engine converter, the use of "set_end_id") and do not require a value. Allow extra-inputs to not require a value for easier usability for those use cases.

### DIFF
--- a/genai-perf/genai_perf/subcommand/common.py
+++ b/genai-perf/genai_perf/subcommand/common.py
@@ -89,15 +89,16 @@ def get_extra_inputs_as_dict(args: Namespace) -> Dict[str, Any]:
             semicolon_count = input_str.count(":")
             if input_str.startswith("{") and input_str.endswith("}"):
                 request_inputs.update(load_json_str(input_str))
-            elif semicolon_count == 0:
+            elif semicolon_count == 0:  # extra input as a flag
                 request_inputs[input_str] = None
             elif semicolon_count == 1:
                 input_name, value = input_str.split(":", 1)
 
                 if not input_name or not value:
                     raise ValueError(
-                        f"Input name or value is empty in --extra-inputs: {input_str}\n"
-                        "Expected input format: 'input_name:value'"
+                        f"Input name or value is empty in --extra-inputs: "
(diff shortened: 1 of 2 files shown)


debermudez (approved):

genai-perf/genai_perf/subcommand/common.py
```diff
@@ -121,6 +118,11 @@ def get_extra_inputs_as_dict(args: Namespace) -> Dict[str, Any]:
                         f"Input name already exists in request_inputs dictionary: {input_name}"
                     )
                 request_inputs[input_name] = value
+            else:
+                raise ValueError(
+                    f"Invalid input format for --extra-inputs: {input_str}\n"
+                    "Expected input format: 'input_name:value'"
```
nv-hwoo, line 124: Can we also update the error message to reflect the fact that no value (or flag) extra input is one of the expected format?

genai-perf/genai_perf/subcommand/common.py
```diff
@@ -83,18 +83,15 @@ def calculate_metrics(args: Namespace, tokenizer: Tokenizer) -> ProfileDataParse
...
             if input_str.startswith("{") and input_str.endswith("}"):
                 request_inputs.update(load_json_str(input_str))
-            else:
-                semicolon_count = input_str.count(":")
-                if semicolon_count != 1:
-                    raise ValueError(
-                        f"Invalid input format for --extra-inputs: {input_str}\n"
-                        "Expected input format: 'input_name:value'"
-                    )
+            elif semicolon_count == 0:
```
nv-hwoo, line 92: (nit) for better context
```suggestion
            elif semicolon_count == 0:  # extra input as a flag
```

dyastremsky: Done!

genai-perf/genai_perf/subcommand/common.py
dyastremsky: Good idea, done!

nv-hwoo (approved):
lgtm!

PR #263, opened 2025-01-31
Add session-args to PA subcommand when using multi-turn

lkomali:
This PR 
1. Adds a new function `add_session_args` in `wrapper.py` to add session-concurrency or session-rate to PA subcommand when passed by the user.
By default, if using multi-turn and neither of `session-concurrency` or `session-rate` is given by the GAP user, the default mode is `session-concurrency` and default `session-concurrency` is `1`.
2. Adds a unit test in in `test_wrapper.py` to test if the sessions args are added correctly.


dyastremsky (commented):
Great work! Can we add parameters to the test case for session rate and session concurrency being passed in without the num-sessions? That should work in the case of payload file, I think.

dyastremsky:
@debermudez I made tickets for and commented out the synthetic and session rate pieces to keep as much code as possible. This looked good to me otherwise. I also added the test I mentioned above. I think this would be good to merge, based on your review/approval.

If we want to keep it out of the larger feature branch to keep that one smaller, we could also delay merging this, even after approval.

dyastremsky:
> Great work! Can we add parameters to the test case for session rate and session concurrency being passed in without the num-sessions? That should work in the case of payload file, I think.

Done. Added these cases.

debermudez (approved):

PR #265, opened 2025-01-31
Add the ability to format output using jinja2 template

fy-nv:
With this change, we can format how the input data is converted into payloads using the `--output-template` option in the CLI. There are two ways of using this feature:

1. Use one of the named templates (e.g. `nv-embedqa`).
2. Provide a [Jinja](https://jinja.palletsprojects.com/en/stable/) template for formatting the payloads. See below for an example of how the conversion works.

### Usage 1: Named template

```bash
genai-perf profile --verbose --input-file /tmp/data.jsonl --concurrency 5 --artifact-dir results/genai-perf/run_20250205-234426 --model MY_MODEL --url localhost:8001 --batch-size 5 --extra-inputs output-template:nv-embedqa

# translates into

perf_analyzer -m MY_MODEL --async --input-data results/genai-perf/run_20250205-234426/inputs.json -i grpc --streaming --concurrency-range 5 --service-kind triton -u localhost:8001 --request-count 0 --warmup-request-count 0 --verbose --profile-export-file results/genai-perf/run_20250205-234426/profile_export.json --measurement-interval 10000 --stability-percentage 999
```

### Usage 2: Jinja template

The jinja template below will produce the same payload (i.e. `inputs.json`), and the resolved  `perf_analyzer` command is the same as above
```bash
template='
[
    {% for item in texts %}
        {"text": [{{ item|tojson }}]}{% if not loop.last %},{% endif %}
    {% endfor %}
]
'
```

This is saved in `custom_template.jinja`. Then, the command to run is:

```bash
genai-perf profile --verbose --input-file /tmp/data.jsonl --concurrency 5 --artifact-dir results/genai-perf/run_20250205-234426 --model MY_MODEL --url localhost:8001 --batch-size 5 --extra-inputs custom_template.jinja.
```

For example, if each row of data contains the following:
```
texts = ["example1", "example2"]
```
the formatted payload will look like

```
 [
    {"text: ["example1"]},
    {"text: ["example2"]}
 ]
```

the resulting `inputs.json` will look like
```
{
    "data": [
        {"text: ["example1"]},
        {"text: ["example2"]},
        ...
    ]
}
```

### DIFF
--- a/genai-perf/genai_perf/inputs/converters/template_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/template_converter.py
@@ -1,0 +1,95 @@
+# Copyright 2024-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of NVIDIA CORPORATION nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import json
+from typing import Any, Dict, Optional
+
+from genai_perf.exceptions import GenAIPerfException
+from genai_perf.inputs.converters.base_converter import BaseConverter
+from genai_perf.inputs.inputs_config import InputsConfig
+from genai_perf.inputs.retrievers.generic_dataset import GenericDataset
+
+NAMED_TEMPLATES = {
+    "nv-embedqa": """[
+        {% for item in texts %}
+            {"text": [{{ item|tojson }}]}{% if not loop.last %},{% endif %}
+        {% endfor %}
+    ]""",
+}
+
+
+class TemplateConverter(BaseConverter):
+    """Convert a GenericDataset to a request body.
+
+    The template should render a list of text strings, and return a list of payloads.
+    """
+
+    def resolve_template(self, template_name_or_content: Optional[str]):
+        try:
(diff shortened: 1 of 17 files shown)


genai-perf/genai_perf/inputs/converters/template_converter.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+
+    def _parse_template(self, template: str):
+        try:
+            import jinja2
+        except ImportError:
+            raise ImportError(
+                "Jinja2 is required for template processing. Install it using: pip install jinja2."
+            )
+        return jinja2.Template(template)
```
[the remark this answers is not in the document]
dyastremsky: Fei, are you able to correct the above?

genai-perf/pyproject.toml
```diff
@@ -98,3 +98,6 @@ quiet-level = 3
 # Same as Black.
 line-length = 88
 indent-width = 4
+
+[project.optional-dependencies]
+test = ["jinja2"]
```
dyastremsky, line 103: Nit: Can you add a newline here?

genai-perf/tests/test_exporters/test_json_exporter.py
```diff
@@ -227,6 +227,7 @@ class TestJsonExporter:
           "u": null,
           "num_dataset_entries": 100,
           "num_prefix_prompts": 0,
+          "output_template": null,
```
dyastremsky, line 230: Nit: For the Triton Tools team, we may not want to export this template. I forget where in the code we exclude certain fields from being exported, but we might want to add this as an exception later.

fy-nv: Done. 
https://github.com/triton-inference-server/perf_analyzer/pull/265/files#diff-cb9938a41942e737251b1ad6af53c5f58a48c84ca78a5b3ffdfdb290a2fe97a3R69

genai-perf/pyproject.toml
```diff
@@ -47,22 +47,23 @@ maintainers = []
 keywords = []
 requires-python = ">=3.10,<4"
 dependencies = [
-  "tritonclient",
+  "fastparquet",
```
dyastremsky, line 50: Alphabetized these and added Jinja2 as a requirement.

genai-perf/tests/test_converters/test_tensorrtllm_engine_converter.py
```diff
@@ -109,30 +106,34 @@ def test_convert_with_chat_template(self):
...
-                    assert "data" in result
-                    assert isinstance(result["data"], list)
-                    assert len(result["data"]) == 2
-
-                    for payload in result["data"]:
-                        assert "input_ids" in payload
-                        assert "content" in payload["input_ids"]
-                        assert payload["input_ids"]["content"] == [101, 202, 303]
+
+        result = trtllm_engine_converter.convert(generic_dataset, config)
```
dyastremsky, line 110: Updated the test here to use Jinja2 rather than mock it.

genai-perf/genai_perf/inputs/converters/template_converter.py
```diff
@@ -0,0 +1,111 @@
...
+                    "Template does not render a list of strings to a list of items. "
+                    f"For example, {test_texts} is rendered into {payloads}."
+                )
+        except Exception as e:
+            raise GenAIPerfException(e)
+
+    def convert(
+        self, generic_dataset: GenericDataset, config: InputsConfig
+    ) -> Dict[Any, Any]:
+        output_template = config.extra_inputs.get("output_template")
```
debermudez, line 101: same as above, why output template? I think output of the server when i hear output template, and think it might be confusing.

dyastremsky: This is a good catch. I was using the original PR terminology. I think "payload template" might work. So that it's clear what the template is for and it won't class with other templates we use in the future.

debermudez (approved):
Amazing! Thank you both for the work here.

genai-perf/genai_perf/parser.py
```diff
@@ -100,6 +97,7 @@ class EndpointConfig:
         "v2/models/{MODEL_NAME}/generate", "triton", ic.OutputFormat.TRITON_GENERATE
     ),
     "kserve": EndpointConfig(None, "triton", ic.OutputFormat.TENSORRTLLM),
+    "template": EndpointConfig(None, "template", ic.OutputFormat.TEMPLATE),
```
fy-nv, line 100: I am not sure how this is intended to be used (I tried using this before but I wasn't successful).

I think setting `--service-kind=triton` and `--endpoint-type=template` will result in an error due to 

```
    if endpoint_config.service_kind != args.service_kind:
        parser.error(
            f"Invalid endpoint-type '{args.endpoint_type}' for service-kind '{args.service_kind}'."
        )
```
https://github.com/triton-inference-server/perf_analyzer/pull/265/files#diff-531d67ceb8ef974e32746d3e8b7b2cd958bac761568f74220ef436b018cacf99R187

PR #271, opened 2025-02-05
Miscellaneous fixes for absolute timestamp/ fixed schedule mode

lkomali:
This PR does some miscellaneous fixes
1. `concurrency`, `request-rate`, `request-count` and `warmup-request-count` are not passed when using this mode.
2. Unit test for method that payload input is not used with incompatible args mentioned above (covering valid and invalid cases)
3. `request-count`, `warmup-request-count`, `meaurement-interval`, `stability-percentage` with their default values are not added in PA subcommand.
4. Unit test to check if the above to be skipped args are actually skipped in PA subcommand 
5. Remove `--schedule <list of timestamps>` in PA subcommand
6. Add `--fixed-schedule` to PA subcommand as this is how it will be supported in PA from now
7. Unit test to check `--fixed-schedule` is added to PA subcommand when using payload input
8. Exclude `hash_ids` from being added to optional_data
9. Break `test_inferred_prompt_source` unit test in `test_cli.py` into two unit tests (valid and invalid case)
10. Update `input-file` section in `README`
11. Any miscellaneous changes

### DIFF
--- a/genai-perf/README.md
+++ b/genai-perf/README.md
@@ -475,11 +475,13 @@ Add a custom header to the requests. Headers must be specified as
 (default: `None`)
 
 ##### `--input-file <path>`
-The input file or directory for profiling. Each line should be a JSON object with a `text` or `image` field
-in JSONL format. Example: `{"text": "Your prompt here"}`. To use synthetic files, prefix with `synthetic:` followed by
-a comma-separated list of filenames without extensions (Example: `synthetic:queries,passages`).
-To use a payload file with a fixed schedule workload, prefix with `payload:` followed by the filename (Example: `payload:input.jsonl`).
-(default: `None`)
+The input file or directory for profiling. Each line should be a JSON object
+with a `text` or `image` field in JSONL format. Example:
+`{"text": "Your prompt here"}`. To use synthetic files, prefix with
+`synthetic:` followed by a comma-separated list of filenames without
+extensions (Example: `synthetic:queries,passages`). To use a payload file with
+a fixed schedule workload, prefix with `payload:` followed by the filename
+(Example: `payload:input.jsonl`). (default: `None`)
 
 ##### `--num-dataset-entries <int>`
 
--- a/genai-perf/README.md
+++ b/genai-perf/README.md
@@ -484,6 +484,10 @@ multiple files, prefix the path with 'synthetic:', followed by a
 comma-separated list of filenames. The synthetic filenames should not have
 extensions. For example, 'synthetic:queries,passages'.
 (default: `None`)
+To use payload file for fixed schedule workload, prefix the
+path with 'payload:', followed by the filename. The payload filename should not
+have an extension. For example, 'payload:input'.
+
 
 ##### `--num-dataset-entries <int>`
 
--- a/genai-perf/README.md
+++ b/genai-perf/README.md
@@ -475,19 +475,11 @@ Add a custom header to the requests. Headers must be specified as
 (default: `None`)
 
 ##### `--input-file <path>`
-
-The input file or directory containing the content to use for
-profiling. Each line should be a JSON object with a 'text' or 'image' field
-in JSONL format. Example: {\"text\": \"Your prompt here\"}"
-To use synthetic files for a converter that needs
-multiple files, prefix the path with 'synthetic:', followed by a
-comma-separated list of filenames. The synthetic filenames should not have
-extensions. For example, 'synthetic:queries,passages'.
+The input file or directory for profiling. Each line should be a JSON object with a `text` or `image` field
+in JSONL format. Example: `{"text": "Your prompt here"}`. To use synthetic files, prefix with `synthetic:` followed by
+a comma-separated list of filenames without extensions (Example: `synthetic:queries,passages`).
+To use a payload file with a fixed schedule workload, prefix with `payload:` followed by the filename (Example: `payload:input.jsonl`).
 (default: `None`)
-To use payload file for fixed schedule workload, prefix the
-path with 'payload:', followed by the filename. The payload filename should not
-have an extension. For example, 'payload:input'.
-
 
 ##### `--num-dataset-entries <int>`
 
--- a/genai-perf/genai_perf/parser.py
+++ b/genai-perf/genai_perf/parser.py
@@ -26,6 +26,7 @@
 
 
 import argparse
+import json
 import os
 import sys
 from dataclasses import dataclass
@@ -511,19 +512,22 @@ def _infer_prompt_source(args: argparse.Namespace) -> argparse.Namespace:
             logger.debug(
                 f"Input source is synthetic data: {args.synthetic_input_files}"
             )
+
         elif input_file_str.startswith("payload:"):
             args.prompt_source = ic.PromptSource.PAYLOAD
-            input_file_str = input_file_str.split(":", 1)[1]
-            if not input_file_str:
-                raise ValueError(
-                    f"Invalid payload input: '{input_file_str}' is missing the file path"
-                )
-            args.payload_input_file = Path(f"{input_file_str}.jsonl")
-            if not args.payload_input_file.is_file():
-                raise ValueError(f"'{args.payload_input_file}' is not a valid file")
+            payload_file = Path(input_file_str.split(":", 1)[1])
+            if not payload_file:
+                raise ValueError("Invalid file path: Path is None or empty.")
+
+            if not payload_file.is_file():
+                raise ValueError(f"File not found: {payload_file}")
+
+            args.payload_input_file = payload_file
+
             logger.debug(
                 f"Input source is a payload file with timing information in the following path: {args.payload_input_file}"
             )
+
         else:
             args.prompt_source = ic.PromptSource.FILE
             logger.debug(f"Input source is the following path: {args.input_file}")
@@ -546,7 +550,7 @@ def _convert_str_to_enum_entry(args, option, enum):
 
 
 def file_or_directory(value: str) -> Path:
-    if value.startswith("synthetic:") or value.startswith("payload:"):
+    if value.startswith("synthetic:"):
         return Path(value)
     else:
         path = Path(value)
@@ -828,8 +832,8 @@ def _add_input_args(parser):
         "should not have extensions. For example, "
         "'synthetic:queries,passages'. For payload data, prefix the path with 'payload:', "
         "followed by a JSON string representing a payload object. The payload should "
-        "contain fields such as 'timestamp', 'input_length', 'output_length', "
-        "and you can optionally add 'text_input', 'session_id', 'hash_ids', and 'priority'. "
+        "contain a 'timestamp' field "
+        "and you can optionally add 'input_length', 'output_length','text_input', 'session_id', 'hash_ids', and 'priority'. "
         'Example: \'payload:{"timestamp": 123.45, "input_length": 10, "output_length": 12, '
         '"session_id": 1, "priority": 5, "text_input": "Your prompt here"}\'.',
     )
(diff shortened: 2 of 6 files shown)


genai-perf/README.md
```diff
@@ -478,6 +478,10 @@ multiple files, prefix the path with 'synthetic:', followed by a
 comma-separated list of filenames. The synthetic filenames should not have
 extensions. For example, 'synthetic:queries,passages'.
 (default: `None`)
+To use payload file for fixed schedule workload, prefix the
```
dyastremsky, line 481:
```suggestion
To use a payload file with a fixed schedule workload, prefix the
```

genai-perf/README.md
```diff
@@ -478,6 +478,10 @@ multiple files, prefix the path with 'synthetic:', followed by a
...
+path with 'payload:', followed by the filename. The payload filename should not
```
dyastremsky, line 482: Why should it not have an extension? Is there a way to auto-detect the extension and remove it programmatically? To make usability better.

genai-perf/genai_perf/wrapper.py
```diff
@@ -138,10 +121,10 @@ def build_cmd(args: Namespace, extra_args: Optional[List[str]] = None) -> List[s
 
         if args.prompt_source == PromptSource.PAYLOAD:
             skip_args += [
-                "request_count",
-                "measurement_interval",
                 "stability_percentage",
                 "warmup_request_count",
+                "request_count",
```
dyastremsky, line 126: Nit: Can you alphabetize this?

genai-perf/genai_perf/wrapper.py
```diff
@@ -155,8 +138,8 @@ def build_cmd(args: Namespace, extra_args: Optional[List[str]] = None) -> List[s
             f"{args.artifact_dir / DEFAULT_INPUT_DATA_JSON}",
         ]
         cmd += Profiler.add_protocol_args(args)
-        cmd += Profiler.add_inference_load_args(args)
         cmd += Profiler.add_payload_args(args)
+        cmd += Profiler.add_inference_load_args(args)
```
dyastremsky, line 142: Is this order dependent? If so, please see if you can design a solution that is not order-dependent. That's a bug waiting to happen. Testing should catch it, but I would expect all of these functions to be able to be called in any order.

Also, as a nit, it'd be nice if these were alphabetical or ordered in some other clear way.

genai-perf/tests/test_wrapper.py
```diff
@@ -185,3 +187,47 @@ def test_headers_passed_correctly(
...
+        mocker.patch.object(Path, "is_file", return_value=True)
+        mocker.patch("genai_perf.utils.remove_file")
+        monkeypatch.setattr("sys.argv", args)
+        args, extra_args = parser.parse_args()
+        cmd = Profiler.build_cmd(args, extra_args)
+        cmd_string = " ".join(cmd)
+
+        assert "--fixed-schedule" in cmd_string
+
+    def test_skipped_args_for_payload_input_not_in_cmd(self, monkeypatch, mocker):
```
dyastremsky, line 209: Nit: I'd rename this to start with test_payload_... that helps logically order the tests during execution and makes it easier to find the right test organizationally. You want to start with the biggest bucket/category, then move towards the smallest one. test_skipped_args won't bucket this test with similar tests... we'd want this test to run and be found with other payload tests, I'd think. They should all start with "test_payload...".

Let me know if you think differently. This is subjective, but I think the naming could be helpful for reading the code and unit tests.

dyastremsky (commented):
Great work! Left a few comments and nits.

genai-perf/README.md
lkomali: the code is appending `.jsonl` extension to the filename. As far as I understand, the requirement is the file should be in `jsonl` format.  So I think, that's why if the user gives a filename, .jsonl is being appended in `infer_prompt_source` method in `parser.py`.
Let me know if we should allow user specifying the extension or I should update README to make sure users create jsonl files only.

genai-perf/genai_perf/wrapper.py
lkomali: No, it's not order dependent. While modifying the order got changed.

genai-perf/tests/test_wrapper.py
lkomali: I think that sounds good. I will rename it to start with test_payload. Also, I think it's easier for anyone seeing the code to understand that these set of tests are all for `payload` input.

genai-perf/README.md
dyastremsky: I think the description saying each lines should be a JSON object makes it a JSONL file by definition, assuming we don't rely on the JSONL file extension, which it sounds like we do. I don't think we should... but the quickest solutio would be specifying that the input file(s) is/are JSONL files.

Also, this should be before the default arg value. Generally, I think we need to update this description. It's too long. Do you or @debermudez have thoughts on this? I think it was expedient to overload it once with synthetic, but now we have two different possible prefixes and a very long description. We should shorten the description and redirect to documentation somewhere else that's clearer on how to use these.

genai-perf/genai_perf/parser.py
```diff
@@ -267,6 +273,8 @@ def _check_load_manager_args(args: argparse.Namespace) -> argparse.Namespace:
     """
     Check inference load args
     """
+    if args.prompt_source == ic.PromptSource.PAYLOAD:
```
lkomali, line 276: Since I'm checking here if the input is payload, `concurrency` is not set to any value. So, artifact_dir name will not have `concurrency` or `request-rate` in it because both of them are None.
Like example artifact_dir is `llama2-openai-chat`

dyastremsky: Sounds good.

genai-perf/README.md
dyastremsky: I'll review this. When you're rebasing, could you try to find ways to squash and use merge commits where possible? Otherwise I need to review the file from scratch, since I can't just see the lines that were modified in the latest commit(s).

genai-perf/README.md
```diff
@@ -469,14 +469,10 @@ Add a custom header to the requests. Headers must be specified as
...
-profiling. Each line should be a JSON object with a 'text' or 'image' field
-in JSONL format. Example: {\"text\": \"Your prompt here\"}"
-To use synthetic files for a converter that needs
-multiple files, prefix the path with 'synthetic:', followed by a
-comma-separated list of filenames. The synthetic filenames should not have
-extensions. For example, 'synthetic:queries,passages'.
+The input file or directory for profiling. Each line should be a JSON object with a `text` or `image` field
+in JSONL format. Example: `{"text": "Your prompt here"}`. To use synthetic files, prefix with `synthetic:` followed by
+a comma-separated list of filenames without extensions (Example: `synthetic:queries,passages`).
+To use a payload file with a fixed schedule workload, prefix with `payload:` followed by the filename (Example: `payload:input.jsonl`).
```
dyastremsky, line 475: Again, I'd love for you and @debermudez to chat a bit about this design. My gut says that we should move all this overloading to a tutorial somewhere. Like just mention the base case, but then have tutorials on how to use synthetic and payload for those specific use cases. I appreciate you shortening this, but I think we need it to only have the general case to be simpler for a casual user.

dyastremsky (commented):
I started to review then noticed this pull request got a lot bigger. Was that intentional or did some other changes get pulled into it?

I'll wait to review until you confirm it's ready for review.

genai-perf/README.md
matthewkotila, line 1: (responding in a file so a thread can be used)

> @dyastremsky I started to review then noticed this pull request got a lot bigger. Was that intentional or did some other changes get pulled into it?
> 
> I'll wait to review until you confirm it's ready for review.

This PR's compare branch ([lkomali-request-timing-miscellaneous-fix](https://github.com/triton-inference-server/perf_analyzer/tree/lkomali-request-timing-miscellaneous-fix)) needs to be rebased onto its base branch ([feature-request-level-timing](https://github.com/triton-inference-server/perf_analyzer/tree/feature-request-level-timing)).

genai-perf/README.md
```diff
@@ -469,14 +469,10 @@ Add a custom header to the requests. Headers must be specified as
...
 ##### `--input-file <path>`
-
-The input file or directory containing the content to use for
-profiling. Each line should be a JSON object with a 'text' or 'image' field
-in JSONL format. Example: {\"text\": \"Your prompt here\"}"
-To use synthetic files for a converter that needs
-multiple files, prefix the path with 'synthetic:', followed by a
-comma-separated list of filenames. The synthetic filenames should not have
-extensions. For example, 'synthetic:queries,passages'.
+The input file or directory for profiling. Each line should be a JSON object with a `text` or `image` field
```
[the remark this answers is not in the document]
lkomali: I have to check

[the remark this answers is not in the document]
lkomali: For now I have formatted to have <=80 characters per line. 
I will shorten the description as per your suggestion below.

genai-perf/README.md
debermudez: This has always felt like a not intuitive way to pass these args. I have never really felt this approach was ergonomic and I think we are just piling on here when we should take a step back and redesign the way we pass these arguments. 
 I think this goes away entirely when we swap to using a configuration file because it is more flexible but I need to review it. I think our time is better spent getting this feature done and revisiting after the configuration file introduction instead of a new document that is not going to be valid in 4-6 weeks. 
That is my personal take but im open to other approaches.

dyastremsky: Are we no longer going to have arg descriptions with the configuration file approach? And is that going to be the only way of feeding args now?

debermudez: That is a good point.
So the ask is to highlight the various options and link to a tutorial doc rather than having the explanation here?

dyastremsky: It was. It's not a must, just that I'd rather the special options be highlighted in the associated tutorials (e.g. for synthetic payloads, for timing strategy files, etc.). Otherwise, it'll be confusing for the user.

Having these here is helpful as a developer to remember these options exist, but perhaps we switch that to a comment. Or we ultimately clean it up to be more commonsense for GA, e.g. having two arguments: --input-file-type and --input-file.

genai-perf/genai_perf/parser.py
```diff
@@ -818,8 +831,8 @@ def _add_input_args(parser):
         "should not have extensions. For example, "
         "'synthetic:queries,passages'. For payload data, prefix the path with 'payload:', "
         "followed by a JSON string representing a payload object. The payload should "
-        "contain fields such as 'timestamp', 'input_length', 'output_length', "
-        "and you can optionally add 'text_input', 'session_id', 'hash_ids', and 'priority'. "
+        "contain a 'timestamp' field "
```
debermudez, line 834: Yea okay. Seeing this typed out now, this is getting very verbose for a single option.
We likely want to link out to a tutorial here. @dyastremsky for viz

debermudez: @lkomali i think we make this a separate ticket
lets get this PR in, but we can add the doc in the next PR.

genai-perf/genai_perf/wrapper.py
```diff
@@ -138,8 +121,8 @@ def build_cmd(args: Namespace, extra_args: Optional[List[str]] = None) -> List[s
 
         if args.prompt_source == PromptSource.PAYLOAD:
             skip_args += [
-                "request_count",
                 "measurement_interval",
+                "request_count",
```
debermudez, line 125: This should never happen right?
This is an incompatible arg for this mode and would have raised a parser error no?

lkomali: If we don't add it to skips_args, request_count 0 is being appended. 
If someone passes a value for it, GAP will throw an error. Even if somebody doesn't pass this option, request-count 0 is added to the subcommand. The same with warmup-request-count too. That's why, I had to add them to skip_args section.

debermudez: Got it. 
That makes sense to me.

debermudez (approved):

PR #273, opened 2025-02-05
Add goodput CLI doc which was missing from the main Readme for Genai-Perf.

ganeshku1:
Add goodput CLI doc which was missing from the main Readme for Genai-Perf.

### DIFF
--- a/genai-perf/README.md
+++ b/genai-perf/README.md
@@ -617,8 +617,8 @@ An option to enable verbose mode. (default: `False`)
 
 An option to print the version and exit.
 
-##### `--goodput <key:value>`
-##### `--g <key:value>`
+##### `-g <list>`
+##### `--goodput <list>`
 
 An option to provide constraints in order to compute goodput. Specify goodput
 constraints as 'key:value' pairs, where the key is a valid metric name, and the
--- a/genai-perf/README.md
+++ b/genai-perf/README.md
@@ -617,6 +617,16 @@ An option to enable verbose mode. (default: `False`)
 
 An option to print the version and exit.
 
+##### `--goodput <key:value>`
+##### `--g <key:value>`
+
+An option to provide constraints in order to compute goodput. Specify goodput
+constraints as 'key:value' pairs, where the key is a valid metric name, and the
+value is a number representing either milliseconds or a throughput value per
+second. For example, 'request_latency:300' or
+'output_token_throughput_per_request:600'. Multiple key:value pairs can be
+provided, separated by spaces. (default: `None`)
+
 </br>
 
 <!--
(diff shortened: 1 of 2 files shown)


dyastremsky (commented):
Thanks for catching this missing arg and making this change!

These help messages should match the parser.py [messages](https://github.com/triton-inference-server/perf_analyzer/blob/main/genai-perf/genai_perf/parser.py#L758-L769) exactly. Also, please specify the default value of None. And the value for this flag should be <str>, not options.

Also, any changes made here need to also be made to README_template.md. Finally, please break this into multiple lines so that no single line is greater than 80 characters.

ganeshku1:
> Thanks for catching this missing arg and making this change!
> 
> These help messages should match the parser.py [messages](https://github.com/triton-inference-server/perf_analyzer/blob/main/genai-perf/genai_perf/parser.py#L758-L769) exactly. Also, please specify the default value of None. And the value for this flag should be list, not options. You can look to other --model as a reference, since that flag functions similarly.
> 
> Also, any changes made here need to also be made to README_template.md. Finally, please break this into multiple lines so that no single line is greater than 80 characters.

@dyastremsky  updated the the PR with this feedback included.
When you get a chance please feel free to review.

genai-perf/README.md
```diff
@@ -617,6 +617,16 @@ An option to enable verbose mode. (default: `False`)
 
 An option to print the version and exit.
 
+##### `--goodput <key:value>`
```
dyastremsky, line 620: This deviates from the other args, like --model and --server-metrics-url are documented. I think this should be <list>.

genai-perf/README.md
```diff
@@ -617,6 +617,16 @@ An option to enable verbose mode. (default: `False`)
...
+##### `--g <key:value>`
+
+An option to provide constraints in order to compute goodput. Specify goodput
+constraints as 'key:value' pairs, where the key is a valid metric name, and the
+value is a number representing either milliseconds or a throughput value per
+second. For example, 'request_latency:300' or
+'output_token_throughput_per_request:600'. Multiple key:value pairs can be
+provided, separated by spaces. (default: `None`)
+
```
dyastremsky, line 629: Can you make the same changes to README_template? They both need to be updated.

dyastremsky (commented):
Great updates! A couple of last comments.

genai-perf/README.md
ganeshku1: Sorry missed this one, updated the README_templeate

genai-perf/README.md
```diff
@@ -617,6 +617,16 @@ An option to enable verbose mode. (default: `False`)
 
 An option to print the version and exit.
 
+##### `--goodput <list>`
+##### `--g <list>`
```
dyastremsky, line 621: Two more changes I missed.

1. Can you reorder this? It looks like the README always lists the args alphabetically, including the single-letter flags being first (i.e. `-g` then `--goodput`).
2. There should only be one flag in front of the g (i.e. `-g` instead of `--g`).

Please make sure to make identical changes in README and README_template. Thanks again for making these changes!

ganeshku1: Done, thanks for catching this fixed.

dyastremsky (approved):

ganeshku1:
> Excellent work, Ganesh!

Thank you @dyastremsky

PR #275, opened 2025-02-05
Support Numpy 2

dyastremsky:


debermudez (approved):

dyastremsky:
Thanks for approving so quickly! CI is running.

nv-hwoo:
Just to double check: I  think we were restricting the numpy version to 1.x due to some breaking changes in 2.0 API (https://github.com/triton-inference-server/client/pull/690) but that was mainly affecting triton python client and not genai-perf, correct?

debermudez:
> Just to double check: I think we were restricting the numpy version to 1.x due to some breaking changes in 2.0 API ([triton-inference-server/client#690](https://github.com/triton-inference-server/client/pull/690)) but that was mainly affecting triton python client and not genai-perf, correct?

I believe we restricted it because we didnt have the bandwidth to verify the API change and no one was directly asking for it.
With the new ask from vllm and internal folks plus the additional testing around our use of the library, it is a good time to do the update.

zhouyuan:
@dyastremsky thank you so much for the fix. Do you happen to know if there will a new release to include this change recently?

dyastremsky:
> @dyastremsky thank you so much for the fix. Do you happen to know if there will a new release to include this change recently?

I think this should make it into 25.02.

rgommers:
> I think this should make it into 25.02.

Can I ask when this is planned and where to look for releases? https://pypi.org/project/genai-perf is at 0.0.10, which doesn't give the impression that 25.02 is next.

nv-hwoo:
@rgommers you can expect it to be part of 0.0.11 version, which should be released soon.

PR #276, opened 2025-02-06
Add dynamic GRPC converter

dyastremsky:
Support dynamic GRPC benchmarking in GenAI-Perf with a new converter. This is the proof of concept.

See below:
image


genai-perf/genai_perf/wrapper.py
```diff
@@ -113,17 +113,18 @@ def build_cmd(args: Namespace, extra_args: Optional[List[str]] = None) -> List[s
 
         utils.remove_file(args.profile_export_file)
 
+        # TODO: Temp, remove below change
```
nv-hwoo, line 116: just a placeholder comment (to not forget to remove)

genai-perf/genai_perf/inputs/converters/dynamic_grpc_converter.py
```diff
@@ -0,0 +1,64 @@
...
+    ) -> Dict[Any, Any]:
+        request_body: Dict[str, Any] = {"data": []}
+
+        for file_data in generic_dataset.files_data.values():
+            for index, row in enumerate(file_data.rows):
+                model_name = self._select_model_name(config, index)
+                payload = {
+                    "model": model_name,
+                }
+                payload["ipc_stream"] = row.texts[0]
```
nv-hwoo, line 60: We actually don't need to specify any field other than `"ipc_stream"` :) 
```suggestion
                payload = {"ipc_stream": row.texts[0]}
```

genai-perf/genai_perf/parser.py
```diff
@@ -168,16 +166,16 @@ def _check_conditional_args(
     """
 
     # Endpoint and output format checks
-    if args.service_kind == "openai":
+    if args.service_kind in ["openai"]:
```
nv-hwoo, line 169: I think you forgot to undo this change :)

genai-perf/genai_perf/wrapper.py
```diff
@@ -113,17 +113,18 @@ def build_cmd(args: Namespace, extra_args: Optional[List[str]] = None) -> List[s
...
         cmd = [
-            f"perf_analyzer",
+            f"build/install/bin/perf_analyzer",
             f"-m",
             f"{args.formatted_model_name}",
```
nv-hwoo, line 120: since we ripped out `--async` from this default command list, can we also rip the `-m` out as well? Since dynamic grpc doesn't use this flag.

genai-perf/genai_perf/parser.py
dyastremsky: Good catch, thanks!

genai-perf/genai_perf/wrapper.py
dyastremsky: Sure, I'll add this. We still require a model be passed in anyway though, so it won't make much of a difference. If we want to remove the model name being required, that's a bigger change. I think we use it in so many places (artifacts, visualizations, etc.) that we'd want its own PR for making model name optional. I don't think it's worth the effort unless it becomes optional for multiple use cases.

genai-perf/genai_perf/wrapper.py
dyastremsky: Thanks! Removed.

dyastremsky:
Feedback addressed. I also re-ran the command in the PR description to ensure it still works.

nv-hwoo:
Closing as we opened a new PR: https://github.com/triton-inference-server/perf_analyzer/pull/287

PR #277, opened 2025-02-06
fix incorrect token length for long isl

tedzhouhk:
if isl is long, the corpus needs to be repeated >1 times.


tedzhouhk:
closing cause current branch has 1.8M tokens.

PR #280, opened 2025-02-07
Add support for fixed schedule inference load mode

matthewkotila:


.pre-commit-config.yaml
```diff
@@ -42,7 +42,7 @@ repos:
     args: [--max-line-length=88, --select=C,E,F,W,B,B950, --extend-ignore = E203,E501]
     types_or: [python, cython]
 - repo: https://github.com/pre-commit/mirrors-clang-format
-  rev: v16.0.5
+  rev: v18.1.3
```
matthewkotila, line 45: This is the version of clang-format that gets installed via `apt install clang-format` on ubuntu 24.04.

src/CMakeLists.txt
```diff
@@ -204,7 +204,7 @@ set(
   session_concurrency/response_json_utils.h
   session_concurrency/session_concurrency_manager.h
   rapidjson_utils.h
-  session_concurrency_mode.h
+  inference_load_mode.h
```
matthewkotila, line 207: Instead of each inference load mode having its own bool stating whether it is enabled or not, I just made one enum to hold which of the inference load modes was chosen via its CLI option.

src/command_line_parser.cc
```diff
@@ -927,11 +928,12 @@ CLParser::ParseCommandLine(int argc, char** argv)
       {"endpoint", required_argument, 0, long_option_idx_base + 61},
       {"request-count", required_argument, 0, long_option_idx_base + 62},
       {"warmup-request-count", required_argument, 0, long_option_idx_base + 63},
-      {"schedule", required_argument, 0, long_option_idx_base + 64},
+      {"fixed-schedule", no_argument, 0, long_option_idx_base + 64},
```
matthewkotila, line 931: `--schedule=<n>,<n>,<n>,...` is replaced by `--fixed-schedule` (no values).

It signals PA to look for `timestamp` key in input data JSON.

src/command_line_parser.cc
```diff
@@ -927,11 +928,12 @@ CLParser::ParseCommandLine(int argc, char** argv)
...
       {"session-concurrency", required_argument, 0, long_option_idx_base + 65},
       {0, 0, 0, 0}};
 
   // Parse commandline...
+  bool using_deprecated_concurrency{false};
```
matthewkotila, line 936: Did some unrelated cleanup of deprecated options.

src/command_line_parser.cc
```diff
@@ -1016,7 +1022,12 @@ CLParser::ParseCommandLine(int argc, char** argv)
           break;
         }
         case long_option_idx_base + 7: {
-          params_->using_concurrency_range = true;
+          if (params_->inference_load_mode != InferenceLoadMode::None) {
+            Usage(
+                "Cannot use both " + to_string(params_->inference_load_mode) +
+                " and Concurrency inference load modes.");
+          }
+          params_->inference_load_mode = InferenceLoadMode::Concurrency;
```
matthewkotila, line 1030: Inference load modes now check at the time they are detected whether more than one was used, and errors out if so.

src/command_line_parser.cc
```diff
@@ -1777,6 +1830,11 @@ CLParser::ParseCommandLine(int argc, char** argv)
     }
   }
 
+  if (params_->inference_load_mode == InferenceLoadMode::None ||
+      using_deprecated_concurrency) {
+    params_->inference_load_mode = InferenceLoadMode::Concurrency;
+  }
+
```
matthewkotila, line 1837: This enforces the usual default mode of concurrency.

params_->inference_load_mode needed to hold `None` mode to detect repeated usage, so we set actual default here.

src/command_line_parser.cc
```diff
@@ -1791,16 +1849,32 @@ CLParser::ParseCommandLine(int argc, char** argv)
   }
 
   // Overriding the max_threads default for request_rate search
-  if (!params_->max_threads_specified && params_->targeting_concurrency()) {
+  if (!params_->max_threads_specified &&
+      params_->inference_load_mode == InferenceLoadMode::Concurrency) {
```
matthewkotila, line 1853: Anywhere that `targeting_concurrency()` or `using_request_rate_range` or `using_custom_intervals`, etc, was used now just checks the `inference_load_mode` enum value.

src/command_line_parser.cc
```diff
@@ -1791,16 +1849,32 @@ CLParser::ParseCommandLine(int argc, char** argv)
...
+        throw std::runtime_error(
+            "'" + input_data_path +
+            "' is not a file. When using `--fixed-schedule`, all paths "
+            "passed to --input-data must be files.");
+      }
+    }
+
+    params_->request_count = pa::DataLoader::GetDatasetSize(params_->user_data);
+  }
+
```
matthewkotila, line 1877: Previously, request count was set to the size of `--schedule` list, which was known at command line argument parsing time (here).

Now, it is based on the number of payloads in the user-provided input data JSON.

Ideally, this would be parsed closer to where the input data is used, but because of the current design of PA, it was important to know the request count earlier than that point, thus this unideal code to scan the input data to determine the request count.

This is not the ideal solution but making the proper solution would have extended the timeline much farther than I or the customers would have wanted.

src/command_line_parser.cc
```diff
@@ -1815,8 +1889,7 @@ CLParser::VerifyOptions()
   if (params_->model_name.empty()) {
     Usage("Failed to parse -m (model name). The value must be specified.");
   }
-  if (params_->concurrency_range.start <= 0 ||
-      params_->concurrent_request_count < 0) {
```
matthewkotila, line 1819: `concurrent_request_count` was part of the deprecation options which were cleaned up and now instead use the `concurrency_range` data structure to hold state.

src/command_line_parser.cc
```diff
@@ -1861,53 +1934,26 @@ CLParser::VerifyOptions()
...
-  if (params_->using_request_rate_range && params_->using_old_options) {
-    Usage("Cannot use concurrency options with --request-rate-range.");
-  }
-
-  if (!params_->schedule.empty() && params_->warmup_request_count > 0) {
-    Usage("Cannot use warmup options with --schedule");
-  }
-
-  if (!params_->schedule.empty() && params_->using_concurrency_range) {
-    Usage("Cannot use concurrency options with --schedule");
```
matthewkotila, line 1882: Getting rid of checks specific to deprecated options since they leverage other existing checks now, same for the inference load mode checks here--they leverage other existing checks now.

src/command_line_parser.cc
```diff
@@ -1932,24 +1978,15 @@ CLParser::VerifyOptions()
...
-  if (params_->using_custom_intervals && params_->using_old_options) {
-    Usage("Cannot use deprecated options with --request-intervals.");
-  }
-
-  if ((params_->using_custom_intervals) &&
-      (params_->using_request_rate_range || params_->using_concurrency_range)) {
-    Usage(
-        "Cannot use --concurrency-range or --request-rate-range "
-        "along with --request-intervals.");
-  }
```
matthewkotila, line 1950: Checked elsewhere.

src/command_line_parser.cc
```diff
@@ -2124,7 +2161,12 @@ CLParser::VerifyOptions()
     }
   }
 
-  if (params_->session_concurrency_mode == SessionConcurrencyMode::Enabled &&
+  if (params_->inference_load_mode == InferenceLoadMode::FixedSchedule &&
+      params_->kind != cb::BackendKind::OPENAI) {
+    Usage("Fixed schedule mode is only supported with OpenAI service kind.");
+  }
```
matthewkotila, line 2167: Right now, I've limited `--fixed-schedule` mode to `--service-kind=openai`, but it might be able to work outside of that. Would need more testing. Customer requirement currently is only for `openai`.

src/command_line_parser.h
```diff
@@ -57,7 +59,6 @@ struct PerfAnalyzerParameters {
   std::unordered_map<std::string, std::vector<int64_t>> input_shapes;
   std::vector<cb::ModelIdentifier> bls_composing_models;
   uint64_t measurement_window_ms = 5000;
-  bool using_concurrency_range = false;
```
matthewkotila, line 60: This was previously used for the deprecated options.

Deprecated options now leverage existing options' state.

src/command_line_parser.h
```diff
@@ -128,17 +123,6 @@ struct PerfAnalyzerParameters {
...
-  //
-  bool targeting_concurrency() const
-  {
-    return (
-        using_concurrency_range || using_old_options ||
-        !(using_request_rate_range || using_custom_intervals ||
-          is_using_periodic_concurrency_mode ||
-          session_concurrency_mode == SessionConcurrencyMode::Enabled));
-  }
-
```
matthewkotila, line 141: Unneeded now, code can just check inference_load_mode value.

src/command_line_parser.h
```diff
@@ -155,15 +139,11 @@ struct PerfAnalyzerParameters {
   // The profile export file path.
   std::string profile_export_file{""};
 
-  bool is_using_periodic_concurrency_mode{false};
   Range<uint64_t> periodic_concurrency_range{1, 1, 1};
   uint64_t request_period{10};
   size_t warmup_request_count{0};
 
-  std::vector<float> schedule{};
```
matthewkotila, line 163: As mentioned, the fixed schedule source of truth is now in the input data JSON `timestamp` fields instead of being passed in via `--schedule=<n>,<n>,...`.

src/custom_request_schedule_manager.cc
```diff
@@ -78,25 +77,74 @@ CustomRequestScheduleManager::GenerateSchedule()
...
+  }
+
+  const size_t dataset_size{data_loader_->GetTotalSteps(0)};
+
+  for (size_t dataset_index{0}; dataset_index < dataset_size; ++dataset_index) {
+    const auto timestamp{GetTimestamp(dataset_index)};
+    schedule.push_back(timestamp);
+  }
+
+  std::sort(schedule.begin(), schedule.end());
```
matthewkotila, line 126: Interestingly, the schedule doesn't _need_ to be supplied in order, so we guarantee it's in order here before divvying up among threads.

src/data_loader.cc
```diff
@@ -451,6 +451,48 @@ DataLoader::ValidateIndexes(int stream_id, int step_id) const
   return cb::Error::Success;
 }
 
+size_t
+DataLoader::GetDatasetSize(const std::vector<std::string>& input_data_paths)
```
matthewkotila, line 455: I don't like this function, as mentioned [here](https://github.com/triton-inference-server/perf_analyzer/pull/280/files#r1946939872), but it was a quick way to scan the input data JSON(s) and pre-check out many requests there will be so that that information can be given to InferenceProfiler class to determine when benchmarking is done.

src/command_line_parser.cc
lkomali: Which option is deprecated? The variable says deprecated concurrency, but why is concurrency deprecated? Sorry trying to get context and understand here.

src/inference_profiler.cc
matthewkotila, line 1: Most changes from this file are from auto-formatter.

src/inference_profiler.cc
```diff
@@ -1420,7 +1415,8 @@ InferenceProfiler::CollectData(
     PerfStatus& summary, uint64_t window_start_ns, uint64_t window_end_ns,
     std::vector<RequestRecord>&& request_records)
 {
-  InferenceLoadMode id{summary.concurrency, summary.request_rate};
+  ProfileDataCollector::InferenceLoadMode id{
```
matthewkotila, line 1418: I made a new enum called `InferenceLoadMode` to hold the mode for when parsing the CLI options happens, but it turns out this identifier was previously used.

The `ProfileDataCollector` class used this for work regarding writing recorded requests and responses to the profile export file.

I just moved that struct with the same name into the `ProfileDataCollector` class since it's related to that.

src/mock_profile_data_exporter.h
matthewkotila, line 1: Every change here related to the moving of the `InferenceLoadMode` struct into the `ProfileDataCollector` class mentioned [here](https://github.com/triton-inference-server/perf_analyzer/pull/280/files#r1947016555).

src/model_parser.cc
```diff
@@ -293,13 +293,20 @@ ModelParser::InitOpenAI(
   response_output->second.datatype_ = "JSON";
   response_output->second.shape_.push_back(1);
 
+  // OpenAI in fixed schedule mode takes a timestamp input
+  if (inference_load_mode == InferenceLoadMode::FixedSchedule) {
+    auto delay_input = inputs_->emplace("timestamp", ModelTensor()).first;
+    delay_input->second.name_ = "timestamp";
+    delay_input->second.datatype_ = "UINT64";
+    delay_input->second.shape_.push_back(1);
+  }
```
matthewkotila, line 302: As mentioned, the new `--fixed-schedule` mode looks in the input data JSON for `timestamp` key to find timestamps.

We have to add the `timestamp` as an expected input in the ModelParser so that the DataLoader can read it in without errors.

src/command_line_parser.cc
lkomali: Oh! is this for '-d'?

src/command_line_parser.cc
lkomali: Is the comment on line 1851 correct?

src/profile_data_collector.h
```diff
@@ -37,47 +37,47 @@
 
 namespace triton { namespace perfanalyzer {
 
-/// Data structure to hold which inference load mode was used for an experiment.
-/// Only one data member will be nonzero, indicating the inference load mode for
-/// a particular experiment.
-struct InferenceLoadMode {
```
matthewkotila, line 43: As mentioned, the `InferenceLoadMode` struct that is used for `ProfileDataCollector` got moved from the `triton::perf_analyzer` namespace into the `ProfileDataCollector` class

So that `triton::perf_analyzer::InferenceLoadMode` could be used for the enum holding what mode the user wants from CLI option parsing (e.g. concurrency, request rate, fixed schedule, etc).

src/request_rate_manager.cc
```diff
@@ -283,7 +283,10 @@ RequestRateManager::ResumeWorkers()
   start_time_ = std::chrono::steady_clock::now();
 
   // Wake up all the threads to begin execution
-  execute_ = true;
+  {
+    std::lock_guard<std::mutex> lock(wake_mutex_);
+    execute_ = true;
+  }
```
matthewkotila, line 289: This just seemed like an uncaught potential race condition to me, when looking at how the [condition variable docs](https://en.cppreference.com/w/cpp/thread/condition_variable) say it should be used.

There are likely other race conditions still present but hopefully this removes one, and CI did pass.

src/test_command_line_parser.cc
matthewkotila, line 1: As mentioned, command line parsing now uses an `inference_load_mode` enum to track what inference load mode the user is requesting, rather than N bools like `using_concurrency_range`, `using_request_rate_range`, etc.

Changes in this file are about using the new enum and also removing usage of old deprecated option state.

The deprecated options aren't gone, but they rely on state from un-deprecated options, now.

lkomali (commented):
Awesome work Matt! Appreciate you for cleaning up a bunch of things. And thank you for adding comments that helped me understand the reason for your changes. It was really easy to follow through the code. Thanks for this work!

src/test_custom_request_schedule_manager.cc
```diff
@@ -37,82 +38,81 @@ namespace triton::perfanalyzer {
...
+    std::vector<std::chrono::nanoseconds> actual_timestamps{};
+
+    for (const auto& iworker : workers_) {
+      const auto worker{std::dynamic_pointer_cast<RequestRateWorker>(iworker)};
+      for (size_t i{0}; i < worker->schedule_->intervals.size(); ++i) {
+        const auto timestamp{worker->GetNextTimestamp()};
+        actual_timestamps.push_back(timestamp);
+      }
     }
-  }
```
matthewkotila, line 74: It seems like there was a bug here (or I just don't understand it).

Old unit test was looping through each worker and grabbing the first timestamp from each other? Seems like that wouldn't check the entire schedule if any worker had more than one piece of the schedule.

The new unit test loops through each worker and grabs all pieces of the schedule that it owns and groups it all and sorts it and compares with expected schedule.

src/test_custom_request_schedule_manager.cc
```diff
@@ -37,82 +38,81 @@ namespace triton::perfanalyzer {
...
+
+  std::vector<std::chrono::milliseconds> schedule{};
 
   const auto& ParameterizeSchedule{[&]() {
+    using std::literals::chrono_literals::operator""ms;
+
     SUBCASE("schedule A")
     {
-      params.schedule = {1.0, 2.0, 3.0, 4.0, 5.0};
+      schedule = {1000ms, 2000ms, 3000ms, 4000ms, 5000ms};
```
matthewkotila, line 90: This mimics the schedule as given by user.

As mentioned, previously, user would give schedule in seconds, now they must give it in milliseconds.

Code still converts to nanoseconds for internal representation, same as before PR.

src/command_line_parser.cc
lkomali: Found answer to my question after I looked through the whole PR.

matthewkotila: No deprecated options have been removed. They're just "simplified" so that after the command line parsing, the rest of the codebase does not know deprecated options exist, because they just assign their values to equivalent existing options that aren't deprecated.

src/command_line_parser.cc
matthewkotila: Seems like that comment is wrong. Changed.

.pre-commit-config.yaml
matthewkotila, line 1: (responding in file so a thread can be used)

> @lkomali: Awesome work Matt! Appreciate you for cleaning up a bunch of things. And thank you for adding comments that helped me understand the reason for your changes. It was really easy to follow through the code. Thanks for this work!
> I had only 1 question. The PR looks good to me.
> Would wait for @dyastremsky review

Thanks Harshini! I'm fine with merging without David's review, but I see him adding reactions to threads, so I'll wait a bit for feedback.

No pressure for deep analysis, @dyastremsky 🙏.

matthewkotila: Have at it 🫡

dyastremsky (approved):
Very clean. I appreciate how much improvement and reduction of technical debt you included in this PR. Sounds like it's not long until we can deprecate long-ago-deprecated features.

PR #282, opened 2025-02-08
Make template work with Triton service-kind

dyastremsky:
The template converter was previously having issues due to not having a valid service-kind that can be input. Fix that by making it work with the Triton service-kind.

### DIFF
--- a/genai-perf/docs/customizable_payloads.md
+++ b/genai-perf/docs/customizable_payloads.md
@@ -32,6 +32,8 @@ With GenAI-Perf, you can customize how input data is formatted into payloads
 using templates. This allows you to define how payloads are
 structured when sent to an inference server.
 
+This feature is only available for the Triton service-kind currently.
+
 These provide less customizability than a
 [customizable frontend](customizable_frontends.md), which is used when more
 endpoint-specific logic is necessary. Templates are used when
@@ -59,6 +61,7 @@ Run the following command:
 genai-perf profile \
   --model MY_MODEL \
   --tokenizer MY_MODEL \
+  --service-kind triton \
   --num-payloads 2 \
   --extra-inputs payload_template:nv-embedqa
 ```
@@ -105,6 +108,7 @@ you can run the command:
 genai-perf profile \
   --model MY_MODEL \
   --tokenizer MY_MODEL \
+  --service-kind triton \
   --num-payloads 2 \
   --extra-inputs payload_template:custom_template.jinja
 ```
--- a/genai-perf/docs/customizable_payloads.md
+++ b/genai-perf/docs/customizable_payloads.md
@@ -62,6 +62,7 @@ genai-perf profile \
   --model MY_MODEL \
   --tokenizer MY_MODEL \
   --service-kind triton \
+  --endpoint-type template \
   --num-payloads 2 \
   --extra-inputs payload_template:nv-embedqa
 ```
@@ -109,6 +110,7 @@ genai-perf profile \
   --model MY_MODEL \
   --tokenizer MY_MODEL \
   --service-kind triton \
+  --endpoint-type template \
   --num-payloads 2 \
   --extra-inputs payload_template:custom_template.jinja
 ```
--- a/genai-perf/genai_perf/parser.py
+++ b/genai-perf/genai_perf/parser.py
@@ -26,7 +26,6 @@
 
 
 import argparse
-import os
 import sys
 from dataclasses import dataclass
 from enum import Enum, auto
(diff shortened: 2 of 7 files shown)


genai-perf/genai_perf/parser.py
```diff
@@ -436,7 +436,7 @@ def _set_artifact_paths(args: argparse.Namespace) -> argparse.Namespace:
         if args.service_kind == "openai":
             name += [f"{args.service_kind}-{args.endpoint_type}"]
         elif args.service_kind == "triton":
-            name += [f"{args.service_kind}-{args.backend.to_lowercase()}"]
+            name += [f"{args.service_kind}-{args.backend}"]
```
dyastremsky, line 439: I don't know why this wasn't causing an issue previously, since this code has been here for months.

Args.backend is a string, so there is no to_lowercase() function. This line returned an error when using the default backend value. Moreover, the enum helper function means all of the strings are already lowercase.

genai-perf/genai_perf/parser.py
```diff
@@ -97,7 +97,7 @@ class EndpointConfig:
         "v2/models/{MODEL_NAME}/generate", "triton", ic.OutputFormat.TRITON_GENERATE
     ),
     "kserve": EndpointConfig(None, "triton", ic.OutputFormat.TENSORRTLLM),
-    "template": EndpointConfig(None, "template", ic.OutputFormat.TEMPLATE),
+    "template": EndpointConfig(None, "triton", ic.OutputFormat.TEMPLATE),
```
fy-nv, line 100: I'd suggest that we simplify the logic [here](https://github.com/triton-inference-server/perf_analyzer/pull/282/files#diff-531d67ceb8ef974e32746d3e8b7b2cd958bac761568f74220ef436b018cacf99R175) because  `endpoint_type` and `service_kind` become somewhat circular; that is, when `endpoint_type` is `None`, then it is dependent on `service_kind`, but later it is [checked again](https://github.com/triton-inference-server/perf_analyzer/pull/282/files#diff-531d67ceb8ef974e32746d3e8b7b2cd958bac761568f74220ef436b018cacf99R187) against the `service_kind` after it's resolved from `_endpoint_type_map`.

Maybe use tuple `(service_kind, endpoint_type)` as lookup key in  `_endpoint_type_map` to avoid confusion?

genai-perf/docs/customizable_payloads.md
```diff
@@ -59,6 +61,7 @@ Run the following command:
 genai-perf profile \
   --model MY_MODEL \
   --tokenizer MY_MODEL \
+  --service-kind triton \
```
fy-nv, line 64: We also need `--endpoint-type template`

fy-nv (changes requested):
After starting a toy Triton server per https://github.com/triton-inference-server/python_backend/tree/main

I was able to test the MR successfully with the following (note that I added `--endpoint-type template`)

```bash
export TRITON_RELEASE=25.01
export GIT_SHA=dyas-template-fix

git clone https://github.com/triton-inference-server/perf_analyzer.git
git checkout ${GIT_SHA}

docker run -ti --network host \
    -v $(pwd)/perf_analyzer:/code/perf_analyzer \
    nvcr.io/nvidia/tritonserver:${TRITON_RELEASE}-py3-sdk /bin/bash

pushd /code/perf_analyzer/genai-perf
pip install -e .


template='
[
    {% for item in texts %}
        {
            "INPUT0": [{{ item|length }}.0, 0.0, 0.0, 0.0],
            "INPUT1": [1.0, 2.0, 3.0, 4.0]
        }{% if not loop.last %},{% endif %}
    {% endfor %}
]
'
echo $template > custom_template.jinja


genai-perf profile \
    --endpoint-type template \
    --url localhost:8001 \
    --model add_sub \
    --extra-inputs payload_template:custom_template.jinja
```

genai-perf/genai_perf/parser.py
dyastremsky: Thanks for this suggestion! We'll do this in a future PR. Service-kind has been deprecated for a while. We intend to delete it soon, which would make this sort of approach straightforward.

genai-perf/docs/customizable_payloads.md
dyastremsky: Good catch. Done!

genai-perf/genai_perf/parser.py
```diff
@@ -199,9 +198,10 @@ def _check_conditional_args(
         if endpoint_config.endpoint:
             args.endpoint = endpoint_config.endpoint.format(MODEL_NAME=model_name)
 
-    if args.service_kind == "triton" and args.endpoint_type == "kserve":
+    if args.service_kind == "triton" and args.endpoint_type in ["kserve", "template"]:
```
dyastremsky, line 201: Fix for this endpoint-type, since this converts the string to be an Enum, which is important to avoid type errors later. Long-term, we'll move away from service-kind.

genai-perf/tests/test_wrapper.py
```diff
@@ -79,8 +79,6 @@ def test_profile_export_filepath(self, monkeypatch, arg, expected_filepath):
             "profile",
             "-m",
             "test_model",
-            "--service-kind",
```
dyastremsky, line 82: Removing some code, since this is a default value and excluding it makes the code less verbose.

fy-nv (approved):
Thanks!

PR #284, opened 2025-02-13
feat: Fixed Schedule

lkomali:
This is the feature PR for fixed schedule mode to replay a trace e.g. mooncake.

User can provide an input payload file as shown below
```
{"timestamp": 0, "text_input": "First message!", input_length": 6755, "output_length": 500}
{"timestamp": 0, "text_input": "Second message!","input_length": 7319, "output_length": 490}
{"timestamp": 0, "text_input": "Third message!","input_length": 7234, "output_length": 794}
```
Calling it a payload file because it makes changes at the payload level.
This file can be provided to GAP using `--input-file payload:<path to file>`

Gap generates the inputs.json which is used by PA to send requests at the exact timestamp specified in the file.

The `"text_input"` is optional and in case it is not provided, the `SyntheticPromptGenerator` will create the prompts.

An example run is as follows
image


.pre-commit-config.yaml
matthewkotila, line 1: How is this PR intending to be reviewed?

I noticed there's one new commit added to the [feature-request-level-timing](https://github.com/triton-inference-server/perf_analyzer/tree/feature-request-level-timing) since my PR (#280) went in. Was that reviewing in a PR? I can't find one for it.

Wouldn't it be better to review that commit separately from a PR for merging the entire feature branch into main?

.pre-commit-config.yaml
```diff
@@ -84,4 +84,4 @@ repos:
     name: Add License
     entry: python tools/add_copyright.py
     language: python
-    require_serial: true
+    require_serial: true
```
lkomali, line 87: I'm not sure what changes and this file is popping up in diff.

.pre-commit-config.yaml
matthewkotila: I mainly ask cuz that commit undid a change I did to the clang format version and idk why haha.

matthewkotila: Hence my whole comment about wanting to review any new changes.

.pre-commit-config.yaml
matthewkotila: The newline at the end of the file was deleted in the latest commit.

.pre-commit-config.yaml
matthewkotila: Oh, looks like you force pushed to remove the clang format change.

matthewkotila: Basically it would have been easier to discuss all of this if whatever is going on with the latest commit was just put into its own PR.

lkomali: oh! I thought it was a mistake on my end that i changed the clang-format and reverted it back. I should've checked earlier. It wasn't the intention of the last commit to undo any version updates. It's only to fix errors in GAP.
Let me undo my unintentional change.

genai-perf/genai_perf/inputs/retrievers/base_file_input_retriever.py
```diff
@@ -56,7 +56,7 @@ def _verify_file(self, filename: Path) -> None:
 
     def _get_content_from_input_file(self, filename: Path) -> Union[
         Tuple[List[str], List[str]],
-        Tuple[List[str], List[str], List[Dict[Any, Any]]],
+        Tuple[List[str], List[int], List[Dict[Any, Any]]],
```
nv-hwoo, line 59: Is `List[int]` here timestamps?

genai-perf/genai_perf/inputs/retrievers/payload_input_retriever.py
```diff
@@ -127,27 +127,49 @@ def _get_content_from_input_file(
...
+        """
+        Retrieves and validates timestamp from input data
+        """
+        timestamp = data.get("timestamp")
+        if timestamp is None:
+            raise ValueError("Each data entry must have a 'timestamp' field.")
+        try:
+            timestamp = int(timestamp)
+        except ValueError:
+            raise ValueError("Invalid timestamp value.")
```
nv-hwoo, line 146: I think we should give user more information on why it failed. Maybe like this:
```suggestion
raise ValueError(f"Invalid timestamp: Expecting an integer but received '{timestamp}'.")
```

genai-perf/genai_perf/subcommand/profile.py
```diff
@@ -81,6 +82,9 @@ def _report_output(
     elif args.request_rate:
         infer_mode = "request_rate"
         load_level = f"{args.request_rate}"
+    elif args.prompt_source == ic.PromptSource.PAYLOAD:
+        infer_mode = "request_rate"
+        load_level = "1.0"
```
nv-hwoo, line 87: Why are we restricting the level to 1.0?

genai-perf/genai_perf/inputs/retrievers/payload_input_retriever.py
```diff
@@ -127,27 +127,49 @@ def _get_content_from_input_file(
...
+
     def _check_for_optional_data(self, data: Dict[str, Any]) -> Dict[Any, Any]:
         """
         Checks if there is any optional data in the file to pass in the payload.
         """
         optional_data = {
             k: v
             for k, v in data.items()
-            if k not in ["text", "text_input", "timestamp", "hash_ids"]
+            if k
```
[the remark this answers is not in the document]
lkomali: sure, will change that.

genai-perf/genai_perf/inputs/retrievers/base_file_input_retriever.py
lkomali: yes, it's timestamps

genai-perf/genai_perf/inputs/retrievers/payload_input_retriever.py
lkomali: Will modify

genai-perf/genai_perf/subcommand/profile.py
lkomali: When using fixed schedule, we don't set concurrency or request rate. I don't know why but the profile_export.json from PA has mode set to `request_rate` and load_level `1.0` when using fixed schedule. 
Since we are using these values to get profile results, we set it to these values in this mode.
There probably is a better way to handle this instead of setting values like this, but I'm not aware of it.

genai-perf/genai_perf/inputs/retrievers/base_file_input_retriever.py
nv-hwoo: I feel like the return type described here is very complicated. Is there any way to improve this? I see we defined some type aliases. Can we take advantage of those to give more context to the return type?

For example, 
```suggestion
        Tuple[TextData, Timestamps, OptionalData],
```

.pre-commit-config.yaml
```diff
@@ -25,63 +25,63 @@
...
+    - id: mypy
+      types_or: [python, cython]
+      additional_dependencies: [tokenize-rt==3.2.0, types-requests]
+  - repo: local
+    hooks:
+    - id: add-license
+      name: Add License
+      entry: python tools/add_copyright.py
+      language: python
+      require_serial: true
```
dyastremsky, line 87: No newline here. Also, why was the spacing added? Are all the pre-commit hooks still functioning?

genai-perf/genai_perf/inputs/retrievers/payload_input_retriever.py
```diff
@@ -127,27 +127,47 @@ def _get_content_from_input_file(
...
+        Retrieves and validates timestamp from input data
+        """
+        timestamp = data.get("timestamp")
+        if timestamp is None:
+            raise ValueError("Each data entry must have a 'timestamp' field.")
+        try:
+            timestamp = int(timestamp)
+        except ValueError:
+            raise ValueError(
+                f"Invalid timestamp: Expecting an integer but received '{timestamp}'."
```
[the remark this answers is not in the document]
lkomali: I will add a unit test.

nv-hwoo (dismissed):

.pre-commit-config.yaml
dyastremsky: @lkomali It also looks like the Clang pre-commit hook version was also updated and spaces were added before every line.

lkomali: clang-format updated to v18.1.3 by @matthewkotila in his last PR and i reverted that change thinking i made a mistake.
I don't know why the spaces are being added like they were added when i commit.

genai-perf/genai_perf/wrapper.py
```diff
@@ -66,7 +65,8 @@ def add_inference_load_args(args: Namespace) -> List[str]:
     @staticmethod
     def add_payload_args(args: Namespace) -> List[str]:
         cmd = []
-        cmd += ["--fixed-schedule"]
+        if args.prompt_source == PromptSource.PAYLOAD:
+            cmd += ["--fixed-schedule"]
```
matthewkotila, line 69: Kind of a semantic question, but does the existence of `payload:...` in `--input-file` inherently imply that there will be payloads containing the `timestamp` field?

Because it seems more direct/accurate to say that `--fixed-schedule` should be added to the PA command when `timestamp` field is included in all of the payloads. Not simply just the existence of payloads.

Am I misunderstanding something?

dyastremsky: @lkomali @debermudez @nv-hwoo This was what I was asking about earlier. It does seem that it'd be impossible to use the payloads for multi-turn with this logic. There could be a hacky workaround, but it'd be ideal to find a better way to do this before the feature is delivered.

I don't think we want to block the merging of this onto main, but it looks like we actually do need to prioritize some form of detection of fixed schedule vs multi-turn for this feature.

dyastremsky: Thanks for identifying this, Matt.

dyastremsky (approved):

genai-perf/genai_perf/wrapper.py
matthewkotila: @dyastremsky @lkomali I see the PR was merged, but I didn't feel that this conversation was resolved.

Can we get more of a response here to my feedback?

PR #285, opened 2025-02-14
Make PA CLI errors easier to parse

dyastremsky:
Return any errors with the PA CLI after the CLI help menu. This makes it so that the error is immediately visible without scrolling up through all of the CLI args.

After changes, the error goes to the bottom.

Invalid short option:
image

Invalid long option:
image

Invalid general option:
image

### DIFF
--- a/src/command_line_parser.cc
+++ b/src/command_line_parser.cc
@@ -93,10 +93,6 @@ CLParser::FormatMessage(std::string str, int offset) const
 void
 CLParser::Usage(const std::string& msg)
 {
-  if (!msg.empty()) {
-    std::cerr << "Error: " << msg << std::endl;
-  }
-
   std::cerr << "Usage: " << argv_[0] << " [options]" << std::endl;
   std::cerr << "==== SYNOPSIS ====\n \n";
   std::cerr << "\t--version " << std::endl;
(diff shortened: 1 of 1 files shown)


matthewkotila (approved):

src/command_line_parser.cc
```diff
@@ -1812,7 +1813,17 @@ CLParser::ParseCommandLine(int argc, char** argv)
           params_->filename = optarg;
           break;
         case '?':
-          Usage();
+          if (optopt) {
+            Usage(
+                "Error: Invalid short option: '-" +
+                std::string(1, static_cast<char>(optopt)) + "'");
+          } else if (optind > 1 && argv[optind - 1][0] == '-') {
```
matthewkotila, line 1820: Wouldn't you wanna check that the first _two_ characters of `argv[optind - 1]` are `--` to tell user that they passed an invalid long option?

dyastremsky: That case would be handled by the first conditional clause (optopt would be set). See the screenshots above.

In theory, the check for the dash is not necessary, because in our current code it would always evaluate to true. However, in a future PR, we could refactor the code to parse args that do not have flags (with the current code, they get ignored... e.g. `perf_analyzer -m model fake_arg` would ignore fake_arg). In that case, this extra condition is important.

matthewkotila (approved):

PR #290, opened 2025-02-20
Remove session id from payload

matthewkotila:
This is a quick fix to have PA not send session ID to server.

### DIFF
--- a/src/session_concurrency/request_handler.cc
+++ b/src/session_concurrency/request_handler.cc
@@ -71,6 +71,8 @@ RequestHandler::SendRequestAndWaitForResponse(
 {
   auto payload{payload_dataset_manager_->GetPayload(dataset_index)};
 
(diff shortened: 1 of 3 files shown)


src/session_concurrency/request_handler.cc
```diff
@@ -71,6 +71,8 @@ RequestHandler::SendRequestAndWaitForResponse(
 {
   auto payload{payload_dataset_manager_->GetPayload(dataset_index)};
 
+  PayloadJsonUtils::RemoveSessionID(payload);
+
```
matthewkotila, line 75: This is the place where we would take the next payload from the input dataset and add the chat history into it (the multi-turns).

So it's also a good place to strip out the session ID, which would have been present in the input dataset, but now the requirements changed so we shouldn't send the session ID with the payload.

debermudez (approved):

nv-hwoo (approved):

PR #291, opened 2025-02-20
Support hash ids, input_length, and output_length for the fixed schedule feature

debermudez:
In this PR,
1. Added support for caching tokens generated using SyntheticPromptGenerator based on hash_ids and input_length provided in the input file.
2. If output_length is also provided, it will be sent as max_tokens.
3. Unit tests updated accordingly.
4. Small cleanup in PayloadInputRetriever.

### DIFF
--- a/genai-perf/genai_perf/inputs/converters/base_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/base_converter.py
@@ -80,15 +80,13 @@ class BaseConverter:
         """
         if "max_tokens" in optional_data:
             return optional_data["max_tokens"]
-        elif config.output_tokens_mean != DEFAULT_OUTPUT_TOKENS_MEAN:
-            return int(
-                sample_bounded_normal(
-                    mean=config.output_tokens_mean,
-                    stddev=config.output_tokens_stddev,
-                    lower=1,  # output token must be >= 1
-                )
+        return int(
+            sample_bounded_normal(
+                mean=config.output_tokens_mean,
+                stddev=config.output_tokens_stddev,
+                lower=1,  # output token must be >= 1
             )
-        return DEFAULT_OUTPUT_TOKENS_MEAN
+        )
 
     def _add_request_params(
         self,
--- a/genai-perf/genai_perf/inputs/converters/base_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/base_converter.py
@@ -80,13 +80,15 @@ class BaseConverter:
         """
         if "max_tokens" in optional_data:
             return optional_data["max_tokens"]
-        return int(
-            sample_bounded_normal(
-                mean=config.output_tokens_mean,
-                stddev=config.output_tokens_stddev,
-                lower=1,  # output token must be >= 1
+        elif config.output_tokens_mean != DEFAULT_OUTPUT_TOKENS_MEAN:
+            return int(
+                sample_bounded_normal(
+                    mean=config.output_tokens_mean,
+                    stddev=config.output_tokens_stddev,
+                    lower=1,  # output token must be >= 1
+                )
             )
-        )
+        return DEFAULT_OUTPUT_TOKENS_MEAN
 
     def _add_request_params(
         self,
--- a/genai-perf/genai_perf/inputs/converters/base_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/base_converter.py
@@ -72,11 +72,14 @@ class BaseConverter:
     def _get_max_tokens(
         self, config: InputsConfig, optional_data: Dict[str, Any]
     ) -> int:
-        if (
-            config.output_tokens_mean == DEFAULT_OUTPUT_TOKENS_MEAN
-            and "max_tokens" in optional_data
-        ):
-            return config.max_tokens_list.pop(0)
+        """
+        Return the `max_tokens` value to be added in the payload.
+        If `max_tokens` is present in `optional_data`, that value is used.
+        Otherwise, `max_tokens` is sampled from a bounded normal
+        distribution with a minimum value of 1.
+        """
+        if "max_tokens" in optional_data:
+            return optional_data["max_tokens"]
         elif config.output_tokens_mean != DEFAULT_OUTPUT_TOKENS_MEAN:
             return int(
                 sample_bounded_normal(
--- a/genai-perf/genai_perf/inputs/converters/base_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/base_converter.py
@@ -70,7 +70,7 @@ class BaseConverter:
             )
 
     def _get_max_tokens(
-        self, config: InputsConfig, optional_data: Dict[str, Any]
+        self, config: InputsConfig, optional_data: Dict[Any, Any]
     ) -> int:
         """
         Return the `max_tokens` value to be added in the payload.
@@ -94,7 +94,7 @@ class BaseConverter:
         self,
         payload: Dict[Any, Any],
         config: InputsConfig,
-        optional_data: Dict[str, Any],
+        optional_data: Dict[Any, Any],
     ) -> None:
         for key, value in config.extra_inputs.items():
             payload[key] = value
@@ -115,7 +115,7 @@ class BaseConverter:
         triton_format=False,
     ) -> Dict[str, Any]:
         self._add_request_params(payload, config, row.optional_data)
-        self._add_payload_params(payload, row.optional_data)
+        self._add_payload_optional_data(payload, row)
         record: Dict[str, Any] = {}
         if not triton_format:
             record["payload"] = [payload]
--- a/genai-perf/genai_perf/inputs/converters/base_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/base_converter.py
@@ -28,9 +28,13 @@ import random
 from typing import Any, Dict
 
 from genai_perf.exceptions import GenAIPerfException
-from genai_perf.inputs.input_constants import ModelSelectionStrategy
+from genai_perf.inputs.input_constants import (
+    DEFAULT_OUTPUT_TOKENS_MEAN,
+    ModelSelectionStrategy,
+)
 from genai_perf.inputs.inputs_config import InputsConfig
 from genai_perf.inputs.retrievers.generic_dataset import DataRow, GenericDataset
+from genai_perf.utils import sample_bounded_normal
 
 
 class BaseConverter:
@@ -65,8 +69,29 @@ class BaseConverter:
                 f"Model selection strategy '{config.model_selection_strategy}' is unsupported"
             )
 
+    def _get_max_tokens(
+        self, config: InputsConfig, optional_data: Dict[str, Any]
+    ) -> int:
+        if (
+            config.output_tokens_mean == DEFAULT_OUTPUT_TOKENS_MEAN
+            and "max_tokens" in optional_data
+        ):
+            return config.max_tokens_list.pop(0)
+        elif config.output_tokens_mean != DEFAULT_OUTPUT_TOKENS_MEAN:
+            return int(
+                sample_bounded_normal(
+                    mean=config.output_tokens_mean,
+                    stddev=config.output_tokens_stddev,
+                    lower=1,  # output token must be >= 1
+                )
+            )
+        return DEFAULT_OUTPUT_TOKENS_MEAN
+
     def _add_request_params(
-        self, payload: Dict[Any, Any], config: InputsConfig
+        self,
+        payload: Dict[Any, Any],
+        config: InputsConfig,
+        optional_data: Dict[str, Any],
     ) -> None:
         for key, value in config.extra_inputs.items():
             payload[key] = value
@@ -86,8 +111,8 @@ class BaseConverter:
         row: DataRow,
         triton_format=False,
     ) -> Dict[str, Any]:
-        self._add_request_params(payload, config)
-        self._add_payload_optional_data(payload, row)
+        self._add_request_params(payload, config, row.optional_data)
+        self._add_payload_params(payload, row.optional_data)
         record: Dict[str, Any] = {}
         if not triton_format:
             record["payload"] = [payload]
--- a/genai-perf/genai_perf/inputs/inputs_config.py
+++ b/genai-perf/genai_perf/inputs/inputs_config.py
@@ -106,7 +106,6 @@ class InputsConfig:
     # Number of entries to gather
     length: int = DEFAULT_LENGTH
 
-    max_tokens_list: List[int] = field(default_factory=list)
     # The model name
     model_name: List[str] = field(default_factory=list)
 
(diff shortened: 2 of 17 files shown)


genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
```diff
@@ -30,13 +30,16 @@ class SyntheticPromptGenerator:
     _corpus_length = 0
     _prefix_prompts: List[str] = []
     logger = logging.getLogger(__name__)
+    _cache: Dict[int, str] = {}
```
dyastremsky, line 33: Long-term, it'd be good to document for someone reading the code what hash_ids are and what the _cache is, plus how they work. Not a blocker.

genai-perf/genai_perf/parser.py
```diff
@@ -426,6 +426,12 @@ def _print_warnings(args: argparse.Namespace) -> None:
             "Custom tokenizer code can be executed. "
             "This should only be used with repositories you trust."
         )
+    if args.prompt_source == ic.PromptSource.PAYLOAD and args.output_tokens_mean:
```
dyastremsky, line 429: Won't this warning always go off, since we set a default mean? This will add noise. Not a blocker, but we should consider removing it in the near future. We can assume the user will do the right thing.

If users start asking for a warning, the we can always add it only in the case where a payload has an OSL provided and the user specifies a global OSL.

dyastremsky (commented):
LGTM. Only had one semi-blocking comment, and that was just to remove a few outdated lines of code. Otherwise, looks good to approve.

genai-perf/genai_perf/inputs/retrievers/payload_input_retriever.py
```diff
@@ -111,32 +109,82 @@ def _get_content_from_input_file(
...
-                        )
-                    prompt = prompt if prompt else prompt_alt
+                    prompt = self._get_prompt(data)
                     prompts.append(prompt.strip() if prompt else prompt)
                     timestamp = self._get_valid_timestamp(data)
                     timestamps.append(timestamp)
                     optional_data = self._check_for_optional_data(data)
                     optional_datas.append(optional_data)
-        return prompts, timestamps, optional_datas
+                    max_token = data.get("output_length")
```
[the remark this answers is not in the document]
lkomali: Oh yes, forgot to remove this, I will remove it now.

genai-perf/genai_perf/inputs/inputs_config.py
```diff
@@ -106,6 +106,7 @@ class InputsConfig:
     # Number of entries to gather
     length: int = DEFAULT_LENGTH
 
+    max_tokens_list: List[int] = field(default_factory=list)
```
nv-hwoo, line 109: Can we add a description of this member?

genai-perf/genai_perf/inputs/converters/base_converter.py
```diff
@@ -65,8 +69,29 @@ def _select_model_name(self, config: InputsConfig, index: int) -> str:
...
             )
 
+    def _get_max_tokens(
+        self, config: InputsConfig, optional_data: Dict[str, Any]
+    ) -> int:
+        if (
+            config.output_tokens_mean == DEFAULT_OUTPUT_TOKENS_MEAN
+            and "max_tokens" in optional_data
+        ):
+            return config.max_tokens_list.pop(0)
```
nv-hwoo, line 79: Can we add a docstring  that describes the behavior of this method?

genai-perf/genai_perf/inputs/inputs_config.py
lkomali: This should be removed. I implemented by using `max_tokens_list` earlier and then moved to adding max_tokens in optional_data instead of as a member of input_config. So, I removed this, I will push the updated changes.

genai-perf/genai_perf/inputs/converters/base_converter.py
lkomali: Fixed a small error in the function, will add docstring.

genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
lkomali: Adding a docstring fo the `generate_prompt_with_token_reuse` function. The fixed schedule doc should also have details about this.

genai-perf/genai_perf/inputs/converters/base_converter.py
```diff
@@ -65,8 +69,29 @@ def _select_model_name(self, config: InputsConfig, index: int) -> str:
                 f"Model selection strategy '{config.model_selection_strategy}' is unsupported"
             )
 
+    def _get_max_tokens(
```
nv-hwoo, line 72: I don't think base converter should have this method. This is specific to LLM related converters. What I think could be a better alternative would be to modify `config.output_tokens_mean` _before_ calling `_add_request_params()` in the LLM converters. That way, we probably don't need `_get_max_tokens` in the base converter.

For example, something like below inside OpenAIChatCompletionsConverter::convert
```python
if "max_tokens" in optional_data:
    config.output_tokens_mean = config.max_tokens_list.pop(0)

request_body["data"].append(
    self._finalize_payload(payload, config, row)
)
```

genai-perf/genai_perf/parser.py
dyastremsky: Yeah. If you wanted to hide this for most folks in the short-term, you could do something like `... and args.output_token_mean != ic.DEFAULT_OUTPUT_TOKENS_MEAN`.

genai-perf/genai_perf/inputs/converters/base_converter.py
dyastremsky: Won't that require the logic to be included in every converter? We could have an LLM converter class that's derived from the base converter class if there's utility. This is a general helper method that's meant to abstract away the logic for getting max tokens for any converters that needs the value.

genai-perf/tests/test_cli.py
```diff
@@ -1087,6 +1088,27 @@ def test_check_payload_input_args_valid(self, monkeypatch, mocker):
         except SystemExit:
             pytest.fail("Unexpected error in test")
 
+    def test_print_warnings_payload(self, monkeypatch, mocker):
+        expected_warning_message = "--output-tokens-mean is incompatible with output_length in the payload input file. output-tokens-mean will be ignored in favour of per payload settings"
```
nv-hwoo, line 1092: nit: break it into multi line strings

genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
```diff
@@ -30,13 +30,16 @@ class SyntheticPromptGenerator:
...
 
     @classmethod
     def create_synthetic_prompt(
         cls,
         tokenizer: Tokenizer,
         prompt_tokens_mean: int = 550,
         prompt_tokens_stddev: int = 250,
+        hash_ids: Optional[List[int]] = None,
+        block_size: int = 512,
```
nv-hwoo, line 42: What is "block" here? Is it a prompt length? Also no caller seems to pass values for this argument.

genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
```diff
@@ -120,6 +128,29 @@ def _generate_prompt(cls, tokenizer: Tokenizer, num_tokens: int) -> str:
...
+                if index == len(prompt_hash_list) - 1:
+                    size_to_use = prompt_tokens_mean - (index * block_size)
+                if hash_index not in cls._cache:
+                    prompt = cls._generate_prompt(tokenizer, size_to_use)
+                    cls._cache[hash_index] = prompt
+
+                final_prompt.append(cls._cache[hash_index])
+            prompt = " ".join(final_prompt)
+
+        return prompt
```
nv-hwoo, line 152: What happens if `prompt_hash_list` is `None`? Where does `prompt` come from?

genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
```diff
@@ -120,6 +128,29 @@ def _generate_prompt(cls, tokenizer: Tokenizer, num_tokens: int) -> str:
...
         return tokenizer.decode(prompt_tokens)
 
+    @classmethod
+    def _generate_prompt_with_token_reuse(
+        cls,
+        tokenizer: Tokenizer,
+        prompt_tokens_mean: int,
+        prompt_hash_list: list,
+        block_size: int,
+    ) -> str:
```
nv-hwoo, line 138: Could you provide a description of the behavior of this method? This method seems to contain pretty complex logic and a good description will save a lot of time for future readers.

lkomali: Yep adding a docstring.

genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
lkomali: block_size default 512 basically, block_size number of tokens form a block and that block has a hash_id.
so if users provides
hash_ids: [0,1]
hash_ids: [0,2], first chunk of tokens i.e first 512 prompts are the same and the last 512 or remaining size which is calculated using input_length provided by user will differ).
The idea is to ultimately add a CLI option for user to specify block_size. But right now, default value is being used.

genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
lkomali: if its None or empty, tokens are generated in the standard way we have. 

```
if hash_ids:
            return cls._generate_prompt_with_token_reuse(
                tokenizer, prompt_tokens_mean, hash_ids, block_size
            )
```
We have this check in `create_synthetic_prompt()` method, and if its None or empty
`cls._generate_prompt(tokenizer, num_prompt_tokens)`
this method will be called which is the default behavior.

genai-perf/genai_perf/inputs/converters/base_converter.py
nv-hwoo: Why would other converters require this logic  to be included? Wouldn't it be just openai converters?

genai-perf/tests/test_cli.py
lkomali: i hoped pre-commit would do it, i will split it.

genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
nv-hwoo: Then the condition `if prompt_hash_list is not None:` in this method seems redundant/unnecessary?

lkomali: yeah true, redundant checking.

genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
nv-hwoo: I see so block == sequence of tokens. Could you add that to the docstring? I will definitely forget your explanation if I needed to revisit this code later in the future.

genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
nv-hwoo: I see. Can we remove this line?

genai-perf/genai_perf/inputs/retrievers/payload_input_retriever.py
```diff
@@ -101,8 +97,10 @@ def _get_content_from_input_file(
 
         Returns
         -------
-        Tuple[List[str], Dict, str]
-            A list of prompts, and optional data.
+        A dictionary containing extracted data:
+            - "prompts": List[str] - Extracted prompt texts
+            - "timestamps": List[int] - Corresponding timestamps
+            - "optional_datas": List[Dict[Any, Any]] - Any optional data
```
nv-hwoo, line 103: ❤️

genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
lkomali: things were done in a rush, so missed adding docstrings. 
Thanks for calling that out and clarifying.

genai-perf/genai_perf/inputs/converters/base_converter.py
debermudez: For now but if the nova folks start using other APIs, we will need it elsewhere.
Adding it to the base convertor, makes it easier to access generally.

nv-hwoo: Talked with @dyastremsky offline. We can move on for now but we both agreed it's not ideal to have token specific method inside the base converter. And we want something like LLM utils class in the future to contain these logics.

nv-hwoo (dismissed):

genai-perf/genai_perf/inputs/converters/base_converter.py
```diff
@@ -65,8 +69,32 @@ def _select_model_name(self, config: InputsConfig, index: int) -> str:
...
+    ) -> int:
+        """
+        Return the `max_tokens` value to be added in the payload.
+        If `max_tokens` is present in `optional_data`, that value is used.
+        Otherwise, `max_tokens` is sampled from a bounded normal
+        distribution with a minimum value of 1.
+        """
+        if "max_tokens" in optional_data:
+            return optional_data["max_tokens"]
+        elif config.output_tokens_mean != DEFAULT_OUTPUT_TOKENS_MEAN:
```
dyastremsky, line 83: This if/elif/else has a bug. It's valid for a user to specify a non-default stddev with a default mean. If you specify --output-tokens-stddev=100, this will ignore that. This should look more like:
```
if "max_tokens" in optional_data:
            return optional_data["max_tokens"]
return int(
                sample_bounded_normal(
                    mean=config.output_tokens_mean,
                    stddev=config.output_tokens_stddev,
                    lower=1,  # output token must be >= 1
                )
            )
```

CC: @lkomali @debermudez if this ends up being merged for ease, I'd recommend a quick follow-up PR with this fix.

dyastremsky (dismissed):

genai-perf/genai_perf/inputs/converters/base_converter.py
lkomali: I think the original is to ignore the `stddev`. 
Making this change broke the expected values of unit tests.
I will revert to original implementation for now and maybe tackle it in a follow-up PR as you originally suggested.

dyastremsky: Oh, you're right. I was thinking of prompt tokens, since that's what we usually update. Apologies for the confusion. Thanks for catching and fixing this.

dyastremsky (approved):

dyastremsky:
@debermudez @lkomali I believe you tested hash_ids in the CI. Can you add testing somewhere for output_length as well? I missed that those unit tests weren't in this PR.

PR #292, opened 2025-02-20
Initial replaying datasets doc

debermudez:
Placeholder text

### DIFF
--- a/genai-perf/docs/replaying_datasets.md
+++ b/genai-perf/docs/replaying_datasets.md
@@ -1,0 +1,71 @@
+<!--
+Copyright (c) 2024-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+ * Redistributions of source code must retain the above copyright
+   notice, this list of conditions and the following disclaimer.
+ * Redistributions in binary form must reproduce the above copyright
+   notice, this list of conditions and the following disclaimer in the
+   documentation and/or other materials provided with the distribution.
+ * Neither the name of NVIDIA CORPORATION nor the names of its
+   contributors may be used to endorse or promote products derived
+   from this software without specific prior written permission.
+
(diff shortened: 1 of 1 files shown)


genai-perf/docs/replaying_datasets.md
```diff
@@ -0,0 +1,71 @@
...
+CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+-->
+
+# Replaying a dataset
```
lkomali, line 29: Although this is not wrong, IMO replaying server trace or something similar might be better.
Replaying a dataset is little confusing as it can be interpreted as something to do with datasets and not replaying requests.

genai-perf/docs/replaying_datasets.md
```diff
@@ -0,0 +1,71 @@
...
+## Requirements
+1. New option: `--input-file payload:<path_to_file>`
+2. New payload file format
+
+## Payload file format
+
+The payload file format has a couple possibilities to describe the dataset. Utilzing jsonl,
+a user can define a dataset via the following:
+
+1. Timestamp and text prompt
```
lkomali, line 43: Its better if we define the unit of timestamp here in this section. I think that timestamp should be in milliseconds. GAP doesn't do any conversion, it sends timestamp values as is to PA. So it's better to  check in what unit(seconds, milliseconds etc.) PA is expecting and we define the unit here to ensure consistency.

PR #293, opened 2025-02-21
Redesign session ID out of payload

matthewkotila:


genai-perf/genai_perf/inputs/input_constants.py
```diff
@@ -71,7 +71,7 @@ def to_lowercase(self):
 DEFAULT_SYNTHETIC_FILENAME = "synthetic_data.json"
 DEFAULT_WARMUP_REQUEST_COUNT = 0
 DEFAULT_BACKEND = "tensorrtllm"
-PAYLOAD_METADATA_FIELDS = ["timestamp", "delay"]
+PAYLOAD_METADATA_FIELDS = ["timestamp", "delay", "session_id"]
```
matthewkotila, line 74: This part of GAP was designed well, seems like all I needed to do was add this here and the session IDs as supplied via `--input-file=payload:my_payloads.jsonl` would be moved outside of the payload one level up in the input data JSON given to PA.

genai-perf/genai_perf/subcommand/profile.py
```diff
@@ -86,7 +86,7 @@ def _report_output(
     # Setting to default values to avoid an error.
     elif args.prompt_source == ic.PromptSource.PAYLOAD:
         infer_mode = "request_rate"
-        load_level = "1.0"
+        load_level = "0.0"
```
matthewkotila, line 89: A design artifact of the PA profile data export is that each export requires an inference load mode of either request rate or concurrency and also a particular level for either.

This isn't a very accurate way to describe experiments, and should be redesigned.

In the meantime, GAP needs to access experiment statistics using those values, and in the case of session concurrency mode, it defaults the mode to request rate mode and the level to `0.0`

So this change allows GAP to grab the statistics from session concurrency mode.

It might be better if there's like a version of `get_statistics()` that just returns the first statistics, possibly asserting that there's only one statistic available.

genai-perf/genai_perf/wrapper.py
```diff
@@ -66,7 +66,7 @@ def add_inference_load_args(args: Namespace) -> List[str]:
     def add_payload_args(args: Namespace) -> List[str]:
         cmd = []
         if args.prompt_source == PromptSource.PAYLOAD:
-            cmd += ["--fixed-schedule"]
+            cmd += ["--session-concurrency=2"]
```
matthewkotila, line 69: A hard-coding to allow GAP to run session concurrency mode until [lkomali-support-session-args-synthetic-multiturn](https://github.com/triton-inference-server/perf_analyzer/compare/main...lkomali-support-session-args-synthetic-multiturn) gets merged.

src/model_parser.cc
```diff
@@ -334,6 +334,11 @@ ModelParser::InitOpenAI(
     delay_input->second.datatype_ = "UINT64";
     delay_input->second.shape_.push_back(1);
     delay_input->second.is_optional_ = true;
+
+    auto session_id_input = inputs_->emplace("session_id", ModelTensor()).first;
+    session_id_input->second.name_ = "session_id";
+    session_id_input->second.datatype_ = "BYTES";
+    session_id_input->second.shape_.push_back(1);
```
matthewkotila, line 341: As shown [here](https://github.com/triton-inference-server/perf_analyzer/pull/293/files#r1965109842), session ID is now treated as a top-level "inference input", and is no longer nested inside the `payload` "inference input" (which is also declared a few lines above this).

src/test_payload_dataset_manager.cc
```diff
@@ -46,54 +46,47 @@ TEST_CASE("PayloadDatasetManager::GroupPayloadsBySession")
...
-    const std::string model_name{""};
-    const std::string model_version{""};
-    const int32_t batch_size{0};
-    const InferenceLoadMode inference_load_mode{
-        InferenceLoadMode::SessionConcurrency};
+    auto model_parser{std::make_shared<ModelParser>()};
 
-    model_parser->InitOpenAI(
-        model_name, model_version, batch_size, inference_load_mode);
+    (*model_parser->Inputs())[name] = session_id_model_tensor;
```
matthewkotila, line 60: Cleaned up some random stuff in this test. Skipped calling `InitOpenAI()` in favor of just creating the `ModelTensor` itself and assigning to a blank `ModelParser` to be used later in the test.

Previously, the test was relying on InitOpenAI() to create `ModelTensor`s and assign them to the `ModelParser`.

This just relies on less code now, further isolating what we want to test.

src/test_payload_dataset_manager.cc
```diff
@@ -46,54 +46,47 @@ TEST_CASE("PayloadDatasetManager::GroupPayloadsBySession")
...
 
-    std::string input_data_json{R"(
+    const std::string input_data_json{R"(
```
matthewkotila, line 62: Added `const` where I could since that's the clean code guideline.

src/test_payload_dataset_manager.cc
```diff
@@ -137,4 +134,88 @@ TEST_CASE("PayloadDatasetManager::GroupPayloadsBySession")
   }
 }
 
+TEST_CASE("PayloadDatasetManager::GetDelayForPayload")
```
matthewkotila, line 137: I saw there was previously no test for this public method, so I made a simple test that fetches a delay input from the DataLoader.

src/test_payload_dataset_manager.cc
```diff
@@ -137,4 +134,88 @@ TEST_CASE("PayloadDatasetManager::GroupPayloadsBySession")
...
+      data_loader, model_parser);
+
+  const size_t dataset_index{0};
+
+  const auto delay{payload_dataset_manager.GetDelayForPayload(dataset_index)};
+
+  CHECK(delay == std::chrono::milliseconds(1000));
+}
+
+TEST_CASE("PayloadDatasetManager::GetPayload")
```
matthewkotila, line 179: Similarly there was no test for this public method, so I added one.

It fetches a payload out of the DataLoader via the PayloadDatasetManager class.

src/test_payload_json_utils.cc
```diff
@@ -35,47 +35,6 @@
 
 namespace triton::perfanalyzer {
 
-TEST_CASE("PayloadJsonUtils::GetSession")
```
matthewkotila, line 38: Again, this functionality was moved into the PayloadDatasetManager.

Session ID no longer is stored inside the payload JSON, so there's no need to have functionality (or a test) for `PayloadJsonUtils` for fetching the session ID.

Session ID is now part of the input data JSON top-level inference inputs themselves, and is fetched via the `PayloadDatasetManager` (`DataLoader` under the hood).

src/inference_profiler.cc
```diff
@@ -1413,8 +1413,12 @@ InferenceProfiler::CollectData(
...
     std::vector<RequestRecord>&& request_records)
 {
-  ProfileDataCollector::InferenceLoadMode id{
-      summary.concurrency, summary.request_rate};
+  ProfileDataCollector::InferenceLoadMode id{};
+  if (dynamic_cast<CustomRequestScheduleManager*>(manager_.get())) {
+    id = {0, 0.0};
+  } else {
+    id = {summary.concurrency, summary.request_rate};
+  }
```
matthewkotila, line 1421: PA's profile export JSON currently requires a concurrency/request rate associated with each export--this is an artifact of an old design and should be updated eventually.

 Fixed schedule mode has no concept of concurrency or request rate, but since a value is needed to create the profile export JSON, this code specifies the export's "concurrency/request rate" as zero, a sentinel value.

Code on GAP side (see [here](https://github.com/triton-inference-server/perf_analyzer/pull/293/files#diff-9d974f7d793b676e0d1155531fdc4d57c42ad7b1e02f8e4123285d5cfe041205R93)) will be expecting a profile export with concurrency/request rate of zero for the fixed schedule case.

src/session_concurrency/payload_dataset_manager.cc
```diff
@@ -85,6 +85,12 @@ PayloadDatasetManager::GetDelayForPayload(size_t dataset_index) const
     throw std::runtime_error(error.Message());
   }
 
+  if (!delay_ms_tensor_data.data_ptr) {
+    throw std::runtime_error(
+        "Missing 'delay' input for payload with index " +
+        std::to_string(dataset_index));
+  }
+
```
matthewkotila, line 93: Good catch @dyastremsky , the code didn't fully check for the missing delay value case, and now it does, and has been tested both in unit tests and manually :)

src/session_concurrency/payload_dataset_manager.cc
```diff
@@ -130,9 +136,23 @@ PayloadDatasetManager::CreateSessionIdToPayloadsMap() const
...
+  const uint8_t* session_id_buffer{session_id_tensor_data.data_ptr};
+  const size_t session_id_byte_size{session_id_tensor_data.batch1_size};
+
+  const std::string session_id(
+      reinterpret_cast<const char*>(session_id_buffer), session_id_byte_size);
 
-  return PayloadJsonUtils::GetSessionID(payload);
+  return session_id;
```
matthewkotila, line 155: This is the new code that gets the session ID as a model input rather than how it was before where it as a JSON sub-field of the payload model input.

src/session_concurrency/payload_dataset_manager.cc
dyastremsky: Nice job, Matt!

dyastremsky (approved):
Nice clean-up, Matt! Left a couple of questions, but if this is working, this looks good.

genai-perf/genai_perf/subcommand/profile.py
matthewkotila: The reason why the profile export was putting this metadata along with each experiment is because technically PA can sweep through multiple concurrencies or request rates, and all of them would go in the profile export.

I think the profile export design needs to be revisited, frankly. I don't have all the solutions figured out yet for it.

genai-perf/genai_perf/inputs/input_constants.py
matthewkotila, line 1: Whoops, didn't mean to re-request review from you, David. Just waiting on CI to pass to convert to regular PR and then merge.

PR #295, opened 2025-02-21
Update docs and templates

nv-hwoo:


genai-perf/README.md
```diff
@@ -556,6 +556,31 @@ The number of tokens in each prefix prompt. This value is only used if
 being concatenated, the number of tokens in the final prompt may be off by one.
 (default: `100`)
 
+##### `--image-width-mean <int>`
+
+The mean width of images when generating synthetic image data.
```
dyastremsky, line 561: Do you think it'd be useful to specify that you are measuring in pixels (I'm assuming that's the unit)?

genai-perf/README.md
```diff
@@ -556,6 +556,31 @@ The number of tokens in each prefix prompt. This value is only used if
...
+
+##### `--image-height-stddev <int>`
+
+The standard deviation of height of images when generating synthetic image data.
+(default: `0`)
+
+##### `--image-format <str>`
+
+The compression format of the images. If format is not selected,
+format of generated image is selected at random.
```
dyastremsky, line 582: There's not a default compression format?

nv-hwoo: Yes, when image format is not specified, synthetic generator randomly selects the format: https://github.com/triton-inference-server/perf_analyzer/blob/9e682f4550cb9635b2254fba10c4baa47cd27f39/genai-perf/genai_perf/inputs/retrievers/synthetic_image_generator.py#L57

genai-perf/README.md
nv-hwoo: Yeah, I think it would be clearer for the reader. Mentioned pixel as unit in the desc.

genai-perf/README.md
dyastremsky: Interesting. I think it'd be helpful to give JPEG/PNG as examples, as I didn't immediately know that's what this meant. But users can always use genai-perf profile --help to see that, so I don't think this is imperative.

dyastremsky (approved):
Stellar work, Hyunjae!

PR #297, opened 2025-02-24
Compare subcommand changes to work with config file

nv-braf:
Add `compare` to the config file and convert the subcommand's logic to use config instead of argparse.

### DIFF
--- a/genai-perf/genai_perf/config/input/config_analyze.py
+++ b/genai-perf/genai_perf/config/input/config_analyze.py
@@ -34,10 +34,10 @@ class ConfigAnalyze(BaseConfig):
         # yapf: disable
         sweep_parameter_template_comment = \
         (f"Uncomment the lines below to enable the analyze subcommand\n"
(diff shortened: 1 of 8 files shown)


genai-perf/genai_perf/config/input/config_analyze.py
```diff
@@ -34,10 +34,10 @@ def __init__(self) -> None:
         # yapf: disable
         sweep_parameter_template_comment = \
         (f"Uncomment the lines below to enable the analyze subcommand\n"
-         f"For further details see analyze.md sweep_parameters:\n"
-         f"  concurrency:\n"
-         f"    start: {AnalyzeDefaults.MIN_CONCURRENCY}\n"
-         f"    stop: {AnalyzeDefaults.MAX_CONCURRENCY}")
+         f"#  For further details see analyze.md sweep_parameters:\n"
```
nv-braf, line 37: This is just a formatting change to ensure that if the user also uncomments this line the template still works.

debermudez (approved):

PR #299, opened 2025-02-25
Fix PA unknown option parsing

dyastremsky:
Fix unknown option parsing in Perf Analyzer.

Passing unit tests:
image

Error output:
image


src/command_line_parser.cc
```diff
@@ -1837,16 +1837,23 @@ CLParser::ParseCommandLine(int argc, char** argv)
...
+                "Error: Invalid short option '-" +
+                    std::string(1, static_cast<char>(optopt)) + "'",
+                GENERIC_ERROR);
+          } else if (arg.rfind("--", 0) == 0 && optopt == 0) {
+            throw PerfAnalyzerException(
+                "Error: Invalid long option '" + arg + "'", GENERIC_ERROR);
+          } else if (
+              optarg == nullptr && optind > 1 && argv[optind - 1][0] == '-') {
+            throw PerfAnalyzerException(
+                "Error: Missing value for option '" + arg + "'", GENERIC_ERROR);
```
matthewkotila, line 1853: This is really hard to read--can this be rewritten? Maybe with helper functions?

dyastremsky: Good feedback. I updated some of the variable names and added booleans to make this easier to read.

src/command_line_parser.cc
```diff
@@ -1837,16 +1837,31 @@ CLParser::ParseCommandLine(int argc, char** argv)
...
-          } else if (optind > 1 && argv[optind - 1][0] == '-') {
-            Usage(
-                "Error: Invalid long option: '" +
-                std::string{argv[optind - 1]} + "'");
+          std::string last_arg =
+              (optind > 1) ? std::string(argv[optind - 1]) : "";
+
+          bool is_invalid_short_option = optopt && isprint(optopt);
+          bool is_invalid_long_option =
+              last_arg.rfind("--", 0) == 0 && optopt == 0;
```
matthewkotila, line 1845: Since we're using c++ 20 can we use [startswith()](https://en.cppreference.com/w/cpp/string/basic_string/starts_with)?

src/command_line_parser.cc
```diff
@@ -1837,16 +1837,31 @@ CLParser::ParseCommandLine(int argc, char** argv)
...
+          bool is_missing_value =
+              (optarg == nullptr && optind > 1 && last_arg[0] == '-');
```
matthewkotila, line 1847: What does `optind > 1` actually mean? It's little things like this that if the code can better explain, we'll be able to maintain it more easily going forward.

dyastremsky: It means that the index is greater than 1. The logic that this required was relatively complex. I tried to simplify it as much as I could, but anything simpler would fail some edge cases.

If you have simpler logic or a better way to do this, please feel free to suggest a path forward. I'd happily have this be more intuitive. Whoever wrote the code originally made the choice to use `getopt`, so I had to learn it to make this code work within the getopt framework. If an engineer isn't familiar with getopt, they'd have to read docs to understand how parsing working unless we want to use a different parsing framework and change the file to use it.

src/command_line_parser.cc
dyastremsky: My counter to this would be that the rest of the file using rfind, not starts_with. We can create a refactor ticket, but I'd rather be consistent if we're making a change for stylistic reasons.

matthewkotila: I've actually started to change my philosophy on this.

We will never get around to changing all the other stuff. I prefer to break patterns if it involves writing code that is locally easier to maintain.

src/command_line_parser.cc
matthewkotila: What does the index being greater than 1 mean?

dyastremsky: @nicolasnoble, can we get your suggestions here? I don't know how to convert this into something more readable. I learned the `getopt_long` flags and options because that's what this parser uses, but short of turning this code into a lesson on how to use the C library, I don't know any way to make it easier for a user to read this.

src/command_line_parser.cc
dyastremsky: I tried changing it to use `starts_with` and it complained that the member does not exist. I followed [our build instructions](https://github.com/triton-inference-server/perf_analyzer/blob/main/docs/install.md#build-from-source).

src/command_line_parser.cc
nicolasnoble: We... could refactor the whole of the CLI parser, really. I've written and used command line parsers, and there's some that are readily available, like https://github.com/Taywee/args which is a single header MIT licensed parser, meaning it's a very simple dependency to add. It'd be a bit (?) of an undertaking to overhaul the current parser to use this instead, but it's probably worth it and would reduce a lot of the cruft we currently have.

nicolasnoble: Thinking about it a bit more, it's probably doable to have two flags systems co-existing in order to migrate from a C-style getops to a C++ interface instead piece by piece, in order to avoid having to refactor everything all at once. We'll try to discuss this tomorrow.

src/command_line_parser.cc
matthewkotila: You did `last_arg.starts_with("--")`?

Strange. cmake tells g++ to use `--std=c++20` [here](https://github.com/triton-inference-server/perf_analyzer/blob/aed6e3d4dddf18a5f13c60152f2433c33eb808a2/src/CMakeLists.txt#L241) and [here](https://github.com/triton-inference-server/perf_analyzer/blob/9e682f4550cb9635b2254fba10c4baa47cd27f39/CMakeLists.txt#L32)...

src/command_line_parser.cc
matthewkotila: I think a whole refactor is a great idea but too big for this PR.

I answered my own question by asking chatgpt and reading the docs.

The reason for checking `optind > 1` is basically to answer the question "has at least one option already been processed?", which kinda makes sense because an option with a required argument would need to first process the option itself (--my-option, i.e. optind=1) before moving on to process the options value (my_value, .e.g. --my-option=my_value, i.e. optind=2).

src/command_line_parser.cc
dyastremsky: I did. Let me try a clean rebuild again shortly. It looks like the version of gcc and CMake that apt gets should support C++ 20, but we'll see.

src/command_line_parser.cc
matthewkotila: I'm ok with it as whatever it is right now.

src/command_line_parser.cc
matthewkotila: I tested a small explain of starts_with in the main function of PA and compiled and it worked.

matthewkotila: I figured out the issue, it's from building the unit tests binary, it was missing a similar declaration of c++20.

Can you add this to src/CMakeLists.txt:417?

```cmake
target_compile_features(
  perf_analyzer_unit_tests
  PRIVATE
    cxx_std_${TRITON_MIN_CXX_STANDARD}
)

```

dyastremsky: Thanks for investigating and finding a fix! I thought it might be CMake-related. I need to invest more time into understanding CMake when I have capacity.

matthewkotila: My long-term goal is to have the cmake clean enough where there isn't as much of a need to be a wizard at it.

nicolasnoble: > My long-term goal is to have the cmake clean enough where there isn't as much of a need to be a wizard at it.

I don't think that's possible :P

dyastremsky: @matthewkotila, your change worked for compiling. The unit tests were passing locally but then were hanging mid-way through. I'm rerunning them now and seeing if the CI passes.

dyastremsky: I have been trying to run locally but it keeps hanging:
image

And it's failing the CI as well, though it does seem to finish there. I am trying to rerun it in case it's a one-off.

As a note, respectfully, this is why I prefer to batch new changes into their own PR. I appreciate your point that this is about readability and not just style, but at this point, it has failed to compile, you needed to spend time figuring out how to get it to compile, and now I have hangs/failures that I cannot explain. I have spent several extra hours on this on and off. I'm neither super familiar with this PA code nor am I familiar with the C++ 20 functions, so me introducing them here has resulted in a lot of work that was not accounted for in the initial ticket or PR whereby I was just trying to fix the CI failure. I don't think the work was justified. It's possible the work would have been quick for someone more familiar with this class and C++ 20, but I think introducing new code that one is unfamiliar always comes with risk.

ganeshku1: Thank you @matthewkotila and appreciate the efforts here.
Not trying to discourage the good suggestion about changes related to C++20, Can we create a follow up ticket for the C++ 20 changes and make the changes as part of that ticket?
We can take that task up as soon as next sprint?

ganeshku1: Looks like David Y already made the changes. If this is holding up lets create a separate PR and ticket for further changes related to C++20.

dyastremsky: I added the C++ 20 change, but it's resulted in a failure in the CI and a hang locally with no clear explanation for why that would have made the difference. Reverting it and creating a follow-up ticket for C++ 20 changes sounds really good to me, if the team is okay with it.

I'm going to revert the change and see if it then passes CI again.

matthewkotila: Oh, yeah, sure if adding c++20 here causes new issues then that makes sense to postpone that component of the cleanup 😅.

To be clean, I want an optimization between clean code and quick PRs, haha...

dyastremsky: Actually, PA unit tests are still hanging locally, and I don't think the failure is related to my changes. @matthewkotila, is it possible it has to do with https://github.com/triton-inference-server/perf_analyzer/pull/293? It looks like it's this failure: `test_profile_data_exporter.cc:219: FATAL ERROR: test case CRASHED: SIGSEGV - Segmentation violation signal`.

dyastremsky: Validated the theory. Even after reverting, the CI fails in the same spot. I think this is a new failure that was introduced in the last 24 hours, unrelated to this change. We can leave the C++ 20 update, but we do need to fix the new SIGSEGV which exists independently from this PR.

matthewkotila: I determined the root cause of the hanging (seg fault in my case) and fixed with latest commit.

Running new CI, after it passes, we can merge this.

src/test_profile_data_exporter.cc
```diff
@@ -244,9 +246,9 @@ TEST_CASE("profile_data_exporter: AddDataToJSON")
         reinterpret_cast<const uint8_t*>(data),
         reinterpret_cast<const uint8_t*>(data) + sizeof(data));
     exporter.AddDataToJSON(json, buf, "BOOL");
-    CHECK(json[0] == true);
-    CHECK(json[1] == false);
-    CHECK(json[2] == true);
+    CHECK(json[0].GetBool() == true);
+    CHECK(json[1].GetBool() == false);
+    CHECK(json[2].GetBool() == true);
```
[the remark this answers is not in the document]
dyastremsky: Great job, thanks for adding these fixes! 👏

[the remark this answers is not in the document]
dyastremsky: @nv-hwoo @ganeshku1, would either of you be able to do a quick review? Matt and I can't approve this PR for merging.

nv-hwoo (approved):

PR #301, opened 2025-02-27
Add unit and integration testing for multi-turn

dyastremsky:
Add unit and integration testing for multi-turn, which is used with sessions (e.g. --session-concurrency).

This PR adds testing for multi-turn in the payload retriever tests (which already test for timestamp and session ID, so this adds delay). It also adds an integration test for inputs being generated from a payload file and written to an inputs file.

### DIFF
--- a/genai-perf/README.md
+++ b/genai-perf/README.md
@@ -633,7 +633,7 @@ export file is `profile_export.json`, the genai-perf file will be exported to
 
 ### Session Options
 
-##### `--session-concurrency`
+##### `--session-concurrency <int>`
 
 The number of concurrenct sessions to benchmark. (default: `None`)
 
--- a/genai-perf/README.md
+++ b/genai-perf/README.md
@@ -631,6 +631,12 @@ exported to `<profile_export_file>_genai_perf.json` and
 export file is `profile_export.json`, the genai-perf file will be exported to
 `profile_export_genai_perf.csv`. (default: `profile_export.json`)
 
+### Session Options
+
+##### `--session-concurrency`
+
+The number of concurrenct sessions to benchmark. (default: `None`)
+
 ### Other Options
 
 ##### `--tokenizer <str>`
--- a/genai-perf/tests/integration_tests/test_multiturn.py
+++ b/genai-perf/tests/integration_tests/test_multiturn.py
@@ -1,0 +1,92 @@
+# Copyright 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of NVIDIA CORPORATION nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+import json
+from pathlib import Path
+from unittest.mock import mock_open, patch
+
+import pytest
+from genai_perf.inputs.input_constants import PromptSource
+from genai_perf.inputs.inputs import Inputs
+from genai_perf.inputs.inputs_config import InputsConfig
+from genai_perf.inputs.retrievers.generic_dataset import GenericDataset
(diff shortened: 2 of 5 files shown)


templates/genai-perf-templates/README_template
nv-hwoo, line 1: Just making sure: Did you run `python generate_docs.py` after the change?

genai-perf/tests/integration_tests/test_multiturn.py
```diff
@@ -0,0 +1,92 @@
...
+                self.input_type = PromptSource.PAYLOAD
+                self.extra_inputs = {}
+
+        return MockConfig()
+
+    @pytest.fixture
+    def retriever(self, mock_config):
+        return PayloadInputRetriever(mock_config)
+
+    def test_multi_turn_input__generation(self, monkeypatch, mock_config):
```
nv-hwoo, line 59:
```suggestion
    def test_multi_turn_input_generation(self, monkeypatch, mock_config):
```

genai-perf/tests/integration_tests/test_multiturn.py
```diff
@@ -0,0 +1,92 @@
...
+            assert isinstance(
+                output_data, dict
+            ), f"Expected dict, got: {type(output_data)}"
+            assert "data" in output_data, f"Unexpected JSON structure: {output_data}"
+
+            for entry in output_data["data"]:
+                assert "session_id" in entry, "Session ID missing in output entry"
+                assert (
+                    "delay" in entry or "timestamp" in entry
+                ), "Either delay or timestamp must be present in output entry"
```
nv-hwoo, line 92: Using loop to check the values are fine, but I would prefer defining an "expected" data structure and compare that against the actual one. Something like:
```python
expected_data = {
    "data": {
        "session_id": ...,
        "delay": ...,
    }
}

assert expected_data == actual_data
```
One big benefit of this approach is that it's super crystal clear what the expected return values are given the input.

genai-perf/README.md
```diff
@@ -631,6 +631,12 @@ exported to `<profile_export_file>_genai_perf.json` and
 export file is `profile_export.json`, the genai-perf file will be exported to
 `profile_export_genai_perf.csv`. (default: `profile_export.json`)
 
+### Session Options
+
+##### `--session-concurrency`
```
nv-hwoo, line 636:
```suggestion
##### `--session-concurrency <int>`
```

genai-perf/tests/integration_tests/test_multiturn.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @pytest.fixture
+    def retriever(self, mock_config):
+        return PayloadInputRetriever(mock_config)
+
+    def test_multi_turn_input__generation(self, monkeypatch, mock_config):
+        input_data = (
+            '{"text": "Turn 1", "session_id": "abc", "delay": 2}\n'
+            '{"text": "Turn 2", "session_id": "abc", "delay": 3}\n'
+            '{"text": "Turn 3", "session_id": "abc", "timestamp": 5000}\n'
```
lkomali, line 63: Is this functionally correct? 
The way multi-turn is implemented today, we can specify a mix of delay and timestamp and PA ignores a request if it doesn't contain "delay" in multi-turn mode.
However, I think GAP should throw an error in this case instead of PA silently ignoring the request. 
The test is fine in itself but I think this validation should be added in the feature.

templates/genai-perf-templates/README_template
```diff
@@ -631,6 +631,12 @@ exported to `<profile_export_file>_genai_perf.json` and
 export file is `profile_export.json`, the genai-perf file will be exported to
 `profile_export_genai_perf.csv`. (default: `profile_export.json`)
 
+### Session Options
+
+##### `--session-concurrency`
```
nv-hwoo, line 636:
```suggestion
##### `--session-concurrency <int>`
```

lkomali (commented):
LGTM, just one question/comment.

genai-perf/tests/integration_tests/test_multiturn.py
dyastremsky: I see. I updated the unit test. Thanks for pointing this out!

I think PA today returns the error and GAP forwards it to the user, but we could always run it and validate it.

templates/genai-perf-templates/README_template
dyastremsky: I was copying the changes to both READMEs. Just ran it -- no changes.

genai-perf/tests/integration_tests/test_multiturn.py
dyastremsky: I need to push back on this. The clarity is nice, but then you have a very brittle unit test. We see this in test_json_parser, which requires a change every time we modify the code. What I'd rather have is tests that minimally test only the functionality necessary and nothing else, even if they are slightly less readable.

Open to a longer discussion on this. Maybe as a team? I'd like us aligned on the tests we're writing, with folks agreeing on the rationale and approach for writing tests.

nv-hwoo: > We see this in test_json_parser, which requires a change every time we modify the code. 

Are you referring to profile data parser tests (esp. LLM profile data parser test)? If so, I believe the problem there is different one. The issues with the profile data parser test are that
1. It needs profile export json file to check the functionality.
2. Each functionality (e.g. OpenAI endpoints, Triton endpoints, etc) requires different profile export contents.
3. It's trying to test too many things (e.g. parse json file, calculate metrics, calculate statistics). One reason for this is due to how the current profile data parser has blown up over time.

> Open to a longer discussion on this. Maybe as a team? I'd like us aligned on our approach to writing tests and the rationale behind it.

Right, I agree this is probably worth discussing with the team as a whole. I'm sure there's some way that could strike the balance between clarity and flexibility.

dyastremsky: No, no. The JSON exporter tests. I believe it's [this one](https://github.com/triton-inference-server/perf_analyzer/blob/main/genai-perf/tests/test_exporters/test_json_exporter.py#L305). To be clear, I'm not trying to call out anyone's design or testing. I don't even know who wrote it at this point. It's just that whenever we add/remove an arg, that test immediately breaks. And that's because we [hardcore the entire JSON we expect](https://github.com/triton-inference-server/perf_analyzer/blob/59899c5f542df56e4357866eeb9a8283ec4caf58/genai-perf/tests/test_exporters/test_json_exporter.py#L141-L280) rather than testing specific functionalities.

One could make the argument that we want to test every field in the JSON, but I don't think that's the test intent. If it is, breaking up the test to individually test fields would be better. But a test that we constantly have to update results in more maintenance/headache and increases the likelihood a mistake gets overlooked because people are used to needing to fix it whenever it breaks with a new feature.

nv-hwoo: Ah, hahaha when I saw json "parser" I just immediately thought you meant data parser 😂 Anyways, this is not a huge blocker for me since the test is bitesize and straightforward enough. Thanks for explaining your thoughts!

nv-hwoo (approved):

genai-perf/tests/integration_tests/test_multiturn.py
dyastremsky: Ah haha! Any time, I appreciate you raising this and discussing it. I would love to talk about it offline sometime. I think we have done a good job on the testing infrastructure, but it could be nice to strategize as a team about it to iterate and make it even more frictionless.

PR #289, opened 2025-02-20
Add per-session metrics to the resulting JSON file

nv-hwoo:
Example exported JSON file with session statistics:
```
{
  "request_throughput": {
    "unit": "requests/sec",
    "avg": 2.2453037464000505
  },
  "request_latency": {
    "unit": "ms",
    "avg": 144.78743075,
    ...
  },
  ...,
  "input_config": {
    ...
  },
  "sessions": {
    "181df3d3-d00d-465f-bdb3-6363547563ea": {
      "request_throughput": {
        "unit": "requests/sec",
        "avg": 1.1226518732000252
      },
      "request_latency": {
        "unit": "ms",
        "avg": 140.6232415,
        ...
      },
      ...
    },
    "c8af0fc8-c266-4052-b8d8-7e7031f65e2d": {
      "request_throughput": {
        "unit": "requests/sec",
        "avg": 1.5406063077538594
      },
      "request_latency": {
        "unit": "ms",
        "avg": 148.95162,
        ...
      },
      ...
    }
  }
}
```

PR #302, opened 2025-02-27
Fix session ID missing from profile export JSON

matthewkotila:
This is to enable a new PR that will restore the intended functionality of [Add per-session metrics to the resulting JSON file #289](https://github.com/triton-inference-server/perf_analyzer/pull/289).

Before this PR, session ID was missing from the profile export JSON:

```json
{
  "experiments": [
    {
      "experiment": {
        "mode": "request_rate",
        "value": 0.0
      },
      "requests": [
        {
          "timestamp": 0,
          "request_inputs": {
            "payload": "my_request_body_1"
          },
          "response_timestamps": [
            1
          ],
          "response_outputs": [
            {
              "response": "my_response_body_1"
            }
          ]
        },
        {
          "timestamp": 2,
          "request_inputs": {
            "payload": "my_request_body_2"
          },
          "response_timestamps": [
            3
          ],
          "response_outputs": [
            {
              "response": "my_response_body_2"
            }
          ]
        }
      ],
      "window_boundaries": [
        0,
        18446744073709551615
      ]
    }
  ],
  "version": "0.0.0",
  "service_kind": "openai",
  "endpoint": "v1/chat/completions"
}
```

After this PR, session ID is included as an input next to the `payload` input in the profile export JSON:

```json
{
  "experiments": [
    {
      "experiment": {
        "mode": "request_rate",
        "value": 0.0
      },
      "requests": [
        {
          "timestamp": 0,
          "request_inputs": {
            "delay": 1,
            "session_id": "my_session_id",
            "payload": "my_request_body_1"
          },
          "response_timestamps": [
            1
          ],
          "response_outputs": [
            {
              "response": "my_response_body_1"
            }
          ]
        },
        {
          "timestamp": 2,
          "request_inputs": {
            "session_id": "my_session_id",
            "payload": "my_request_body_2"
          },
          "response_timestamps": [
            3
          ],
          "response_outputs": [
            {
              "response": "my_response_body_2"
            }
          ]
        }
      ],
      "window_boundaries": [
        0,
        18446744073709551615
      ]
    }
  ],
  "version": "0.0.0",
  "service_kind": "openai",
  "endpoint": "v1/chat/completions"
}
```

### DIFF
--- a/src/session_concurrency/payload_dataset_manager.cc
+++ b/src/session_concurrency/payload_dataset_manager.cc
@@ -30,7 +30,9 @@
 
 #include <chrono>
 #include <cstdint>
+#include <cstring>
 #include <memory>
+#include <optional>
 #include <stdexcept>
 #include <string>
 #include <unordered_map>
@@ -41,7 +43,6 @@
 #include "../data_loader.h"
 #include "../model_parser.h"
 #include "../tensor_data.h"
-#include "payload_json_utils.h"
 
 namespace triton::perfanalyzer {
 
@@ -73,8 +74,8 @@ PayloadDatasetManager::GroupPayloadsBySession() const
   return session_datasets;
 }
 
-std::chrono::milliseconds
-PayloadDatasetManager::GetDelayForPayload(size_t dataset_index) const
+std::optional<std::chrono::milliseconds>
+PayloadDatasetManager::GetDelay(size_t dataset_index) const
 {
   TensorData delay_ms_tensor_data{};
 
@@ -86,9 +87,7 @@ PayloadDatasetManager::GetDelayForPayload(size_t dataset_index) const
   }
 
   if (!delay_ms_tensor_data.data_ptr) {
-    throw std::runtime_error(
-        "Missing 'delay' input for payload with index " +
-        std::to_string(dataset_index));
+    return {};
   }
 
   const uint64_t delay_ms{
@@ -146,8 +145,12 @@ PayloadDatasetManager::GetSessionID(size_t dataset_index) const
     throw std::runtime_error(error.Message());
   }
 
-  const uint8_t* session_id_buffer{session_id_tensor_data.data_ptr};
-  const size_t session_id_byte_size{session_id_tensor_data.batch1_size};
+  uint32_t session_id_byte_size{};
+  std::memcpy(
+      &session_id_byte_size, session_id_tensor_data.data_ptr, sizeof(uint32_t));
+
+  const uint8_t* session_id_buffer{
+      session_id_tensor_data.data_ptr + sizeof(uint32_t)};
 
   const std::string session_id(
       reinterpret_cast<const char*>(session_id_buffer), session_id_byte_size);
--- a/src/session_concurrency/payload_dataset_manager.h
+++ b/src/session_concurrency/payload_dataset_manager.h
@@ -28,8 +28,8 @@
 #include <stddef.h>
 
 #include <chrono>
-#include <cstdint>
 #include <memory>
+#include <optional>
 #include <string>
 #include <unordered_map>
 #include <vector>
@@ -47,7 +47,9 @@ class PayloadDatasetManager {
 
   std::vector<std::vector<size_t>> GroupPayloadsBySession() const;
 
-  std::chrono::milliseconds GetDelayForPayload(size_t dataset_index) const;
+  std::string GetSessionID(size_t dataset_index) const;
+
+  std::optional<std::chrono::milliseconds> GetDelay(size_t dataset_index) const;
 
   std::string GetPayload(size_t dataset_index) const;
 
@@ -55,8 +57,6 @@ class PayloadDatasetManager {
   using PayloadsMapType = std::unordered_map<std::string, std::vector<size_t>>;
 
   PayloadsMapType CreateSessionIdToPayloadsMap() const;
-
-  std::string GetSessionID(size_t dataset_index) const;
 
   std::shared_ptr<const DataLoader> data_loader_{};
   const std::shared_ptr<ModelParser> parser_{};
--- a/src/session_concurrency/request_handler.cc
+++ b/src/session_concurrency/request_handler.cc
@@ -28,11 +28,11 @@
 
 #include <rapidjson/allocators.h>
 #include <rapidjson/document.h>
-#include <rapidjson/stringbuffer.h>
-#include <rapidjson/writer.h>
 #include <stddef.h>
 
+#include <chrono>
 #include <cstdint>
+#include <cstring>
 #include <functional>
 #include <future>
 #include <memory>
@@ -73,6 +73,10 @@ RequestHandler::SendRequestAndWaitForResponse(
 
   PayloadJsonUtils::UpdateHistoryAndAddToPayload(payload, chat_history);
 
+  auto& request_inputs{request_record.request_inputs_.emplace_back()};
+
+  RecordRequestInputs(payload, dataset_index, request_inputs);
+
   auto response_promise{std::make_shared<std::promise<void>>()};
   std::future<void> response_future{response_promise->get_future()};
 
@@ -80,6 +84,52 @@ RequestHandler::SendRequestAndWaitForResponse(
       payload, std::move(response_promise), chat_history, request_record);
 
   WaitForResponse(std::move(response_future));
+}
+
+void
+RequestHandler::RecordRequestInputs(
+    const std::string& payload, size_t dataset_index,
+    RequestRecord::RequestInput& request_inputs) const
+{
+  RecordPayloadInput(payload, request_inputs);
+
+  const auto session_id{payload_dataset_manager_->GetSessionID(dataset_index)};
+  RecordSessionIDInput(session_id, request_inputs);
+
+  const auto delay{payload_dataset_manager_->GetDelay(dataset_index)};
+  if (delay) {
+    RecordDelayInput(*delay, request_inputs);
+  }
+}
+
+void
+RequestHandler::RecordPayloadInput(
+    const std::string& payload,
+    RequestRecord::RequestInput& request_inputs) const
+{
+  std::vector<uint8_t> payload_buf(payload.begin(), payload.end());
+  request_inputs.emplace("payload", RecordData(std::move(payload_buf), "JSON"));
+}
+
+void
+RequestHandler::RecordSessionIDInput(
+    const std::string& session_id,
+    RequestRecord::RequestInput& request_inputs) const
+{
+  std::vector<uint8_t> session_id_buf(session_id.begin(), session_id.end());
+  request_inputs.emplace(
+      "session_id", RecordData(std::move(session_id_buf), "BYTES"));
+}
+
+void
+RequestHandler::RecordDelayInput(
+    const std::chrono::milliseconds delay,
+    RequestRecord::RequestInput& request_inputs) const
+{
+  const uint64_t delay_ms{static_cast<uint64_t>(delay.count())};
+  std::vector<uint8_t> delay_buf(sizeof(uint64_t));
+  std::memcpy(delay_buf.data(), &delay_ms, sizeof(uint64_t));
+  request_inputs.emplace("delay", RecordData(std::move(delay_buf), "UINT64"));
 }
 
 void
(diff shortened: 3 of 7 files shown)


src/session_concurrency/payload_dataset_manager.cc
```diff
@@ -30,7 +30,9 @@
 
 #include <chrono>
 #include <cstdint>
+#include <cstring>
 #include <memory>
+#include <optional>
 #include <stdexcept>
```
matthewkotila, line 36: Missing headers trying to adhere to include-what-you-use philosophy.

src/session_concurrency/payload_dataset_manager.cc
```diff
@@ -73,8 +74,8 @@ PayloadDatasetManager::GroupPayloadsBySession() const
   return session_datasets;
 }
 
-std::chrono::milliseconds
-PayloadDatasetManager::GetDelayForPayload(size_t dataset_index) const
+std::optional<std::chrono::milliseconds>
+PayloadDatasetManager::GetDelay(size_t dataset_index) const
```
matthewkotila, line 78: Part of this PR is fetching the delay along with other inputs (session ID, payload) _before_ the request gets sent to record for the profile export JSON.

GetDelay() previously was only called at a point in the code where it was correct to expect a delay input.

So consequently, GetDelay would throw an exception if the delay input was missing.

Now, in order to include the inputs in the profile export JSON, it was easier to have this function return an optional, to indicate whether there was a delay for a particular request.

A [different spot](https://github.com/triton-inference-server/perf_analyzer/pull/302/files#diff-364ae5c05536cf1350b7890278a35b571113c32da7ca5f465d503a78ac996031R168-R172) later in the code is used to check for the case where delay is expected and missing.

src/session_concurrency/payload_dataset_manager.cc
```diff
@@ -86,9 +87,7 @@ PayloadDatasetManager::GetDelayForPayload(size_t dataset_index) const
   }
 
   if (!delay_ms_tensor_data.data_ptr) {
-    throw std::runtime_error(
-        "Missing 'delay' input for payload with index " +
-        std::to_string(dataset_index));
+    return {};
```
matthewkotila, line 90: As mentioned, GetDelay() now returns an optional in the case that the delay is missing, rather than throwing an exception.

As mentioned, GetDelay is now called before the request is sent in order to record request inputs for the profile export JSON. I.e. it's called _before_ the code knows if there's a delay for a particular request.

It's easier to return an optional for both cases.

The new place where the code expects a delay and errors if it's missing is here:

https://github.com/triton-inference-server/perf_analyzer/pull/302/files#diff-364ae5c05536cf1350b7890278a35b571113c32da7ca5f465d503a78ac996031R168-R172

The new place where

src/session_concurrency/payload_dataset_manager.cc
```diff
@@ -146,8 +145,12 @@ PayloadDatasetManager::GetSessionID(size_t dataset_index) const
...
   }
 
-  const uint8_t* session_id_buffer{session_id_tensor_data.data_ptr};
-  const size_t session_id_byte_size{session_id_tensor_data.batch1_size};
+  uint32_t session_id_byte_size{};
+  std::memcpy(
+      &session_id_byte_size, session_id_tensor_data.data_ptr, sizeof(uint32_t));
+
+  const uint8_t* session_id_buffer{
+      session_id_tensor_data.data_ptr + sizeof(uint32_t)};
```
matthewkotila, line 153: This was kinda a bug fix.

Previously, the code was leaving in the first-4-bytes-are-the-length-of-the-rest-of-the-bytes thing.

Now, it's excluded, and the recorded session ID is just the session ID itself as typed in the input data JSON.

src/session_concurrency/payload_dataset_manager.h
```diff
@@ -56,8 +58,6 @@ class PayloadDatasetManager {
 
   PayloadsMapType CreateSessionIdToPayloadsMap() const;
```
matthewkotila, line 60: `GetSessionID()` is moved from private to public since it is now called in the place where the session ID input is recorded for the profile export JSON.

src/session_concurrency/request_handler.cc
```diff
@@ -73,6 +73,10 @@ RequestHandler::SendRequestAndWaitForResponse(
 
   PayloadJsonUtils::UpdateHistoryAndAddToPayload(payload, chat_history);
 
+  auto& request_inputs{request_record.request_inputs_.emplace_back()};
+
+  RecordRequestInputs(payload, dataset_index, request_inputs);
```
matthewkotila, line 78: Request inputs are now recorded slightly earlier than they were before so that access to the other inputs (session ID, delay) are easier.

src/session_concurrency/request_handler.cc
```diff
@@ -82,6 +86,52 @@ RequestHandler::SendRequestAndWaitForResponse(
...
+{
+  RecordPayloadInput(payload, request_inputs);
+
+  const auto session_id{payload_dataset_manager_->GetSessionID(dataset_index)};
+  RecordSessionIDInput(session_id, request_inputs);
+
+  const auto delay{payload_dataset_manager_->GetDelay(dataset_index)};
+  if (delay) {
+    RecordDelayInput(*delay, request_inputs);
+  }
```
matthewkotila, line 102: As mentioned, delay is only recorded if it exists.

src/session_concurrency/request_handler.cc
```diff
@@ -98,7 +148,7 @@ RequestHandler::SendRequest(
 
   const auto inputs{PrepareInputs(payload)};
 
-  RecordRequest(inputs, request_record);
+  request_record.start_time_ = std::chrono::system_clock::now();
```
matthewkotila, line 151: As mentioned [here](https://github.com/triton-inference-server/perf_analyzer/pull/302/files#r1972733350), request inputs are recorded slightly earlier, but we still record the start time as close to the client backend call as we can, here.

src/session_concurrency/request_handler.cc
```diff
@@ -170,9 +220,7 @@ RequestHandler::RecordResponse(
 
   request_record.response_timestamps_.push_back(end_time);
 
-  request_record.response_outputs_.emplace_back();
-
-  auto& response_outputs{request_record.response_outputs_.back()};
+  auto& response_outputs{request_record.response_outputs_.emplace_back()};
```
matthewkotila, line 223: A little cleanup. Since c++17, `emplace_back()` for vector actually returns a reference to the newly constructed.

src/session_concurrency/request_handler.cc
```diff
@@ -239,38 +287,6 @@ RequestHandler::PrepareInputs(const std::string& payload) const
...
-    size_t byte_size{0};
-    input->RawData(&buf, &byte_size);
-
-    std::vector<uint8_t> buf_vec(buf, buf + byte_size);
-
-    request_inputs.emplace(
-        input->Name(), RecordData(std::move(buf_vec), data_type));
-  }
-}
-
```
matthewkotila, line 273: As mentioned, this code got moved to an earlier place in the flow.

src/session_concurrency/session_concurrency_manager.cc
```diff
@@ -141,31 +142,36 @@ SessionConcurrencyManager::SendSequentialRequestsForOneSession(
   rapidjson::Document chat_history(rapidjson::kArrayType);
 
   for (size_t i{0}; i < one_session_payloads.size(); ++i) {
-    const size_t payload_index{one_session_payloads[i]};
+    const size_t payload_dataset_index{one_session_payloads[i]};
```
matthewkotila, line 145: Little cleanup. Seemed like a more clear variable name.

src/session_concurrency/session_concurrency_manager.cc
```diff
@@ -141,31 +142,36 @@ SessionConcurrencyManager::SendSequentialRequestsForOneSession(
...
+    size_t payload_dataset_index) const
 {
-  const auto delay{payload_dataset_manager_->GetDelayForPayload(payload_index)};
+  const auto delay{payload_dataset_manager_->GetDelay(payload_dataset_index)};
+
+  if (!delay) {
+    throw std::runtime_error(
+        "Missing 'delay' input for payload with index " +
+        std::to_string(payload_dataset_index));
+  }
```
matthewkotila, line 172: As mentioned, since GetDelay() in PayloadDatasetManager can now return optional for missing delay, this is the new place where an expected delay that's missing causes an exception to be thrown.

src/session_concurrency/payload_dataset_manager.h
```diff
@@ -56,8 +58,6 @@ class PayloadDatasetManager {
...
 
-  std::string GetSessionID(size_t dataset_index) const;
-
```
matthewkotila, line 60: `GetSessionID()` is moved from private to public since it is now called in the place outside this class where the session ID input is recorded for the profile export JSON.

src/test_payload_dataset_manager.cc
```diff
@@ -217,31 +276,48 @@ TEST_CASE("PayloadDatasetManager::GetPayload")
...
+    const auto error{data_loader->ReadDataFromJSON(
+        model_parser->Inputs(), model_parser->Outputs(), input_data_json)};
 
-  const auto error{data_loader->ReadDataFromJSON(
-      model_parser->Inputs(), model_parser->Outputs(), input_data_json)};
+    REQUIRE(error.IsOk());
 
-  REQUIRE(error.IsOk());
+    const PayloadDatasetManager payload_dataset_manager(
+        data_loader, model_parser);
```
nv-hwoo, line 299: I think we can take this out of subcase?

src/session_concurrency/payload_dataset_manager.cc
nv-hwoo: Where did `session_id_tensor_data.batch1_size` go?

src/session_concurrency/request_handler.cc
nv-hwoo: qq: where is `SendRequest` being called? I only see `SendRequestAndWaitForResponse` being called.

src/test_payload_dataset_manager.cc
```diff
@@ -134,7 +134,66 @@ TEST_CASE("PayloadDatasetManager::GroupPayloadsBySession")
...
+    const auto session_id{payload_dataset_manager.GetSessionID(dataset_index)};
+
+    CHECK(session_id == "my_session_id");
+  }
+
+  SUBCASE("missing session ID")
+  {
+    auto data_loader{std::make_shared<MockDataLoader>()};
+
+    const PayloadDatasetManager payload_dataset_manager(
```
nv-hwoo, line 185: can this be moved out of subcase?

src/session_concurrency/payload_dataset_manager.cc
matthewkotila: It's not needed. It's always going to be equal to whatever integer value the first 4 bytes hold plus 4.

src/session_concurrency/request_handler.cc
matthewkotila: SendRequest is called within SendRequestAndWaitForResponse:

https://github.com/triton-inference-server/perf_analyzer/pull/302/files#diff-f51c753ff7d70252c18dafad999cf7ee58ac73fdedf113e66005f913d276c69fR83-R84

src/test_payload_dataset_manager.cc
matthewkotila: This subcase is testing the valid path through the public member method GetPayload(), so it is needed.

matthewkotila: This takes the data loader as input, which takes the custom JSON string for this particular subcase, so I think it's more sensible to keep it in the subcase.

src/session_concurrency/request_handler.cc
nv-hwoo: Woops. Totally missed that haha 👍

src/session_concurrency/payload_dataset_manager.cc
nv-hwoo: Ah gotcha. You're just getting the size from the first 4 bytes. Makes sense.

src/test_payload_dataset_manager.cc
nv-hwoo: Sure, I only pointed it out since it seemed like we could make the subcase more minimal but I'm fine if you feel strongly about keeping it this way for now.

src/test_payload_dataset_manager.cc
nv-hwoo: Keeping it as is for now.

nv-hwoo (approved):

PR #304, opened 2025-02-27
Enable generating empty prompts

mwawrzos:
When the genai-perf is called with such arguments:
```bash
genai-perf profile \
    --endpoint-type vision \
    --synthetic-input-tokens-mean 0 \
    --synthetic-input-tokens-stddev 0 \
    --image-width-mean 2048 \
    --image-width-stddev 0 \
    --image-height-mean 1648 \
    --image-height-stddev 0 \
    --image-format png \
    ...
```
The genai-perf still produces 1-token prompts.
This MR fixes this issue.

### DIFF
--- a/genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
+++ b/genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
@@ -61,7 +61,7 @@ class SyntheticPromptGenerator:
             )
 
         num_prompt_tokens = max(
-            1, int(random.gauss(prompt_tokens_mean, prompt_tokens_stddev))
+            0, int(random.gauss(prompt_tokens_mean, prompt_tokens_stddev))
         )
 
         return cls._generate_prompt(tokenizer, num_prompt_tokens)


genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
matthewkotila, line 1: This might be a naive question, but what's the value-add of benchmarking with empty prompts?

Is this a common LLM user scenario?

Generally, the mission of GenAI-Perf is to simulate real-world user scenarios regarding generative AI/LLMs, and benchmark the performance of those simulated scenarios.

mwawrzos: With this setting the prompt, but still the image can be generated - let me add this to the description, as my example looks confusing without this.
We are currently testing with genai-perf a model that uses OpenAI API, but has specific constraints.
The model accepts only inputs with image_url input, and the prompt part must be empty or missing.

matthewkotila: Ah gotcha--so some other input is used--makes sense.

debermudez:
I have to double check the unit tests, then I can approve.

dyastremsky:
@nv-hwoo originally wrote that code, so want to tag him to ensure he didn't see anything previously that would be an issue with setting it to 0 instead of 1.

@debermudez, someone needs to run the CI for these changes as well. And ideally write unit tests since 0 is an edge case that can create problems.

nv-hwoo:
@dyastremsky I'll run the CI / add unit test for this

nv-hwoo:
Mirror PR (#305) has been merged. Closing this PR.

PR #306, opened 2025-02-28
Add synthetic support for multi-turn

dyastremsky:
Add synthetic support for multi-turn benchmarking

**Example run commands**

Start a server:
```docker run -it --net=host --rm --gpus=all vllm/vllm-openai:latest --model TinyLlama/TinyLlama-1.1B-Chat-v1.0  --dtype float16 --max-model-len 1024```

**Run GenAI-Perf:**
```genai-perf profile -m TinyLlama/TinyLlama-1.1B-Chat-v1.0 --service-kind openai --endpoint-type chat --num-sessions 10 --session-concurrency 5 --session-turns-mean 2 --session-turns-stddev 0 --session-turn-delay-mean 1000 --session-turn-delay-stddev 5 --synthetic-input-tokens-mean 50 --tokenizer TinyLlama/TinyLlama-1.1B-Chat-v1.0 --output-tokens-mean 50 --num-prefix-prompts 3 --prefix-prompt-length 15```

--session-concurrency enables multi-turn benchmarking
--session-turn-delay-X args set the turn delay distribution (mean, stddev)
--session-turns-X args set the turn distribution (mean, stddev)
--num-prefix-prompts and --prefix-prompt-length set the system prompt parameters. The number of prefix prompts determines the size of the prefix prompt pool. The prefix prompts are only prepended for the first turn in a session.

**Screenshot:**
image

### CODE REVIEW

**Comment by dyastremsky** [comment]
I checked that both multi-turn and stateless modes run. In multi-turn, I have briefly validated the timestamps for the turns.

We recently added the new config files, so I'll need to figure out how those work to add these in the right place and fix the broken config unit test. (CC: @nv-braf) Otherwise, this pull request is ready to polish, review, and merge. I'll add CI testing separately.

**Comment by dyastremsky** [comment]
> I checked that both multi-turn and stateless modes run. In multi-turn, I have briefly validated the timestamps for the turns.
> 
> We recently added the new config files, so I'll need to figure out how those work to add these in the right place and fix the broken config unit test. (CC: @nv-braf) Otherwise, this pull request is ready to polish, review, and merge. I'll add CI testing separately.

It looks like it should go into [this file](https://github.com/triton-inference-server/perf_analyzer/blob/main/genai-perf/genai_perf/config/input/config_command.py) to add it to the config, but I think it'll require some thought into which categories to put these into. Or whether to create a ConfigSessions dataclass.

@nv-braf, I'll sync with you offline tomorrow. We want to get this in soon. It might generally also be good to create a short guide for folks on how we can add new args with the new config option.

File: genai-perf/genai_perf/parser.py [hunk 1]
```diff
@@ -1056,11 +1066,50 @@ def _add_profile_args(parser):
 def _add_session_args(parser):
     session_group = parser.add_argument_group("Session")
 
+    session_load_management_group = session_group.add_mutually_exclusive_group(
```
dyastremsky, line 1069: I might create a separate load management category, since there should only be one of --request-concurrency, --request-rate, --session-concurrency, and --session-rate ever selected. That will improve documentation and error checking, plus it will be more native to ArgParse.

genai-perf/genai_perf/inputs/retrievers/synthetic_data_retriever.py
```diff
@@ -40,57 +41,108 @@
...
-                    row.texts.append(prompt)
-
-                for _ in range(self.config.batch_size_image):
-                    image = SyntheticImageGenerator.create_synthetic_image(
-                        image_width_mean=self.config.image_width_mean,
-                        image_width_stddev=self.config.image_width_stddev,
-                        image_height_mean=self.config.image_height_mean,
-                        image_height_stddev=self.config.image_height_stddev,
-                        image_format=self.config.image_format,
+            data_rows = self._generate_data_rows(use_prefix_prompts)
```
debermudez, line 57: This reads really nicely. Nice clean up here.

genai-perf/genai_perf/parser.py
```diff
@@ -300,6 +300,16 @@ def _check_goodput_args(args):
     return args
 
 
+def _check_session_args(args):
+
+    if args.num_sessions:
+        if args.session_concurrency is None:
```
debermudez, line 306: Why not default to 1 for `session_concurrency`?
Trying to fail early to notify the user?

genai-perf/genai_perf/parser.py
debermudez: Agree 100%. Lets get this in, but we will need to expand the load_management groups soon.

genai-perf/genai_perf/parser.py
```diff
@@ -1056,11 +1066,51 @@ def _add_profile_args(parser):
...
+    )
+
+    session_group.add_argument(
+        "--session-turns-mean",
+        type=int,
+        default=ic.DEFAULT_SESSION_TURNS_MEAN,
+        help="The mean number of turns per session.",
+    )
+
+    session_group.add_argument(
```
debermudez, line 1109: You kept them alphabetized. 👨🏼‍🍳 💋

genai-perf/genai_perf/parser.py
dyastremsky: Great question! I want to avoid default logic right now, because we might end up requiring this anyway after Brian's config PR goes through. If we create a mutually-exclusive load management group, then one of them will be required. So I'm trying to avoid instituting something that will break in the next couple of weeks.

debermudez: Got it. That works for me.

debermudez (approved):

genai-perf/README.md
```diff
@@ -633,9 +633,35 @@ export file is `profile_export.json`, the genai-perf file will be exported to
...
 
+##### `--num-sessions`
+
+The number of sessions to simulate. This is used when generating synthetic
+session data. (default: `0`)
+
 ##### `--session-concurrency <int>`
 
-The number of concurrenct sessions to benchmark. (default: `None`)
+The number of concurrent sessions to benchmark. This must be specified
```
rmccorm4, line 643: Why not reuse --concurrency arg instead of adding new --session-concurrency?

dyastremsky: This is how Perf Analyzer does it, so we need to maintain parity across. I think this is a good point, but right now, without this arg, PA (and GAP) would have no way of the user is benchmarking in sessions mode unless we add another argument (e.g. -`--session-mode`).

Tagging @matthewkotila, who may be able to speak to the rationale on the PA side and whether this is a good future direction.

rmccorm4: num_sessions >= 1 or num_session_turns >=2 would also imply session / multi-turn mode right?

dyastremsky: Yes. But there would be no way to know that benchmarking mode is session-based for the new payload retriever (`--input-file payload:<file_path>`). It can be used for fixed schedule or multi-turn. It allows the user to provide data for specific payloads.

dyastremsky: GAP needs to pass PA --fixed-schedule for fixed schedule benchmarking and --session-concurrency for session benchmarking.

PR #312, opened 2025-03-04
Improve schedule delay UX by adding delay tolerance

matthewkotila:


src/request_rate_worker.cc
```diff
@@ -136,7 +137,8 @@ RequestRateWorker::SleepIfNecessary()
   std::chrono::nanoseconds wait_time = next_timestamp - current_timestamp;
 
   bool delayed = false;
-  if (wait_time.count() < 0) {
+  if (wait_time < std::chrono::nanoseconds::zero() &&
+      std::chrono::abs(wait_time) > delay_tolerance_) {
```
matthewkotila, line 141: This function determines if an about-to-be-sent request is late/delayed.

A request is considered late if the current time since the start of the benchmark is past the request's intended send timestamp.

Because of how the scheduling logic is written, the "start time" of the benchmark is recorded slightly before this function.

This naturally causes an expected small delay since computers are not instantaneous.

Before this PR, if a user specified that they wanted a request to send at timestamp 0, PA would anecdotally/empirically take around 0.1-0.8 milliseconds to go from the [place](https://github.com/triton-inference-server/perf_analyzer/blob/f51dcdb3332946df462c5d5106620013d03f16db/src/request_rate_manager.cc#L283) where the timestamp for the start of the benchmark is recorded to this spot in the code. Again, this is just due to computers not being instantaneous.

The proposed solution is to implement a delay tolerance.

I figured 1ms delay tolerance was fine. For the fixed schedule mode, 1ms is the smallest granularity they can use for the provided schedule, anyway.

src/request_rate_worker.h
```diff
@@ -115,6 +116,8 @@ class RequestRateWorker : public LoadWorker, public IScheduler {
     ctx->SetNumActiveThreads(num_threads_);
   }
 
+  const std::chrono::milliseconds delay_tolerance_{1};
```
matthewkotila, line 119: This could eventually be a user option, but I want to keep it fixed until someone asks otherwise.

dyastremsky (approved):
Great work!

PR #315, opened 2025-03-05
Fix hash blocks generation for the payload retriever

dyastremsky:
Previously, when using the payload retriever, the hash blocks would generate full blocks for every hash ID. On the last one, it would check the number of tokens left. This could lead to some hash blocks being less than the block size. It could lead to issues like this:

image

This PR changes the code to generate a full hash block for every hash ID. If there are still more tokens to get, then get that number of tokens from the hash block.

After changes:
image

### DIFF
--- a/genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
+++ b/genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
@@ -119,7 +119,6 @@ class SyntheticPromptGenerator:
             )
 
         start_idx = random.randrange(cls._corpus_length)
-
         end_idx = start_idx + num_tokens
         prompt_tokens = cls._tokenized_corpus[start_idx:end_idx]
         if end_idx > cls._corpus_length:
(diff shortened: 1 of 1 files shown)


genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
```diff
@@ -175,17 +174,25 @@ def _generate_prompt_with_token_reuse(
...
+                # To ensure that the prompt doesn't merge chunks, we pop the
+                # last token and insert the bos token at the beginning. Length
+                # is maintained and the prompt generates the expected number
+                # of tokens.
                 prompt_tokens = cls._generate_prompt_tokens(size_to_use)
                 prompt_tokens.pop(-1)
                 prompt_tokens.insert(0, tokenizer.bos_token_id())
                 cls._cache[hash_index] = prompt_tokens
-            final_prompt.extend(cls._cache[hash_index])
+            if num_tokens < 0:
```
matthewkotila, line 186: When would this happen, tho?

Would this still produce an error causing L0_genai_perf_fixed_schedule to fail? Just with a different message?

dyastremsky: This should never happen. I'm adding it as a safeguard. I just updated the message to hopefully make this clearer.

genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
```diff
@@ -175,17 +174,25 @@ def _generate_prompt_with_token_reuse(
...
+                # of tokens.
                 prompt_tokens = cls._generate_prompt_tokens(size_to_use)
                 prompt_tokens.pop(-1)
                 prompt_tokens.insert(0, tokenizer.bos_token_id())
                 cls._cache[hash_index] = prompt_tokens
-            final_prompt.extend(cls._cache[hash_index])
+            if num_tokens <= 0:
+                raise ValueError(
+                    "Internal error: number of tokens should never be negative."
+                )
```
rmccorm4, line 189: nit: error message says negative, but if statement includes zero case

dyastremsky: Thanks for catching this. I was updating the error message in response to feedback above and accidentally added this.

genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
matthewkotila: How does this PR avoid producing an error in a test?

You've added two if statements that cover all set of possible cases, with the bad case still producing it's own error.

matthewkotila: Oh wait, the special sauce is the fact that

```python
            final_prompt.extend(cls._cache[hash_index])
```

lives inside of the good if-case now instead of always happening--is that why this fixes the test?

dyastremsky: I aimed to make it clean. :D Compared to before, a few things are happening things:
- Every hash block gets a full block allocated to it.
- After each block gets created, we add the block (or the remaining tokens, whichever is smaller) to the prompt, then update the number of remaining tokens.
- If there are no more tokens requested, then we don't append them to the prompt.

matthewkotila (approved):

dyastremsky:
Thank you @matthewkotila and @rmccorm4 for the reviews. Closing in favor of the new PR https://github.com/triton-inference-server/perf_analyzer/pull/316/files, which is aligned with requirements that have been communicated (specifically, that the input_length must be equal to the number of hash_blocks - 1 + (1-hash_size)).

PR #316, opened 2025-03-05
Add assert for invalid payload file

debermudez:
Add assertion to check payload file inputs.

### DIFF
--- a/genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
+++ b/genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
@@ -176,6 +176,10 @@ class SyntheticPromptGenerator:
         """
         final_prompt: List[int] = []
         size_to_use = block_size
+        last_hash_length = num_tokens - ((len(prompt_hash_list) - 1) * block_size)
+        assert (
+            last_hash_length > 0 and last_hash_length <= block_size
+        ), f"Input_length: {num_tokens}, Hash_ids: {prompt_hash_list}, Block_size: {block_size} are not compatible."
         for index, hash_index in enumerate(prompt_hash_list):
             if index == len(prompt_hash_list) - 1:
                 size_to_use = num_tokens - (index * block_size)
--- a/genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
+++ b/genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
@@ -177,9 +177,11 @@ class SyntheticPromptGenerator:
         final_prompt: List[int] = []
         size_to_use = block_size
         last_hash_length = num_tokens - ((len(prompt_hash_list) - 1) * block_size)
-        assert (
-            last_hash_length > 0 and last_hash_length <= block_size
-        ), f"Input_length: {num_tokens}, Hash_ids: {prompt_hash_list}, Block_size: {block_size} are not compatible."
+        assert 0 < last_hash_length <= block_size, (
+            f"Input_length: {num_tokens}, Hash_ids: {prompt_hash_list}, Block_size: {block_size} "
+            f"are not compatible. The final hash id length: {last_hash_length} must be greater "
+            f"than 0 and less than or equal to {block_size}."
+        )
         for index, hash_index in enumerate(prompt_hash_list):
             if index == len(prompt_hash_list) - 1:
                 size_to_use = num_tokens - (index * block_size)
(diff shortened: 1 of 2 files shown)


genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
```diff
@@ -174,6 +174,10 @@ def _generate_prompt_with_token_reuse(
         """
         final_prompt: List[int] = []
         size_to_use = block_size
+        last_hash_length = num_tokens - ((len(prompt_hash_list) - 1) * block_size)
+        assert (
+            last_hash_length > 0 and last_hash_length <= block_size
+        ), f"Input_length: {num_tokens}, Hash_ids: {prompt_hash_list}, Block_size: {block_size} are not compatible."
```
dyastremsky, line 180: Looks good. Can you update this error test to specify the expected input_length range? And to make it clear what the formula for that range is?

genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
matthewkotila, line 1: Is L0_genai_perf_fixed_schedule intended to have an error/exception case, then?

If so, will there be a paired gitlab MR for that?

debermudez: yep. here: https://gitlab-master.nvidia.com/dl/triton/perf_analyzer_ci/-/merge_requests/73

genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
debermudez: updated

genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
```diff
@@ -174,6 +174,12 @@ def _generate_prompt_with_token_reuse(
         """
         final_prompt: List[int] = []
         size_to_use = block_size
+        last_hash_length = num_tokens - ((len(prompt_hash_list) - 1) * block_size)
+        assert 0 < last_hash_length <= block_size, (
+            f"Input_length: {num_tokens}, Hash_ids: {prompt_hash_list}, Block_size: {block_size} "
```
dyastremsky, line 179: Could you make this error message tell the user what the accept input range is for the number of hash blocks they're requesting? And how that is calculated? We want this to be intuitive so that users know what the error is when they get it.

Also, we don't really use asserts in our code. We raise errors that gets caught by GenAI-Perf in main. Can you do it that way or is there a specific reason to diverge and use an assert here?

matthewkotila (dismissed):

genai-perf/genai_perf/inputs/retrievers/synthetic_prompt_generator.py
debermudez: Updated to make it an exception.

genai-perf/tests/test_retrievers/test_synthetic_prompt_generator.py
```diff
@@ -49,10 +53,30 @@ def test_synthetic_prompt_zero_token(self):
...
             tokenizer=tokenizer,
-            prompt_tokens_mean=123,
+            prompt_tokens_mean=prompt_tokens,
             prompt_tokens_stddev=0,
         )
+        assert len(tokenizer.encode(prompt)) <= 123 + tolerance
+        assert len(tokenizer.encode(prompt)) >= 123 - tolerance
 
-        assert len(tokenizer.encode(prompt)) == 123
+    @pytest.mark.parametrize(
```
debermudez, line 66: @matthewkotila i ended up adding the unit testing here

genai-perf/tests/test_retrievers/test_synthetic_prompt_generator.py
```diff
@@ -49,10 +53,30 @@ def test_synthetic_prompt_zero_token(self):
...
+        prompt_tokens = 123
+        tolerance = 2
         prompt = SyntheticPromptGenerator.create_synthetic_prompt(
             tokenizer=tokenizer,
-            prompt_tokens_mean=123,
+            prompt_tokens_mean=prompt_tokens,
             prompt_tokens_stddev=0,
         )
+        assert len(tokenizer.encode(prompt)) <= 123 + tolerance
+        assert len(tokenizer.encode(prompt)) >= 123 - tolerance
```
debermudez, line 64: boyscout update.
it failed once in all my testing.
i updated it to be a small range

dyastremsky (approved):

PR #319, opened 2025-03-05
Improve input sequence length accuracy for hash ID blocks

dyastremsky:
Filter corpus to include only alphanumeric characters with one space between words. In testing against a large dataset, this proved to work for combining hash ID blocks in the synthetic prompt retriever.

This keeps our corpus as representative as possible without resulting in unexpected token counts when concatenating texts (allowing punctuation or other characters still results in token counts being off).


debermudez (commented):
This is consistent with what was discussed earlier. Waiting on final approval from @IzzyPutterman

dyastremsky:
Closing. Looks like the current main branch does better on the larger dataset.

debermudez (dismissed):

nicolasnoble (approved):

PR #323, opened 2025-03-08
3x faster default synthetic text data generation

nv-hwoo:
By default, we are generating synthetic images along with synthetic prompts even when they are not used. This is negligible when the number of data entries is small (<1000) but quickly becomes noticable when the data entries become large (10k+).

With the change, we see almost **3x faster** synthetic data generation when the number of data entries is set to 10k.
```bash
# before
time genai-perf profile -m model--service-kind openai --endpoint-type chat --num-dataset-entries 10000

real    0m55.109s
user    1m1.694s
sys     0m4.018s

# after
time genai-perf profile -m model --service-kind openai --endpoint-type chat --num-dataset-entries 10000

real    0m18.526s
user    0m25.557s
sys     0m3.508s
```


genai-perf/genai_perf/inputs/retrievers/synthetic_data_retriever.py
```diff
@@ -105,44 +112,49 @@ def _generate_stateless_entries(self, use_prefix_prompts: bool) -> List[DataRow]
...
+            row.images = self._generate_images()
             data_rows.append(row)
 
         return data_rows
 
-    def _create_data_row(
-        self, use_prefix_prompts: bool, session_id: str = ""
-    ) -> DataRow:
-        row = DataRow(texts=[], images=[], optional_data={}, payload_metadata={})
+    def _create_data_row(self, session_id: str = "") -> DataRow:
```
debermudez, line 122: why did you take the work out of create_data_row()?

genai-perf/genai_perf/inputs/retrievers/synthetic_data_retriever.py
```diff
@@ -105,44 +112,49 @@ def _generate_stateless_entries(self, use_prefix_prompts: bool) -> List[DataRow]
...
-        return SyntheticImageGenerator.create_synthetic_image(
-            image_width_mean=self.config.image_width_mean,
-            image_width_stddev=self.config.image_width_stddev,
-            image_height_mean=self.config.image_height_mean,
-            image_height_stddev=self.config.image_height_stddev,
-            image_format=self.config.image_format,
-        )
+    def _generate_prompts(self, use_prefix_prompts: bool) -> List[str]:
+        prompts = []
+        for _ in range(self.config.batch_size_text):
```
debermudez, line 131: why did you unfold the list comprehension into a for loop?

genai-perf/genai_perf/inputs/retrievers/synthetic_data_retriever.py
```diff
@@ -105,44 +112,49 @@ def _generate_stateless_entries(self, use_prefix_prompts: bool) -> List[DataRow]
...
+
+            prompts.append(prompt)
+        return prompts
+
+    def _generate_images(self) -> List[str]:
+        """
+        Generate synthetic images if the image width and height are specified.
+        """
+        images = []
+        if self._include_image:
```
debermudez, line 149: i am not sure so i want your opinion here.
why do the if check here as opposed to outside the method call in _generate_prompts()

debermudez (commented):
mostly questions. 
Code is solid

genai-perf/genai_perf/inputs/retrievers/synthetic_data_retriever.py
nv-hwoo: Because the function name and it's actual behavior has a mismatch. It says create the data row but it's actually creating data row _AND_ generating prompts.

genai-perf/genai_perf/inputs/retrievers/synthetic_data_retriever.py
nv-hwoo: We have multiple places that needs this functionality (since I took it out of `_create_data_row`). It's easier & more readable to call it as a function than list comprehensions.

genai-perf/genai_perf/inputs/retrievers/synthetic_data_retriever.py
nv-hwoo: I just think it's easier to avoid mistakes when the callee takes care of these simple checks. Let's say in the future, we want to call `self._generate_images` in `_generate_multi_turn_sessions`, or somewhere else in the code. Every time we call this function, the caller doesn't need to remember to add `if` statement to avoid unnecessary image data generation. Same for audio data generation once we have one.

genai-perf/genai_perf/inputs/retrievers/synthetic_data_retriever.py
debermudez: I personally agree that for loops are easier to read. However, list comprehensions are more pythonic, so we should get used to reading them, if only to ensure they are an option for us and we dismiss them out of hand.

If you are using the underlying data in multiple locations, i agree the for loop in a method makes more sense.

genai-perf/genai_perf/inputs/retrievers/synthetic_data_retriever.py
debermudez: I think you are highlighting the tradeoffs well. 
I went back and forth on whether the if should be in the caller or callee. There are reasons for either approach.
Ultimately, i think its a style conclusion, and we should just try to remain consistent.

debermudez (approved):

PR #330, opened 2025-03-12
Add a note around pip dependency failures

debermudez:
After a bunch of testing and trying to resolve pip dependency errors, it's actually safe to ignore the pip dependency error that pops up during this tutorial. That can be misleading for users. 
Adding a note to avoid users spending cycles debugging the failures.


genai-perf/README.md
matthewkotila, line 1: Just curious, can you paste the error in this PR for future knowledge-banking?

debermudez:
```ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
grpcio-tools 1.64.0 requires protobuf<6.0dev,>=5.26.1, but you have protobuf 4.25.6 which is incompatible.```

matthewkotila (approved):

PR #333, opened 2025-03-14
Add CLI options for process-export-files subcommand

lkomali:
This PR adds support for relevant CLI options for process-export-files subcommand.

- Users can provide path to directory containing profile export files using `genai-perf profile-export-files <path-to-directory>`
image

- Not specifying a path to directory will throw an error
image

- The list of options supported in this subcommand - `genai-perf process-export-files -h`
image

### DIFF
--- a/genai-perf/genai_perf/parser.py
+++ b/genai-perf/genai_perf/parser.py
@@ -509,11 +509,6 @@ def _set_artifact_paths(args: argparse.Namespace) -> argparse.Namespace:
     return args
 
 
-def _set_profile_export_file_path(args: argparse.Namespace) -> argparse.Namespace:
-    args.profile_export_file = args.artifact_dir / args.profile_export_file
-    return args
-
-
 def parse_goodput(values):
     constraints = {}
     try:
@@ -1298,7 +1293,7 @@ def refine_args(
     elif args.subcommand == Subcommand.COMPARE.to_lowercase():
         args = _check_compare_args(parser, args)
     elif args.subcommand == Subcommand.PROCESS_EXPORT_FILES.to_cli_format():
-        args = _set_profile_export_file_path(args)
+        pass
     else:
         raise ValueError(f"Unknown subcommand: {args.subcommand}")
 
--- a/genai-perf/genai_perf/parser.py
+++ b/genai-perf/genai_perf/parser.py
@@ -509,6 +509,11 @@ def _set_artifact_paths(args: argparse.Namespace) -> argparse.Namespace:
     return args
 
 
+def _set_profile_export_file_path(args: argparse.Namespace) -> argparse.Namespace:
+    args.profile_export_file = args.artifact_dir / args.profile_export_file
+    return args
+
+
 def parse_goodput(values):
     constraints = {}
     try:
@@ -574,6 +579,13 @@ def _convert_str_to_enum_entry(args, option, enum):
 
 
 ### Types ###
+
+
+def directory(value: str) -> Path:
+    path = Path(value)
+    if path.is_dir():
+        return path
+    raise ValueError(f"'{value}' is not a valid directory")
 
 
 def file_or_directory(value: str) -> Path:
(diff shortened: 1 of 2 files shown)


genai-perf/genai_perf/parser.py
```diff
@@ -509,6 +509,11 @@ def _set_artifact_paths(args: argparse.Namespace) -> argparse.Namespace:
     return args
 
 
+def _set_profile_export_file_path(args: argparse.Namespace) -> argparse.Namespace:
```
nv-braf, line 512: I would add a comment here explaining why you are doing this and change the method name to be more descriptive (maybe prepend_artifact_directory_to_profile_export_file_path?).

This logic already exists in our code (currently in wrapper.py, moving to perf_analzyer_config.py with my config work). 

I think you need to do this because you won't be calling PA? But, the more I think about this, I believe this method does not belong in parser. You should pass in both the artifact_dir and profile_export_file to the new subcommand and have the combination logic exist there.

The only things that should be occurring in parser.py is the setting/checking of user values (from the CLI) and the inferring of implicit values (based on CLI inputs). It's bad practice to mutate a user value as this is not what someone who's not familiar with the code would expect and it could be confusing when trying to debug when the value in the variable doesn't match the value on the CLI.

genai-perf/genai_perf/parser.py
```diff
@@ -1025,6 +1037,16 @@ def _add_output_args(parser):
     )
 
 
+def _add_process_export_files_args(parser):
+    process_export_files_group = parser.add_argument_group("Process Export Files")
+    process_export_files_group.add_argument(
+        "input_path",
+        nargs=1,
+        type=directory,
+        help="The path to the directory containing the profile export files.",
```
nv-hwoo, line 1046: qq: What if the profile export files are distributed across multiple directories? Does user need to manually move all the profile export files into one?

genai-perf/genai_perf/parser.py
lkomali: Profile subcommand is doing this in parser.py [here](https://github.com/triton-inference-server/perf_analyzer/blob/abb1578943377ff8dc26e0d2fb704037beda97cb/genai-perf/genai_perf/parser.py#L508)
I followed it to keep it consistent. 
But I understand your point, and I will move this to the handler when I implement it. I will remove it from parser.

genai-perf/genai_perf/parser.py
```diff
@@ -1025,6 +1037,16 @@ def _add_output_args(parser):
     )
 
 
+def _add_process_export_files_args(parser):
+    process_export_files_group = parser.add_argument_group("Process Export Files")
+    process_export_files_group.add_argument(
+        "input_path",
```
nv-hwoo, line 1043: Should this be `--input-path`?

debermudez (commented):
i echo @nv-braf 's comments

genai-perf/genai_perf/parser.py
debermudez: Do you anticipate that being a typical use case?
I dont think it should be on us to aggregate a list of directories.

lkomali: The initial assumption is for the user to provide single directory path containing the profile export files.
But we should discuss this more if that's a good UX or not. 
Right now i'm implementing stage 1 which expects single set of export files i.e. export files from one run and GAP processes it. 
In case of processing export files from multiple runs, which is the next stage, we could discuss and decide f user should move all the files to one directory or user could specify a list of paths and GAP be able to do it.
Thanks for bringing this point.

genai-perf/genai_perf/parser.py
lkomali: The intention is for user to specify the path as genai-perf process-export-files <path to directory> without the --input-path flag.
I discussed this with @debermudez offline yesterday if we should have a flag or not and we leaned towards not having the flag and directly be able to pass the path. 
Let me know your thoughts on this.

genai-perf/genai_perf/parser.py
nv-hwoo: @debermudez haha I don't know since I'm not the user/stakeholder (and haven't talked with the stakeholders for this feature) but just got curious.

@lkomali Thanks for explaining. Yeah I'm fine as long as we are addressing what's the user requires.

genai-perf/genai_perf/parser.py
nv-hwoo: Ah I see. I thought this was a typo.

nv-hwoo (approved):

PR #335, opened 2025-03-14
Improve logging UX and some misc

nv-hwoo:
1. Use `RichHandler` to improve our logging UX
2. Keep `vision` endpoint for now for backward compatibility but add deprecation warning message.

Logging before
image

Logging after
Screenshot 2025-03-14 at 1 54 15 PM

### DIFF
--- a/genai-perf/genai_perf/logging.py
+++ b/genai-perf/genai_perf/logging.py
@@ -26,12 +26,24 @@
 
 import logging
 import logging.config
+import os
+from typing import Optional
 
-DEFAULT_LOG_FORMAT = "%(asctime)s [%(levelname)s] %(name)s:%(lineno)s - %(message)s"
-DEFAULT_DATE_FORMAT = "%Y-%m-%d %H:%M"
+from rich.highlighter import NullHighlighter
+
+DEFAULT_LOG_FORMAT = "%(message)s"
+DEFAULT_DATE_FORMAT = "[%Y-%m-%d %H:%M:%S]"
 
 
-def init_logging() -> None:
+def init_logging(log_level: Optional[str] = None) -> None:
+    """Initialize logging configuration for the genai_perf package.
+
+    Args:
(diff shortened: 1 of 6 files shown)


genai-perf/genai_perf/logging.py
```diff
@@ -43,82 +73,42 @@ def init_logging() -> None:
...
-                "propagate": False,
-            },
-            "genai_perf.plots.plot_manager": {
-                "handlers": ["console"],
-                "level": "DEBUG",
+                "level": log_level,
                 "propagate": False,
             },
-            "genai_perf.export_data.json_exporter": {
+            "genai_perf": {  # All modules in genai_perf package
```
debermudez, line 95: if we define a more specific handler, will it overwrite the general one?

debermudez (commented):
This is fantastic. Great find.
1 small question

genai-perf/genai_perf/logging.py
nv-hwoo: Not sure if I fully understood the question, but if we define a different handler, then we can update the `HANDLER` variable to point to it. (So I guess the answer is yes?)

debermudez (approved):

genai-perf/genai_perf/logging.py
debermudez: that makes sense to me.

genai-perf/genai_perf/logging.py
```diff
@@ -26,13 +26,43 @@
 
 import logging
 import logging.config
+import os
+from typing import Optional
 
-DEFAULT_LOG_FORMAT = "%(asctime)s [%(levelname)s] %(name)s:%(lineno)s - %(message)s"
-DEFAULT_DATE_FORMAT = "%Y-%m-%d %H:%M"
+from rich.highlighter import NullHighlighter
+from rich.logging import RichHandler
```
matthewkotila, line 33: This class be like

gif

PR #337, opened 2025-03-19
Initiate logger in TestConsoleExporter::test_missing_data

dyastremsky:
Since the logger isn't initialized in the below sub-test, the errors are not logged correctly for the unit test to pick them up unless the test is run after other test, breaking independence of tests. Add logger initialization so that the test passes.

Before (failure):
image

After (success):
image


nv-hwoo (approved):
👍

PR #338, opened 2025-03-19
Update CLI defaults

dyastremsky:
This pull request does the following:
- Set request-count by default. This is the most configuration-agnostic way to run GenAI-Perf and is the default approach for most users. Measurement interval is still available as a mutually exclusive option.
- The tokenizer arg is now required. It should reasonably be used in every use case except when token counts are unused (in which case, users can always set to an arbitrary model like gpt2). Omitting it has resulted in users getting unexpected token counts due to GenAI-Perf using the default tokenizer, so this removes a silent failure. There may be a better solution in the long-term, like making the need for a tokenizer dependent on the specified converter, but this will immediately eliminate the case where metrics are silently wrong due to users not specifying a tokenizer.
- Add support for --session-delay-ratio, providing users a knob for easily benchmarking different multi-turn situations (speed up or slow down delays by a given amount) when providing their own payloads
- Set the random seed to a random value if it is not set.

Example of the new session-delay ratio argument:
image


genai-perf/README.md
```diff
@@ -490,6 +490,10 @@ Currently only 1 (mono) and 2 (stereo) are supported.
 
 ##### `-b <int>`
 ##### `--batch-size <int>`
+##### `--batch-size-text <int>
```
dyastremsky, line 493: This arg and description were missing in README, but not README_template.

genai-perf/README.md
```diff
@@ -634,9 +637,11 @@ The concurrency value to benchmark. (default: `None`)
...
 ##### `-p <int>`
 
-The time interval used for each measurement in milliseconds. Perf Analyzer
-will sample a time interval specified and take measurement over the requests
-completed within that time interval. (default: `10000`)
+The time interval used for each measurement in milliseconds.
+Perf Analyzer will sample a time interval specified and take
+measurement over the requests completed within that time interval.
+When using the default stability percentage, GenAI-Perf will benchmark
+for 3*(measurement_interval) milliseconds.
```
dyastremsky, line 644: Long-term, we may want to deprecate this arg in favor of something clearer, where GenAI-Perf natively converts to measurement interval e.g. --benchmarking-duration 1000 gets converted to --measurement-interval 333 under the hood. To avoid any kind of confusion with the PA command printed or prior behavior, I'm leaving this as is for now.

genai-perf/README.md
```diff
@@ -702,13 +713,13 @@ The mean number of turns per session.
 The standard deviation of the number of turns per session.
 (default: `0`)
 
-### Other Options
+### Tokenizer Options
```
dyastremsky, line 716: The parser now has a tokenizer arg section, so I modified the README to match that hierarchy.

genai-perf/docs/goodput.md
```diff
@@ -66,6 +66,7 @@ genai-perf profile \
     --service-kind openai \
     --endpoint-type chat \
     --streaming \
+    --request-count 10 \
```
dyastremsky, line 69: Added request-count here since the other example uses measurement interval. This wasn't strictly necessary, but I liked the symmetry of showing both options.

genai-perf/docs/multi_turn.md
```diff
@@ -103,18 +103,29 @@ with the dataset.
 ### Example Input File
 
 ```bash
-echo '{"session_id": "f81d4fae-7dec-11d0-a765-00a0c91e6bf6", "delay": 1, "input_length": 50, "output_length": 10}
-{"session_id": "f81d4fae-7dec-11d0-a765-00a0c91e6bf6", "delay": 2, "input_length": 50, "output_length": 10}
+echo '{"session_id": "f81d4fae-7dec-11d0-a765-00a0c91e6bf6", "delay": 1000, "input_length": 50, "output_length": 10}
```
dyastremsky, line 106: Since these are in ms, I increased this to be 1000ms. I think it's clearer, as I initially thought this was 1s.

genai-perf/README.md
```diff
@@ -568,14 +572,13 @@ when `--output-tokens-mean` is provided. (default: `0`)
 
 ##### `--random-seed <int>`
 
-The seed used to generate random values. (default: `0`)
```
nv-hwoo, line 571: Can we mention the default behavior here?

genai-perf/genai_perf/parser.py
```diff
@@ -254,6 +254,8 @@ def _check_conditional_args(
                 f"The --generate-plots option is not currently supported with the {args.endpoint_type} endpoint type."
             )
 
+    if args.random_seed is None:
+        args.random_seed = random.randint(0, sys.maxsize)
```
nv-hwoo, line 258: I think we can leave the seed as `None`. The python `random` will by default use system time for randomness: https://docs.python.org/3/library/random.html#random.seed

genai-perf/README.md
dyastremsky: Good call-out. Done.

genai-perf/genai_perf/parser.py
```diff
@@ -1132,8 +1149,7 @@ def _add_tokenizer_args(parser):
     tokenizer_group.add_argument(
         "--tokenizer",
         type=str,
-        default=DEFAULT_TOKENIZER,
-        required=False,
+        required=True,
```
nv-hwoo, line 1152: I think for many users, model name and tokenizer name are the same since both are coming from the same hf repo. Instead of making it required, would it make more sense for gap to first try fetching the tokenizer from hf repo using the model name, and raise an error if it cannot find one? This will remove the need to specify both model and tokenizer for casual users who are not using local tokenizer.

dyastremsky: Great point. I thought about this and don't know why I ended up not adding it.

Let me do that now.

nv-hwoo (dismissed):
LGTM!

dyastremsky:
> LGTM!

Thanks, Hyunjae! Would you be able to do one final pass on the latest commit? Your great suggestion to default the tokenizer name to the model name made it possible to remove extra flags all over the unit testing and docs.

nv-hwoo:
Also tagging @nv-braf just in case the cli change affects the config works.

genai-perf/genai_perf/parser.py
```diff
@@ -509,6 +509,10 @@ def _infer_prompt_source(args: argparse.Namespace) -> argparse.Namespace:
 
 def _infer_tokenizer(args: argparse.Namespace) -> argparse.Namespace:
     if not hasattr(args, "tokenizer") or args.tokenizer is None:
+        if not hasattr(args, "model") or args.model is None:
```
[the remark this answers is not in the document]
dyastremsky: The reason for this was the `compare` subcommand, which I suppose is now being deprecated. I can make the checking less rigorous here.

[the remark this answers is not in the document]
nv-hwoo: Oh you're right. Yes thanks for pointing that out.

genai-perf/genai_perf/parser.py
```diff
@@ -509,6 +509,10 @@ def _infer_prompt_source(args: argparse.Namespace) -> argparse.Namespace:
...
+            raise ValueError(
+                "The --tokenizer option is required when not specifying a model."
```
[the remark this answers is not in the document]
dyastremsky: I'd prefer if we leave the logic as is. As you pointed out, the model is required. For most use cases where there's a language component, it's important to have the right tokenizer or there's a silent error in the token count. If we do the above, we're still risking users not seeing the warning and getting the wrong token-related metrics, which is a big deal.

If users are using a converter that doesn't need a tokenizer, hopefully the tokenizer for their model still exists on HuggingFace. If not, they can use any tokenizer name (e.g. `gpt2`). Long-term, there might be a better design whereby we can make the logic more specific with which models use text/images/audio/other and do better arg processing based on that. For now, my bigger concern is safeguarding against folks getting inaccurate metrics.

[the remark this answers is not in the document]
nv-hwoo: Yeah, true. This is probably a minor UX improvement, so I think we're good here.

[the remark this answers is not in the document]
dyastremsky: I'd also love for us long-term to get rid of the default_tokenizer across the codebase. For LLM models, we want to require the tokenizer everywhere (defaults risk hidden errors). For non-LLM models, we want to avoid making any tokenizer calls or even importing the tokenizer.

genai-perf/genai_perf/parser.py
```diff
@@ -269,6 +268,15 @@ def _check_load_manager_args(args: argparse.Namespace) -> argparse.Namespace:
     return args
 
 
+def _check_measurement_args(args: argparse.Namespace) -> argparse.Namespace:
```
nv-braf, line 271: This logic needs to be replicated in the config as well otherwise what we do with the CLI isn't consistent with the config file.

genai-perf/genai_perf/parser.py
```diff
@@ -499,6 +507,16 @@ def _infer_prompt_source(args: argparse.Namespace) -> argparse.Namespace:
     return args
 
 
+def _infer_tokenizer(args: argparse.Namespace) -> argparse.Namespace:
```
nv-braf, line 510: This logic needs to be replicated in the config as well otherwise what we do with the CLI isn't consistent with the config file.

nv-hwoo (dismissed):
LGTM on my side. I will leave the rest of the review to @nv-braf

genai-perf/genai_perf/parser.py
dyastremsky: Done!

genai-perf/genai_perf/parser.py
dyastremsky: Thanks, added a new commit with the config changes and testing. Can you review it when you have a chance?

dyastremsky:
I just pushed a change to mock all HuggingFace calls in test_tensortllm_engine_converter. It looks like they were unintentionally added at some point.

When we create the new base unit test class, we should disable all network calls as part of it so that we can ensure our unit tests don't make any calls to HuggingFace or any real endpoints when unit testing. We do this manually today, so it'd be better to have an automated check for that.

dyastremsky:
Thanks for refactoring the changes to be more elegant and streamlined, Brian! Other than the two comments by CodeQL, this looks good on my end.

nv-braf (approved):

genai-perf/genai_perf/config/input/config_perf_analyzer.py
```diff
@@ -77,37 +76,63 @@ def parse(self, perf_analyzer: Dict[str, Any]) -> None:
...
+            elif key == "measurement":
+                self.measurement.parse(value)
             else:
                 raise ValueError(
                     f"User Config: {key} is not a valid perf_analyzer parameter"
                 )
 
+    ###########################################################################
+    # Infer Methods
+    ###########################################################################
```
nv-hwoo, line 90: Is this header necessary? Seems like we only have one method under it.

genai-perf/genai_perf/config/input/config_perf_analyzer.py
```diff
@@ -77,37 +76,63 @@ def parse(self, perf_analyzer: Dict[str, Any]) -> None:
...
-                self.warmup = value
             else:
                 raise ValueError(
-                    f"User Config: {key} is not a valid request_count parameter"
+                    f"User Config: {key} is not a valid perf_analyzer_measurement parameter"
+                )
+
+    ###########################################################################
+    # Infer Methods
+    ###########################################################################
```
nv-hwoo, line 128: Same as above

genai-perf/genai_perf/config/input/config_perf_analyzer.py
```diff
@@ -77,37 +76,63 @@ def parse(self, perf_analyzer: Dict[str, Any]) -> None:
...
-            verbose_template_comment="The number of warmup requests to send before benchmarking.",
+            template_comment="If left unset, the default value will be inferred based on the mode.",
+            verbose_template_comment="The number to use for measurement. By default, this will be inferred based on the mode.",
         )
 
-    def parse(self, request_count: Dict[str, Any]) -> None:
-        for key, value in request_count.items():
-            if key == "num":
+    def parse(self, perf_analyzer_measurement: Dict[str, Any]) -> None:
+        for key, value in perf_analyzer_measurement.items():
```
nv-hwoo, line 115: can we give more meaningful names to the variables? `key` and `value` seems very generic and didn't really help me understanding the code.

genai-perf/genai_perf/config/input/config_command.py
```diff
@@ -112,6 +121,8 @@ def _parse_model_names(self, model_names: Any) -> None:
     def _infer_settings(self) -> None:
         self.endpoint.infer_settings(model_name=self.model_names[0])
         self.input.infer_settings()
+        self.perf_analyzer.infer_settings()
+        self.tokenizer.infer_settings(model_name=self.model_names[0])
```
nv-hwoo, line 125: What is this magic number?

nv-hwoo (commented):
Few comments, nothing critical tho

genai-perf/genai_perf/config/input/config_perf_analyzer.py
nv-braf: It's not necessary, but it matches (my) the style of all the other config_*.py files. We only have one method right now, but it's very possible we have more in the future.

genai-perf/genai_perf/config/input/config_command.py
nv-braf: Not sure how to un-magic this! The user provides a list of model names, but we only ever care about the first model name they provide except in very specific instances.

nv-braf: Btw, this magic [0] occurs throughout the GAP code and it's probably beyond the scope of this story to fix it.

nv-hwoo: Yeah I think this is fine since we don't have a good way of handling multiple model names (or multi-lora) atm.

nv-hwoo (approved):

PR #339, opened 2025-03-19
Remove inputs config class

nv-braf:


genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
```diff
@@ -24,35 +24,39 @@
...
-                f"The --batch-size-text flag is not supported for {config.output_format.to_lowercase()}."
+                f"The --batch-size-text flag is not supported for {config.endpoint.output_format.to_lowercase()}."
             )
 
     def convert(
-        self, generic_dataset: GenericDataset, config: InputsConfig
+        self,
+        generic_dataset: GenericDataset,
+        config: ConfigCommand,
+        tokenizer: Optional[Tokenizer] = None,
```
nv-hwoo, line 53: I see we are passing tokenizer to `convert` method but where are we passing the tokenizer to this method?

nv-hwoo: Also it seems like the only converter that needs tokenizer is this trtllmengine converter. I would say we find a different way to pass tokenizer to this converter than changing all the converters that don't need it.

nv-braf: It comes from the BaseConverter class. 
I'm not sure how to go about changing this though since I believe we need to be consistent between all the convert methods that share the base class.
Unless I'm missing something trivial, I think a change like this is beyond the scope of this PR.

genai-perf/tests/test_config_command.py
```diff
@@ -73,7 +73,7 @@ def test_multi_model_name_string(self):
         user_config = yaml.safe_load(yaml_str)
         config = ConfigCommand(user_config)
 
-        self.assertEqual(config.model_names, ["gpt2_multi"])
+        self.assertEqual(config.model_names, ["gpt2", "bert"])
```
dyastremsky, line 76: Is there a way to test what was previously being tested here (the model name string that's going to be used/printed)?

dyastremsky:
Great work! Added a few comments.

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
dyastremsky: I [commented elsewhere](https://github.com/triton-inference-server/perf_analyzer/pull/339#discussion_r2006478911), but wanted to comment here to avoid having two different threads.

I think our rule is that we don't do anything in the base converter that's specific to one converter. Is there a reason we need to pass the tokenizer here? Is it possible for the ConfigCommand class to hold the tokenizer? That's equivalent to what this class was doing before. It seems like ConfigCommand and InputsConfig were designed to do two different things, so some of this might just require thoughtful design to decide how to integrate the two. I think that since this PR is introducing this change, we probably should figure out a way to resolve this issue before merging.

If we want to move this out of scope and do not want to add the tokenizer to ConfigCommand, then one option is having the converters continue using InputsConfig, though it'd be a bit clunkier to call `config.config.<field>` in the meantime.

genai-perf/genai_perf/inputs/converters/base_converter.py
```diff
@@ -52,25 +53,31 @@ def check_config(self, config: InputsConfig) -> None:
         pass
 
     def convert(
-        self, generic_dataset: GenericDataset, config: InputsConfig
+        self,
+        generic_dataset: GenericDataset,
+        config: ConfigCommand,
+        tokenizer: Optional[Tokenizer] = None,
```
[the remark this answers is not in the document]
nv-braf: The Config class can not contain anything other than the values set by the config (so only ConfigField elements). The cleanup here is for InputsConfig to use the Config class (rather than replicating it's own version of a config).
I've also removed the usage of the InputsConfig wherever possible (replacing it with Config).

genai-perf/genai_perf/inputs/inputs_config.py
```diff
@@ -28,153 +28,14 @@
...
-    """
-    A class to hold all of the arguments used for creating the inputs
-    """
-
-    ####################
-    # General Parameters
-    ####################
-
-    # The tokenizer to use when generating synthetic prompts
+    config: ConfigCommand
```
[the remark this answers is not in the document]
nv-braf: The Config class can not contain anything other than the values set by the config (so only ConfigField elements). The cleanup here is for InputsConfig to use the Config class (rather than replicating it's own version of a config).
I've also removed the usage of the InputsConfig wherever possible (replacing it with Config).

I've greatly simplified this class, but we cannot completely remove it.

genai-perf/genai_perf/inputs/retrievers/base_input_retriever.py
```diff
@@ -29,8 +29,10 @@
 
 
 class BaseInputRetriever:
-    def __init__(self, config: InputsConfig):
-        self.config = config
+    def __init__(self, inputs_config: InputsConfig):
+        self.config = inputs_config.config
+        self.tokenizer = inputs_config.tokenizer
+        self.output_directory = inputs_config.output_directory
```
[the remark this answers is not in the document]
nv-braf: Because the output directory changes each time PA is called in the analyze subcommand (based on the parameters being swept).

genai-perf/genai_perf/inputs/converters/base_converter.py
```diff
@@ -52,25 +53,31 @@ def check_config(self, config: InputsConfig) -> None:
...
         """
         raise NotImplementedError("This method should be implemented by subclasses.")
 
-    def _select_model_name(self, config: InputsConfig, index: int) -> str:
-        if config.model_selection_strategy == ModelSelectionStrategy.ROUND_ROBIN:
-            return config.model_name[index % len(config.model_name)]
-        elif config.model_selection_strategy == ModelSelectionStrategy.RANDOM:
-            return random.choice(config.model_name)
+    def _select_model_name(self, config: ConfigCommand, index: int) -> str:
+        if (
```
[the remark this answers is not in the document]
nv-braf: This is something the formatter adds when the compare statement is too long to fit on a single line (I have no control over the ()'s).

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
nv-braf: It is not possible for the ConfigCommand class to hold the tokenizer. I can change the converters back to using the InputsConfig if this is going to be problematic.

genai-perf/tests/test_config_command.py
nv-braf: This functionality was moved into perf_analyzer_config.py and is tested there.

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
nv-braf: I will say that either way things needed to change (since InputsConfig was changing). Either we added a layer of hierarchy to the config references in all the converters (config.config....) or we add in a new optional tokenizer argument which is ignored by all the instances that don't need them.
I thought the latter was a cleaner approach.

nv-hwoo: I think one way we can do is to pass tokenizer to the converter factory and handle the logic of passing the tokenizer to a specific converter inside the factory. This way, we don't global changes to the `convert` method for all converters.

nv-hwoo: > It is not possible for the ConfigCommand class to hold the tokenizer. I can change the converters back to using the InputsConfig if this is going to be problematic.

It makes sense to have config commands in the converter but we just need to solve this problem thats it.

genai-perf/genai_perf/inputs/inputs_config.py
```diff
@@ -28,153 +28,14 @@
 from pathlib import Path
 from typing import Any, Dict, List, Optional
 
+from genai_perf.config.input.config_command import ConfigCommand
 from genai_perf.inputs import input_constants as ic
 from genai_perf.tokenizer import Tokenizer
 
 
+@dataclass
```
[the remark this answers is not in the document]
nv-braf: Removed.

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
nv-braf: I don't understand how I can solve this problem.

dyastremsky:
Please update the customizable frontends [documentation](https://github.com/triton-inference-server/perf_analyzer/blob/main/genai-perf/docs/customizable_frontends.md) to reflect these changes.

genai-perf/tests/test_config_command.py
dyastremsky: I don't see a test there for multi-model. Can you point me to it?

genai-perf/genai_perf/inputs/converters/base_converter.py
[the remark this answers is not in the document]
dyastremsky: I see. So the user specifies the tokenizer as a string, but there's no tokenizer held by that class.

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
dyastremsky: Unfortunately, I think this is pointing to the fact that the two classes are doing different things. We need a really brilliant design to merge them to a class with a shared purpose or keep them separate. Keeping a command_config inside of inputs_config might be the best of both worlds so that we're not duplicating fields.

@nicolasnoble, would you be able to weigh in here? Apologies for always tagging you on complex design issues in reviews. You're good at bringing new insights.

nv-braf: Had a conversation with Hyunjae and I understand how to clean this up.

genai-perf/tests/test_config_command.py
nv-braf: There's no specific test for multi-model - but, we do check that the non-multi-model case was unbroken whenever we check the value for the `--artifact-directory`. Didn't see a need for a specific test for this as my understanding is that the `_multi` is a GAP convention and that the artifact name doesn't actually matter (as long as we are consistent) - which is guaranteed since this is the only function that can create an artifact name.

dyastremsky: Right, the test was there for that consistency. Many users script the artifact ingestion, so we want to make sure the artifact naming (and anywhere else we using this naming) is consistent from release to release. It seems like a small feature unit test would be good to make sure that no changes accidentally break this.

nv-braf: I've added a unit test in PA config generate to cover this case.

nv-braf:
> Please update the customizable frontends [documentation](https://github.com/triton-inference-server/perf_analyzer/blob/main/genai-perf/docs/customizable_frontends.md) to reflect these changes.

Updated!

nv-hwoo (approved):
Looks good to me!

PR #340, opened 2025-03-20
Fixing Compare Subcommand Parsing

nv-braf:
This fixes the bug I introduced into the compare subcommand with the config file support work.

The issue is that the compare subcommand is not supported via the config and therefore I should not have changed the top-level parsing in main to use the config file when doing a compare action.

This could have been caught if we had any L0 testing using the compare subcommand.


dyastremsky (approved):
Nice fix!

PR #342, opened 2025-03-20
Fix process-export-files subcommand

lkomali:
This PR fixes the broken parts after the config file work is merged.
It does nothing new, fix broken parts and unit tests of the process-export-files subcommand work done till now.

### DIFF
--- a/genai-perf/genai_perf/parser.py
+++ b/genai-perf/genai_perf/parser.py
@@ -933,7 +933,7 @@ def _parse_analyze_args(subparsers) -> argparse.ArgumentParser:
 
 def _parse_process_export_files_args(subparsers) -> argparse.ArgumentParser:
     process_export_files = subparsers.add_parser(
-        Subcommand.PROCESS_EXPORT_FILES.to_cli_format(),
+        Subcommand.PROCESS.value,
         description="Subcommand to process export files and aggregate the results.",
     )
     _add_process_export_files_args(process_export_files)
@@ -995,7 +995,7 @@ def refine_args(
         args = _check_compare_args(parser, args)
     elif args.subcommand == Subcommand.TEMPLATE.value:
         pass
-    elif args.subcommand == Subcommand.PROCESS_EXPORT_FILES.to_cli_format():
+    elif args.subcommand == Subcommand.PROCESS.value:
         pass
     else:
         raise ValueError(f"Unknown subcommand: {args.subcommand}")
@@ -1168,9 +1168,9 @@ def parse_args():
             config = _create_template_config(args, argv)
 
             return args, config, None
-        elif args.subcommand == Subcommand.COMPARE.value:
-            # this subcommand is deprecated and is not supported in the config file
-            return args, None, argv[passthrough_index + 1 :]
+        if args.subcommand == Subcommand.PROCESS.value:
+            config = _create_process_export_files_config(args)
+            return args, config, None
         else:
             # For all other subcommands, parse the CLI fully (no config file)
             config = ConfigCommand(
@@ -1237,6 +1237,23 @@ def _create_template_config(args: argparse.Namespace, argv: List[str]) -> Config
         default=TemplateDefaults.FILENAME, value=filename, add_to_template=False
     )
 
+    return config
+
+
+def _create_process_export_files_config(args: argparse.Namespace) -> ConfigCommand:
+    config = ConfigCommand({"model_name": ""})
+
+    config.subcommand = ConfigField(
+        default="process-export-files", value=args.subcommand, required=True
+    )
+    config.input.path = ConfigField(
+        default=None, value=args.input_path[0], required=True
+    )
+    config.output.artifact_directory = args.artifact_dir
+    config.output.profile_export_file = args.profile_export_file
+    config.output.generate_plots = args.generate_plots
+
+    config.verbose = ConfigField(default=False, value=args.verbose)
     return config
 
 
--- a/genai-perf/genai_perf/parser.py
+++ b/genai-perf/genai_perf/parser.py
@@ -1003,6 +1003,15 @@ def refine_args(
     return args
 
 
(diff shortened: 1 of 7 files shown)


genai-perf/genai_perf/parser.py
```diff
@@ -1542,6 +1545,23 @@ def _create_template_config(args: argparse.Namespace, argv: List[str]) -> Config
     return config
 
 
+def _create_process_export_files_config(args: argparse.Namespace) -> ConfigCommand:
```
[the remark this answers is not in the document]
lkomali: yep, was thinking the same after I raised the PR. I will do that.

genai-perf/genai_perf/parser.py
```diff
@@ -1316,14 +1316,23 @@ def refine_args(
...
-    elif args.subcommand == Subcommand.PROCESS_EXPORT_FILES.to_cli_format():
+    elif args.subcommand == Subcommand.PROCESS.value:
         pass
     else:
         raise ValueError(f"Unknown subcommand: {args.subcommand}")
 
     return args
 
 
+def _set_output_config(config: ConfigCommand, args: argparse.Namespace) -> None:
```
lkomali, line 1327: Moved this a private method so to avoid duplication.

genai-perf/genai_perf/parser.py
```diff
@@ -1333,6 +1342,14 @@ def add_cli_options_to_config(
     )
     config.verbose = ConfigField(default=False, value=args.verbose)
 
+    # Process Export Files
+    if args.subcommand == "process-export-files":
+        config.input.path = ConfigField(
+            default=None, value=args.input_path[0], required=True
+        )
+        _set_output_config(config, args)
+        return config
```
lkomali, line 1351: Returning the config here at this point because the rest of the args shouldn't be set in this subcommand.

genai-perf/genai_perf/parser.py
```diff
@@ -1476,7 +1491,10 @@ def parse_args():
 
         if args.subcommand == Subcommand.TEMPLATE.value:
             config = _create_template_config(args, argv)
-
+            return args, config, None
+        elif args.subcommand == Subcommand.PROCESS.value:
```
lkomali, line 1495: I left this still as a separate condition although it does similar to the else condition because
1. `args.formatted_model_name` is not an option in this subcommand, and i don;t want to set it just for the sake of creating a config.
2. The logging line which says Profiling model doesn't apply because we don't profile in this subcommand.

genai-perf/genai_perf/parser.py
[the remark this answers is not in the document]
lkomali: I modified and used the existing `_add_cli_options_to_config`. I added single-line comments to explain my choices. Let me know if they can be improved or if there's a better approach.

genai-perf/genai_perf/parser.py
nv-hwoo: If nothing below this is needed, why not just create a small method for the subcommand (e.g. `add_process_cli_options_to_config` or `_create_process_config` following the template subcommand)? I think that would be much cleaner and removes the need for extra/redundant checking logics in this function.

genai-perf/genai_perf/parser.py
```diff
@@ -1427,10 +1444,8 @@ def add_cli_options_to_config(
     config.input.sessions.turns.stddev = args.session_turns_stddev
 
     # Output
-    config.output.artifact_directory = args.artifact_dir
     # config.output.checkpoint_directory - There is no equivalent setting in the CLI
```
nv-hwoo, line 1447: Unrelated but seems like this comment is not used? Can we clean it up?

nv-hwoo (commented):
Nice work! 👍 Left a few comments.

genai-perf/genai_perf/parser.py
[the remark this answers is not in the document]
nv-hwoo: I just read this after I made the comment that suggests exact opposite lol
@nv-braf If most of the options are not needed for the subcommand, I think it's much cleaner to just have a separate function for it. Less extra checks and clear separation of the logic.

genai-perf/genai_perf/parser.py
nv-braf: This comment is indicating that we have a field (the ability to set the checkpoint directory) available in the config file, that is not present in the CLI.

genai-perf/genai_perf/parser.py
nv-braf: This code is incorrect. Harshini, you need to add a new config entry (similar to what was done for config_analyze.py) for this new subcommand. You should not be instancing ConfigField outside of the config_*.py files.

genai-perf/genai_perf/parser.py
nv-braf, line 1327: This method is redundant. These assignments are already done on lines 1439-1444 (of the original file) in the add_cli_options_to_config() method.

genai-perf/genai_perf/parser.py
nv-braf: With my latest check-in you no longer need to supply a model name. You can change this to just be `ConfigCommand()`

genai-perf/genai_perf/parser.py
lkomali: I will update it.

lkomali:
I had a call with @nv-braf offline on the better way to proceed and make this subcommand work with config yesterday.
I got things to work and added some unit tests, and have to do some refactoring to make it clean.

genai-perf/genai_perf/parser.py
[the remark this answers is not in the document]
lkomali: @nv-hwoo Your suggestion is correct. I'm working on refactoring to make it cleaner.

nv-braf (approved):

PR #344, opened 2025-03-25
Consolidate CLI and Config Checking & Inferencing

nv-braf:
This removes the vast majority of CLI checking and inferencing from `parser.py` (only the goodput dictionary checking remains duplicated). This is now all done in the Config class and going forward all future inferencing and checking should be done in there.
I've updated the unit tests to reflect the new error messages that come from the config class. A few unit tests were no longer relevant (the logic is tested elsewhere and already exists) and were removed.

### DIFF
--- a/genai-perf/genai_perf/config/input/config_command.py
+++ b/genai-perf/genai_perf/config/input/config_command.py
@@ -54,7 +54,7 @@ class ConfigCommand(BaseConfig):
     def __init__(
         self,
         user_config: Optional[Dict[str, Any]] = None,
-        create_template: bool = False,
+        skip_inferencing_and_checking: bool = False,
     ):
         super().__init__()
 
@@ -72,7 +72,7 @@ class ConfigCommand(BaseConfig):
         self.output = ConfigOutput()
         self.tokenizer = ConfigTokenizer()
 
-        self._parse_yaml(user_config, create_template)
+        self._parse_yaml(user_config, skip_inferencing_and_checking)
 
     ###########################################################################
     # Top-Level Parsing Methods
@@ -88,7 +88,7 @@ class ConfigCommand(BaseConfig):
     def _parse_yaml(
         self,
         user_config: Optional[Dict[str, Any]] = None,
-        create_template: bool = False,
+        skip_inferencing_and_checking: bool = False,
     ) -> None:
         if user_config:
             for key, value in user_config.items():
@@ -111,7 +111,7 @@ class ConfigCommand(BaseConfig):
                         f"User Config: {key} is not a valid top-level parameter"
                     )
 
-        if not create_template:
+        if not skip_inferencing_and_checking:
             self.infer_and_check_options()
 
     def _parse_model_names(self, model_names: Any) -> None:
@@ -143,6 +143,7 @@ class ConfigCommand(BaseConfig):
         self._check_payload_input()
 
         self.endpoint.check_for_illegal_combinations()
+        self.input.check_for_illegal_combinations()
 
     def _check_output_tokens_and_service_kind(self) -> None:
         if self.endpoint.service_kind not in ["triton", "tensorrtllm_engine"]:
@@ -160,7 +161,7 @@ class ConfigCommand(BaseConfig):
         ]:
             if self.output.generate_plots:
                 raise ValueError(
-                    "User Config: generate_plots is not supported with the {self.endpoint.output_format} output format"
+                    f"User Config: generate_plots is not supported with the {self.endpoint.output_format} output format"
                 )
 
     def _check_payload_input(self) -> None:
--- a/genai-perf/genai_perf/config/input/config_command.py
+++ b/genai-perf/genai_perf/config/input/config_command.py
@@ -26,7 +26,11 @@ from genai_perf.config.input.config_input import ConfigInput
 from genai_perf.config.input.config_output import ConfigOutput
 from genai_perf.config.input.config_perf_analyzer import ConfigPerfAnalyzer
 from genai_perf.config.input.config_tokenizer import ConfigTokenizer
-from genai_perf.inputs.input_constants import OutputFormat
+from genai_perf.inputs.input_constants import (
+    OutputFormat,
+    PerfAnalyzerMeasurementMode,
+    PromptSource,
+)
 from genai_perf.utils import split_and_strip_whitespace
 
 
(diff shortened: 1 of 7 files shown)


genai-perf/genai_perf/config/input/config_command.py
```diff
@@ -68,15 +72,23 @@ def __init__(
...
+        """
+        self._infer_settings()
+        self._check_for_illegal_combinations()
+        self._check_profile_export_file()
+
     def _parse_yaml(
         self,
         user_config: Optional[Dict[str, Any]] = None,
-        create_template: bool = False,
+        skip_inferencing_and_checking: bool = False,
```
nv-hwoo, line 91: what is "skip inferencing and checking"?

genai-perf/tests/test_cli.py
```diff
@@ -133,27 +133,27 @@ def test_help_version_arguments_output_and_exit(
...
             (
                 ["--endpoint-type", "rankings", "--service-kind", "openai"],
-                {"endpoint": "v1/ranking"},
+                {"endpoint": None},
                 {"endpoint.custom": "v1/ranking"},
             ),
             (
                 ["--endpoint-type", "image_retrieval", "--service-kind", "openai"],
-                {"endpoint": "v1/infer"},
+                {"endpoint": None},
```
nv-hwoo, line 156: what is the difference between endpoint and endpoint.custom? is this something that user sets? why are endpoints here all None?

genai-perf/genai_perf/config/input/config_command.py
```diff
@@ -149,9 +161,33 @@ def _check_output_format_and_generate_plots(self) -> None:
         ]:
             if self.output.generate_plots:
                 raise ValueError(
-                    "User Config: generate_plots is not supported with the {self.endpoint.output_format} output format"
+                    f"User Config: generate_plots is not supported with the {self.endpoint.output_format} output format"
```
dyastremsky, line 164: Thanks for this cleanup! Are you adding unit testing for error handling? We have test_conditional_errors in test_cli.py, which should hopefully be usable for the config now that the checking is being consolidated. We should be unit testing all the errors we throw somewhere centralized.

genai-perf/genai_perf/config/input/config_field.py
```diff
@@ -48,25 +48,53 @@ def __init__(
...
+            self._check_item_bounds(self.value)
+
+    def _check_item_bounds(self, item: Any) -> None:
+        if isinstance(item, (int, float)):
             if self.bounds and "upper" in self.bounds:
-                if self.bounds["upper"] < self.value:
+                if self.bounds["upper"] < item:
                     raise ValueError(
-                        f"User Config: {self.value} exceeds upper bounds (f{self.bounds})"
+                        f"User Config: {item} exceeds upper bounds (f{self.bounds})"
```
dyastremsky, line 86:
```suggestion
                        f"User Config: {item} exceeds upper bounds ({self.bounds})"
```

genai-perf/genai_perf/parser.py
```diff
@@ -1493,8 +1140,12 @@ def parse_args():
             return args, None, argv[passthrough_index + 1 :]
         else:
             # For all other subcommands, parse the CLI fully (no config file)
-            config = ConfigCommand({"model_name": args.formatted_model_name})
+            config = ConfigCommand(
+                {"model_names": args.model}, skip_inferencing_and_checking=True
+            )
             config = add_cli_options_to_config(config, args)
+            config.infer_and_check_options()
```
dyastremsky, line 1147: This flow is a bit confusing. The code initializes the ConfigCommand with skip_inferencing_and_checking=True, then it calls it right after. Is there a way to clean up this flow to allow the cli_options to be used as an optional init parameter, so that these lines just becomes
`config=ConfigCommand(args)` or something similar?

genai-perf/tests/test_cli.py
dyastremsky: It feels like this consolidation requires a unit testing overhaul. If we're moving a lot from the CLI to the config, we need to not just update the unit tests we have (like this one is mostly useless), but test that a config option passed in via the CLI results in the config having the right value later after the checking/infer happens.

We'd also want to make sure that both the success and failure unit tests are comprehensively tested, like we had for the CLI.

dyastremsky: Also, for `endpoint.custom`... is that verbiage correct? It sounds like that's not really a custom endpoint, it's just the endpoint.

dyastremsky (commented):
Excellent refactor. Left a few comments, mostly about testing. I think we need to figure out a good way to make sure testing stays comprehensive with the new Config option and all its paths.

genai-perf/tests/test_cli.py
nv-braf: Everything in the CLI tests are still valid. It's just that the logic was moved into the Config class. When I added the Confiig class I updated test CLI to test both the argparse as well as the Config values. Any changes here (like the one here) reflect that the args value is no longer set (since the duplicated inference logic was removed), but shows that the inferencing is still being done (the config value is checked for correctness).

nv-braf: Yes, `endpoint.custom` is correct.  If you look at parser.py and config_endpoint.py you'll note that the descriptions match (and I would argue that the name in config more closely represents the option's description).

genai-perf/genai_perf/config/input/config_command.py
nv-braf: Yes, All unit testing for error handling is currently done in test_cli. This is how the missing f-string was uncovered.

genai-perf/genai_perf/parser.py
nv-braf: This is because we need to delay inferencing/checking until after we have assigned values to the config file via the `add_cli_options...` method.

The only other option is to make the `add_cli_options` method internal to the Config class, which I am reluctant to do.

genai-perf/genai_perf/config/input/config_command.py
nv-braf: It is also used when creating a blank config file that will be populated via the CLI (as DavidY has noted in his comment below).

genai-perf/genai_perf/config/input/config_field.py
```diff
@@ -48,25 +48,53 @@ def __init__(
...
+            self._check_item_bounds(self.value)
+
+    def _check_item_bounds(self, item: Any) -> None:
+        if isinstance(item, (int, float)):
             if self.bounds and "upper" in self.bounds:
-                if self.bounds["upper"] < self.value:
+                if self.bounds["upper"] < item:
                     raise ValueError(
-                        f"User Config: {self.value} exceeds upper bounds (f{self.bounds})"
+                        f"User Config: {item} exceeds upper bounds ({self.bounds})"
```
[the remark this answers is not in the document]
nv-braf: No, there is not. In general these are tested at the higher level, but you are correct there is not a specific test that covers this case. I went ahead an had copilot write unit tests to cover this entire class.

dyastremsky (approved):
LGTM! 🎉  Deferring to @nv-hwoo's approval for his remaining comments.

nv-hwoo (approved):

PR #345, opened 2025-03-25
Fix broken file input retriever class

nv-hwoo:
File input retriever was silently broken since there was no test covering `retrieve_data()`, which is the core API of the class.
Made the changes to fix the bug and added tests to cover `retrieve_data()` method.


dyastremsky (approved):
Great fix! Thanks for catching this and adding testing. Nice clean-up too!

PR #347, opened 2025-04-02
Add TPS/user metric to LLM benchmarking

nv-hwoo:
The PR adds one of the popular metrics when measuring performance of LLM servers: TPS/user. When benchmarking LLM server in streaming mode, we will now calculate and display TPS/user metric on console. Results will be also stored in both csv and json.

```
                               NVIDIA GenAI-Perf | LLM Metrics                                
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┳━━━━━━━━┓
┃                            Statistic ┃    avg ┃    min ┃    max ┃    p99 ┃    p90 ┃    p75 ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━╇━━━━━━━━┩
│             Time To First Token (ms) │  62.11 │  55.38 │  65.49 │  65.49 │  65.48 │  65.48 │
│            Time To Second Token (ms) │  22.22 │  18.82 │  28.89 │  28.69 │  26.90 │  23.91 │
│                 Request Latency (ms) │ 235.26 │ 235.25 │ 235.27 │ 235.27 │ 235.26 │ 235.26 │
│             Inter Token Latency (ms) │  19.24 │  18.86 │  19.99 │  19.96 │  19.76 │  19.43 │
│     Output Token Throughput Per User │  52.02 │  50.03 │  53.02 │  53.02 │  53.01 │  53.01 │
│                    (tokens/sec/user) │        │        │        │        │        │        │
│      Output Sequence Length (tokens) │  10.00 │  10.00 │  10.00 │  10.00 │  10.00 │  10.00 │
│       Input Sequence Length (tokens) │  30.00 │  30.00 │  30.00 │  30.00 │  30.00 │  30.00 │
│ Output Token Throughput (tokens/sec) │ 127.47 │    N/A │    N/A │    N/A │    N/A │    N/A │
│         Request Throughput (per sec) │  12.75 │    N/A │    N/A │    N/A │    N/A │    N/A │
│                Request Count (count) │   3.00 │    N/A │    N/A │    N/A │    N/A │    N/A │
└──────────────────────────────────────┴────────┴────────┴────────┴────────┴────────┴────────┘
```

Summary of the changes made in the PR:
- Rename from `output_token_throughputs_per_request` --> `output_token_throughputs_per_user`
- Update the record types for the metric
- Wrap the display for long stat names
- Fix/Update unit tests


genai-perf/genai_perf/export_data/console_exporter.py
```diff
@@ -46,6 +45,10 @@ def __init__(self, config: ExporterConfig):
         self._metrics = config.metrics
         self._config = config.config
 
+        # Set the maximum width of the 'Statistic' column.
+        # Any metric name+unit longer than this width will be wrapped.
+        self._max_width = 36
```
nv-hwoo, line 50: Set the max width to 36, the length of the string for `Output Token Throughput (tokens/sec)` stat name. This was the longest name+unit string before this new TPS/user metric and so setting it to this length will not wrap for all the existing metrics. But any new metric that is newly introduced and is longer than this will be wrapped.

nv-braf: Should this be a constant instead of magic number?

nv-braf (approved):
Overall this looks great!

genai-perf/genai_perf/export_data/console_exporter.py
nv-hwoo: I thought about it, and this won't likely be used or needed outside of this class so I just left it as an attribute. I could instead make it a class attribute, if you would like.

genai-perf/genai_perf/export_data/console_exporter.py
```diff
@@ -114,19 +121,12 @@ def _should_skip(self, metric_name: str) -> bool:
...
-        if metric_name == "output_token_throughput_per_request":
-            return True
-
-        # When non-streaming, skip ITL and TTFT
+        # Skip following streaming metrics when non-streaming mode
         streaming_metrics = [
             "inter_token_latency",
             "time_to_first_token",
             "time_to_second_token",
+            "output_token_throughput_per_user",
```
dyastremsky, line 129: It looked like this was previously always skipped. Is it now only skipped in non-streaming mode?

genai-perf/tests/test_exporters/test_console_exporter.py
```diff
@@ -454,20 +432,22 @@ def test_valid_telemetry_verbose(self, monkeypatch, capsys) -> None:
...
-            "└───────────────────────────────────┴────────┴───────┴───────┴───────┴───────┴───────┘\n"
+            "                             NVIDIA GenAI-Perf | LLM Metrics                             \n"
+            "┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━┳━━━━━━━┳━━━━━━━┳━━━━━━━┳━━━━━━━┓\n"
+            "┃                            Statistic ┃    avg ┃   min ┃   max ┃   p99 ┃   p90 ┃   p75 ┃\n"
+            "┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━╇━━━━━━━╇━━━━━━━╇━━━━━━━╇━━━━━━━┩\n"
+            "│             Time To First Token (ms) │   8.00 │  7.00 │  9.00 │  8.98 │  8.80 │  8.50 │\n"
+            "│            Time To Second Token (ms) │   2.00 │  1.00 │  3.00 │  2.98 │  2.80 │  2.50 │\n"
+            "│                 Request Latency (ms) │   5.00 │  4.00 │  6.00 │  5.98 │  5.80 │  5.50 │\n"
+            "│             Inter Token Latency (ms) │  11.00 │ 10.00 │ 12.00 │ 11.98 │ 11.80 │ 11.50 │\n"
+            "│     Output Token Throughput Per User │   5.00 │  4.00 │  6.00 │  5.98 │  5.80 │  5.50 │\n"
```
dyastremsky, line 443: Is this correct? Same with the above screenshots. These values match the request latencies exactly. And TTST == OSL.

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
```diff
@@ -115,102 +115,96 @@ def _parse_requests(self, requests: dict) -> LLMMetrics:
...
+                # TMA-1676: handle empty first/last responses
+                # if the latter response has zero token (e.g. empty string),
+                # then set it default to one for the sake of inter token latency
+                # calculation and to avoid divide by zero.
+                num_token = 1 if n2 == 0 else n2
+                chunked_inter_token_latency.append(round((t2 - t1) / num_token))
+            chunked_inter_token_latencies.append(chunked_inter_token_latency)
+
+            # (per-session) calculate llm metrics
+            if "session_id" in req_inputs:
```
dyastremsky, line 192: Are we going to have this new metric everywhere, even when there are no session IDs? In the screenshots, it seems like output_token_throughput_per_user is used everywhere, even when not benchmarking in session mode. I think this is going to confuse some users and lead to worse UX. If we need to leave that phrasing in for some reason, we'd at least want documentation to answer what "per_user" means, as that's likely to be a question folks will have.

dyastremsky (commented):
Excellent work! Left a few questions.

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
nv-hwoo: I think adding a description about the metric is a good idea for people who might not be familiar with the jargon. Basically, user == request == client. So TPS/user tells how many tokens were generated for each request. This metric is independent of session so regardless of session or not, we can measure TPS/user (as long as we're in streaming the responses).

genai-perf/tests/test_exporters/test_console_exporter.py
nv-hwoo: Yes, the values match because we are assigning the same (dummy) values to the metric (e.g. `[4, 5, 6]`). This is okay because the purpose of these tests are to check if the console is properly displaying the passed values. The calculation itself is being checked by other tests.

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
dyastremsky: Thanks for updating the README! Please don't forget to update the README template too.

Why is it not per request then? We use the term user/session in regards to session-based benchmarking. It seems like engineers running GenAI-Perf generally do treat user == client == session, so there can be many requests per user/client/session.

nv-hwoo: Ah, you're right. Thanks for reminding me! I'll update the template.

This was from the NVIDIA company-wide decision last year ([link](https://docs.google.com/document/d/1lbNlAq0Y9U5_aaYGGArSNOUzBjIe_MNlYs35qv-T9M0/edit?tab=t.0)) when we were trying to come up with a consistent terminology for LLM metrics (since back then everyone was using different jargons that mean the same thing). This was before any session-related feature was discussed so yeah hopefully this is not too confusing on the user. But I think TPS/user metric is pretty well-known and used metric in LLM space (e.g. Jensen GTC keynote has this metric, Dynamo uses this metric to compare perf).

nv-hwoo: Ah, you're right. Thanks for reminding me! I'll update the template.

This was from the NVIDIA company-wide decision last year when we were trying to come up with a consistent terminology for LLM metrics (since back then everyone was using different jargons that mean the same thing). This was before any session-related feature was discussed so yeah hopefully this is not too confusing on the user. But I think TPS/user metric is pretty well-known and used metric in LLM space (e.g. Jensen GTC keynote has this metric, Dynamo uses this metric to compare perf).

dyastremsky (approved):

PR #349, opened 2025-04-03
Fixing URL passing to PA in config generator logic

nv-braf:
The root cause of the bug was that the old code would always pass the URL to PA if one was specified by the user.  And then there were two (special) cases where it would pass a default URL if nothing was specified.

I coded the two special cases (and added unit testing for those), but missed the fact that `url` was not part of the `skip_list`, so in the common case the PA config generator wasn't passing along a user specified url.

I've added unit testing to cover this case and made a slight refactor of the logic to separate out protocol vs. url generation logic.

Note: This would have been caught earlier if we had an CI test that exercised this option.


dyastremsky (commented):
Looks good to me other than Copilot's nitpick (the docstring is missing the service_kind). Great work getting this fixed so quickly!

debermudez (approved):

PR #351, opened 2025-04-04
TPRD-1200: Update default branch post 25.03

mc-nv:
### Related changes:
---
server  https://github.com/triton-inference-server/server/pull/8130
perf_analyzer   https://github.com/triton-inference-server/perf_analyzer/pull/351
core    https://github.com/triton-inference-server/core/pull/433


genai-perf/genai_perf/__init__.py
```diff
@@ -24,4 +24,4 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-__version__ = "0.0.11dev"
+__version__ = "0.0.13dev"
```
matthewkotila, line 27: Why is it going up 2 numbers instead of 1?

mc-nv: I rely on sequence I have in release. 
https://pypi.org/project/genai-perf/#history

Suppose upcoming version will be `0.0.13` after April `0.0.12` release.

matthewkotila: It should be 0.0.12 for March.

The reason 25.03 is releasing in April is because of CI delays.

matthewkotila (approved):

PR #352, opened 2025-04-04
Support benchmarking for the HuggingFace TGI generate endpoint

dyastremsky:
Add support for HuggingFace TGI generate endpoint: https://huggingface.github.io/text-generation-inference/#/Text%20Generation%20Inference/generate

Pending:
- Documentation (a tutorial)

Removal of the service-kind argument will be done in an upcoming PR. Right now, the OpenAI client in Perf Analyzer is used for Hugging Face as well, so that's why the service-kind is OpenAI (plus all the same checks happen). To avoid user confusion, we'll remove service-kind soon and make it detected by GenAI-Perf.

image

### DIFF
--- a/genai-perf/docs/huggingface_tgi.md
+++ b/genai-perf/docs/huggingface_tgi.md
@@ -39,7 +39,7 @@ To launch a Hugging Face TGI server, use the official `ghcr.io` image:
 ```bash
 docker run --gpus all --rm -it \
   -p 8080:80 \
-  -e MODEL_ID=gpt2 \
+  -e MODEL_ID=<model> \
   ghcr.io/huggingface/text-generation-inference
 ```
 
@@ -51,13 +51,13 @@ Run with built-in synthetic prompts:
 
 ```bash
 genai-perf profile \
-  -m gpt2 \
+  -m <model> \
   --service-kind openai \
   --endpoint-type huggingface_generate \
-  --port localhost:8080 \
+  --url localhost:8080 \
   --batch-size-image 1 \
-  --image-width-mean 1 \
-  --image-height-mean 1 \
+  --image-width-mean 100 \
+  --image-height-mean 100 \
   --synthetic-input-tokens-mean 10
 ```
 
@@ -83,7 +83,7 @@ For instance, an example of input file would look something as following:
 
 ```bash
 genai-perf profile \
-  -m gpt2 \
+  -m <model> \
   --service-kind openai \
   --endpoint-type huggingface_generate \
   --url localhost:8080 \
--- a/genai-perf/docs/huggingface_tgi.md
+++ b/genai-perf/docs/huggingface_tgi.md
@@ -39,7 +39,7 @@ To launch a Hugging Face TGI server, use the official `ghcr.io` image:
 ```bash
 docker run --gpus all --rm -it \
   -p 8080:80 \
-  -e MODEL_ID=<model> \
+  -e MODEL_ID=llava-hf/llava-v1.6-mistral-7b-hf \
   ghcr.io/huggingface/text-generation-inference
 ```
 
@@ -51,7 +51,7 @@ Run with built-in synthetic prompts:
 
 ```bash
 genai-perf profile \
-  -m <model> \
+  -m llava-hf/llava-v1.6-mistral-7b-hf \
   --service-kind openai \
   --endpoint-type huggingface_generate \
   --url localhost:8080 \
@@ -83,7 +83,7 @@ For instance, an example of input file would look something as following:
 
 ```bash
 genai-perf profile \
-  -m <model> \
+  -m llava-hf/llava-v1.6-mistral-7b-hf \
   --service-kind openai \
   --endpoint-type huggingface_generate \
   --url localhost:8080 \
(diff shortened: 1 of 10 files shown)


genai-perf/tests/test_data_parser/test_llm_profile_data_parser.py
```diff
@@ -1355,3 +1355,358 @@ def test_no_session_metrics(self, mock_json) -> None:
...
+        self, mock_json, response_format, response, expected_text
+    ) -> None:
+        """Test the text extraction methods for different response formats."""
+        if response_format == ResponseFormat.HUGGINGFACE_GENERATE:
+            mock_json.return_value = {
+                "service_kind": "openai",
+                "endpoint": "huggingface/generate",
+                "experiments": [],
+            }
+        else:
```
nv-hwoo, line 1463: qq: how does mock_json work with triton case when it's set to openai format? Don't they have different response format?

nv-hwoo (commented):
Looks great overall! 🚀 Thanks for adding thorough tests. I left a few suggestions and clarifying comments

genai-perf/docs/huggingface_tgi.md
```diff
@@ -0,0 +1,109 @@
...
+
+```bash
+genai-perf profile \
+  -m gpt2 \
+  --service-kind openai \
+  --endpoint-type huggingface_generate \
+  --port localhost:8080 \
+  --batch-size-image 1 \
+  --image-width-mean 1 \
+  --image-height-mean 1 \
```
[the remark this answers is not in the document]
dyastremsky: Good callout. I wanted to show that this can be used in a multi-modal way. In testing, it seemed to perform fine for [Paligemma](https://huggingface.co/google/paligemma-3b-pt-224).

I know this looks quirky. It's because I wanted to use gpt2 for a small model that is also in other tutorials... it has a very small context window and b64 encoding is quite lossy, so the only image that would work is a 1x1 image. If better, I could hunt down a model with a larger context window that's meant to be multi-modal, or we could get rid of the images.

genai-perf/tests/test_data_parser/test_llm_profile_data_parser.py
dyastremsky: Good catch. You're right. Let me work on this.

genai-perf/docs/huggingface_tgi.md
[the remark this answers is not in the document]
dyastremsky: I updated the model to be a generic model, so a user could use a multi-modal model if they wish.

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
```diff
@@ -412,54 +445,79 @@ def _get_response_output_tokens(self, output_texts: List[str]) -> List[List[int]
...
-        elif data["object"] == "chat.completion.chunk":  # streaming
-            text_output = completions["delta"].get("content", "")
+            return completions.get("text", "")
         else:
             obj_type = data["object"]
             raise ValueError(f"Unknown OpenAI response object type '{obj_type}'.")
-        return text_output
+
+    def _extract_openai_multimodal_text_output(self, response: str) -> str:
+        return self._extract_openai_chat_text_output(response)
```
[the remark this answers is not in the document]
dyastremsky: Done!

genai-perf/docs/huggingface_tgi.md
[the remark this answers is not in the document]
nv-hwoo: Oh why not then specify the Paligemma as our example? Since our docs are all (or mostly) written to be copy-and-paste-able commands, I think it would be good if user has some concrete model to try out for. I also like the idea of multi-modal example too.

genai-perf/tests/test_data_parser/test_llm_profile_data_parser.py
```diff
@@ -1355,3 +1355,432 @@ def test_no_session_metrics(self, mock_json) -> None:
...
+                None,  # Will raise ValueError
+            ),
+        ],
+    )
+    @patch("genai_perf.profile_data_parser.profile_data_parser.load_json")
+    def test_extract_text_output(
+        self, mock_json, service_kind, response_format, response, expected_text
+    ) -> None:
+        """Test the text extraction methods for different response formats."""
+        if service_kind == "triton":
```
nv-hwoo, line 1471: Actually on a second thought, can't we simplify the test case by getting rid of this entirely? Since our goal for this test is to verify that this single `_extract_text_output` method is working properly, we can just instantiate `LLMProfileDataParser` with any pre-defined json variable (e.g. `openai_profile_data`), manually change response format to our target format (which the test already does), and call `_extract_text_output` with target response string.

This makes sense because testing `_extract_text_output` doesn't (and shouldn't) require the knowledge of profile export json format and therefore the test doesn't need to redefine the return values of mock_json for all the different service_kinds.

genai-perf/tests/test_data_parser/test_llm_profile_data_parser.py
dyastremsky: Done!

genai-perf/tests/test_data_parser/test_llm_profile_data_parser.py
dyastremsky: Good thinking, done! The test is more straightforward now.

genai-perf/docs/huggingface_tgi.md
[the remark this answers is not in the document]
dyastremsky: Added a model! I used the same model as the multi-modal tutorial. :)

genai-perf/tests/test_data_parser/test_llm_profile_data_parser.py
```diff
@@ -1355,3 +1355,272 @@ def test_no_session_metrics(self, mock_json) -> None:
...
+            (ResponseFormat.OPENAI_CHAT_COMPLETIONS, "data: [DONE]", ""),
+            (ResponseFormat.OPENAI_CHAT_COMPLETIONS, ": ping", ""),
+            # Unknown format (should raise)
+            ("UNKNOWN_FORMAT", "any response", None),
+        ],
+    )
+    def test_extract_text_output(self, response_format, response_text, expected):
+        config = ConfigCommand({"model_name": "test_model"})
+        config.tokenizer.name = DEFAULT_TOKENIZER
+        parser = LLMProfileDataParser.__new__(LLMProfileDataParser)  # Bypass init
```
nv-hwoo, line 1410: Cool! Didn't know about this in Python :)

nv-hwoo (approved):

PR #354, opened 2025-04-09
Implement process-export-files handler

lkomali:
This PR implements process-export-files subcommand handler which enables user to run the subcommand end-end.

**Description**

The subcommand expects `profile_export.json` and `profile_export_genai_perf.json` files as input which can be provided to GAP by providing path to the directory containing the files
E.g. `genai-perf process-export-files <path-to-input-dir>`

**Example output:**

image
**Note:** I'm using profile_export files stored in artifacts directory from previous profile. The input directory doesn't need to artifacts directory.

**Running using config file**
image

**What does the process-export-files handler do?**
1. Updates `config` object by parsing provided `profile_export_files_genai_perf.json`
2. Creates PerfAnalyzerConfig object
3. Parses `profile_export_genai_perf.json` file to extract `telemetry_stats` if they exist
4. Copies the `profiles_export.json` to `artifact_directory` as required by GAP `ProfileDataParser`.
5. Call `OutputReporter` to report the output.

**GAP output**

As of now, the subcommand can process only profile_export files from one run, so the console and csv outputs should be the same as when  profiling.
For JSON output, the subcommand field and the process field are the only fields with different values.

### DIFF
--- a/genai-perf/genai_perf/config/generate/perf_analyzer_config.py
+++ b/genai-perf/genai_perf/config/generate/perf_analyzer_config.py
@@ -198,8 +198,7 @@ class PerfAnalyzerConfig:
         self, config: ConfigCommand
     ) -> Optional[List[str]]:
         if (
-            hasattr(config.input, "prompt_source")
-            and config.input.prompt_source == PromptSource.PAYLOAD
+            config.input.prompt_source == PromptSource.PAYLOAD
             and not "session_concurrency" in config.perf_analyzer.stimulus
         ):
             stimulus = None
@@ -250,10 +249,7 @@ class PerfAnalyzerConfig:
     def _add_perf_analyzer_args(self, config: ConfigCommand) -> List[str]:
         perf_analyzer_args = []
 
-        if (
-            hasattr(config.input, "prompt_source")
-            and config.input.prompt_source != PromptSource.PAYLOAD
-        ):
+        if config.input.prompt_source != PromptSource.PAYLOAD:
             perf_analyzer_args += [
                 f"--stability-percentage",
                 f"{config.perf_analyzer.stability_percentage}",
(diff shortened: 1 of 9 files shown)


genai-perf/genai_perf/parser.py
```diff
@@ -1233,7 +1233,7 @@ def parse_args():
             # this subcommand is deprecated and is not supported in the config file
             return args, None, argv[passthrough_index + 1 :]
         elif args.subcommand == ic.Subcommand.PROCESS.value:
-            config = ConfigCommand()
+            config = ConfigCommand(skip_inferencing_and_checking=True)
             config = add_cli_options_to_config(config, args)
```
nv-braf, line 1237: After the options are added to the config you then need to `config.infer_and_check_options()`.

genai-perf/genai_perf/config/generate/perf_analyzer_config.py
```diff
@@ -338,7 +342,8 @@ def _add_inference_load_args(self, config: ConfigCommand) -> List[str]:
     def _add_prompt_source_args(self, config: ConfigCommand) -> List[str]:
         prompt_source_args = []
         if (
-            config.input.prompt_source == PromptSource.PAYLOAD
+            hasattr(config.input, "prompt_source")
```
nv-braf, line 345: This isn't necessary. The only reason this is missing is because it wasn't created during the inference step. See my comment about the failure to infer options in the config file (in parser.py).

This applies to all three changes made to this file.

genai-perf/genai_perf/config/input/config_endpoint.py
```diff
@@ -229,8 +229,10 @@ def infer_output_format(self, model_name: str) -> None:
         if self.service_kind == "triton" and self.type in ["kserve", "template"]:
             self.output_format = self.backend
         elif self.service_kind == "tensorrtllm_engine":
+            print("why")
```
nv-braf, line 232: remove print()'s

debermudez (approved):
I think with the new config subcommand, this will change somewhat after a rebase. If this has testing in the CI, I think this is good to go.

PR #355, opened 2025-04-09
Infer service kind instead of using an explicit argument

dyastremsky:
Service kind can be inferred from the backend for Triton or the endpoint type for other cases. The tool no longer needs the user to provide it. Moreover, it can be confusing when there are non-OpenAI frontends supported, such as HuggingFace TEI (rankings) and TGI (huggingface_generate).

This pull request gets rid of the service-kind argument.


genai-perf/genai_perf/config/input/config_endpoint.py
```diff
@@ -180,48 +183,44 @@ def _check_for_valid_url(self, url: str) -> None:
...
+        if self.type:
+            endpoint_config = endpoint_type_map[self.type]
+            self.service_kind = endpoint_config.service_kind
+        elif self.get_field("grpc_method").is_set_by_user:
+            self.service_kind = "dynamic_grpc"
+        else:
+            # Default case
+            self.service_kind = "triton"
+
+        self.infer_type(model_name)
```
rmccorm4, line 212: With the removal of user setting `--service-kind`, would it be correct to say that the default `--endpoint-type` is `kserve` instead of `None`? I know it's not literally `default="kserve"` in the argparse code, but I think by the time it gets here and sets default service_kind = "triton" and then `infer_type -> self.type = "kserve"`, it works out that way, right?

genai-perf/README.md
```diff
@@ -365,31 +363,25 @@ a request in order. Random means that assignment is uniformly random
...
+When benchmarking Triton, this is the backend of the model. For the
 TRT-LLM backend, you currently must set `exclude_input_in_output` to true in the
 model config to not echo the input tokens in the output. (default: tensorrtllm)
 
 ##### `--endpoint <str>`
 
 Set a custom endpoint that differs from the OpenAI defaults. (default: `None`)
 
-##### `--endpoint-type {chat,completions,embeddings,rankings}`
+##### `--endpoint-type <str>`
```
nv-braf, line 374: Is it true that the endpoint type can be anything now? I thought there was still a restricted list of choices (based on the mapping)?

genai-perf/README.md
```diff
@@ -365,31 +363,25 @@ a request in order. Random means that assignment is uniformly random
...
 
-The endpoint-type to send requests to on the server. This is only used with the
-`openai` service-kind. (default: `None`)
+The endpoint-type to send requests to on the server. (default: `None`)
```
rmccorm4, line 376: Corresponding comment on `default: None`: https://github.com/triton-inference-server/perf_analyzer/pull/355/files#r2037846874

genai-perf/genai_perf/config/endpoint_config.py
```diff
@@ -44,7 +44,7 @@ class EndpointConfig:
     "completions": EndpointConfig(
         "v1/completions", "openai", OutputFormat.OPENAI_COMPLETIONS
     ),
-    "dynamic_grpc": EndpointConfig(None, "dynamic_grpc", OutputFormat.DYANMIC_GRPC),
+    "dynamic_grpc": EndpointConfig(None, "dynamic_grpc", OutputFormat.DYNAMIC_GRPC),
```
nv-braf, line 47: Just curious how this typo was discovered? Seems like we have a testing gap...

genai-perf/genai_perf/config/input/config_endpoint.py
```diff
@@ -39,8 +42,8 @@ def __init__(self) -> None:
         )
         self.backend: Any = ConfigField(
             default=EndPointDefaults.BACKEND,
-            choices=OutputFormat,
-            verbose_template_comment="When using the \"triton\" service-kind, this is the backend of the model.\
+            choices=list(OutputFormat)[:2],
```
nv-braf, line 45: Why are the other output formats not allowed? I see this is a bug I introduced when creating the config (comparing to what is in parser). I'm wondering if it would be cleaner to have a different enum that specifies the valid user inputs for backends instead?

genai-perf/genai_perf/config/input/config_endpoint.py
```diff
@@ -56,6 +59,7 @@ def __init__(self) -> None:
         self.service_kind: Any = ConfigField(
```
nv-braf, line 59: If this will always be inferred then the creation of this field should be moved into the infer method and removed from parse().
You can then remove the check that looks for the user setting it, and also set the default to be `kserve` (if that simplifies the logic).

genai-perf/genai_perf/config/input/config_endpoint.py
dyastremsky: That's correct. We're removing the defaults in argparse very soon.

rmccorm4: > That's correct. We're removing the defaults in argparse in [a separate PR](https://nvidia.slack.com/archives/C07FW51CE1J/p1744302323474549).

I think this link is self-pointing at the slack link back to this PR, is that the correct link?

genai-perf/README.md
dyastremsky: There is, and the help menu should be able to provide that via the CLI (and I imagine the config does as well). But we have not been keeping this endpoint-type list up-to-date as we keep adding endpoints, so nvclip, multimodal, etc. are not on this list.

I would rather just mark that it is a string. This provides less information, but I would rather not mislead the users, risk our documentation continually getting out-of-step with implementation, or have a very long list here.

debermudez: As long as there is a clear error message if an invalid value is provided and the documentation has the proper values, this should be fine.

genai-perf/genai_perf/config/input/config_endpoint.py
dyastremsky: The other output formats are not Triton backends. Since the backend names == output format, I think that's the design that was used. If you look at the enum, you can see that it's broken up into categories with comments.

I have looked into redesigning this enum before, but it was never enough of a priority. If it's important to you, we can create a ticket to look into refactoring the enum.

nv-braf: Probably not worth the effort.

genai-perf/genai_perf/config/input/config_endpoint.py
dyastremsky: Whoops, thanks for pointing that out. Here is the new link: https://github.com/triton-inference-server/perf_analyzer/pull/350/files#diff-89352fdfacaf2688a3d610f2686415e446f1d2a916d705ef20612f385e416b90

Based on Ryan and Brian's feedback, I went ahead and just removed the inference of dynamic_grpc from grpc_method, which was making things a bit wonkier. It means that those using dynamic_grpc need to specify it as an endpoint, but I think that's the right choice to make everything consistent.

So now the defaults are Triton and kserve. If a user specifies a different endpoint-type (e.g. dynamic_grpc, completions, etc.), that will be used instead. This seems more consistent and maintainable.

genai-perf/genai_perf/config/input/config_endpoint.py
dyastremsky: Thanks for pointing this out! Updated.

genai-perf/genai_perf/config/input/config_endpoint.py
```diff
@@ -193,35 +173,17 @@ def infer_settings(self, model_name: str) -> None:
         self.infer_service_kind(model_name)
 
     def infer_service_kind(self, model_name: str) -> None:
-        if self.get_field("service_kind").is_set_by_user:
```
nv-braf: You should still be declaring self.type as a ConfigField in this method:
```
self.service_kind: Any = ConfigField(
            default=EndPointDefaults.SERVICE_KIND,
            choices=["triton", "openai", "tensorrtllm_engine", "dynamic_grpc"])
```
This will ensure that the field is never set to an illegal value.

genai-perf/genai_perf/config/input/config_endpoint.py
```diff
@@ -259,13 +220,3 @@ def _check_inferred_backend(self) -> None:
                 raise ValueError(
                     f"The backend should only be used with the following combination: 'service_kind: triton' & 'type: kserve'"
                 )
-
-    def _check_inferred_service_kind(self, endpoint_config: EndpointConfig) -> None:
-        # This is inferred and cannot be checked by the endpoint_config
-        if self.service_kind == "openai" and self.type == "generate":
```
dyastremsky, line 257: Reviewing this change now. I think generate operates differently than the other endpoint-types, because it currently expects the service_kind to be Triton but then wants to pass PA a service_kind of openai. We don't have CI testing for this, so it passed CI even though this endpoint would likely be broken. Cleaning this up and fixing it.

genai-perf/genai_perf/config/input/config_endpoint.py
dyastremsky: That's awesome. Done.

genai-perf/genai_perf/config/input/config_endpoint.py
dyastremsky: It looks like generate is already broken on the main branch. Since this matches the behavior on the main branch, we can leave it as is and merge this once approved. We're going to need to look at the flow of generate and fix it (orthogonal to these changes) before code freeze.

CC: @debermudez @nv-braf @rmccorm4

nv-braf (dismissed):
Looks good!

genai-perf/genai_perf/config/input/config_endpoint.py
```diff
@@ -22,6 +23,8 @@
 from genai_perf.inputs.input_constants import ModelSelectionStrategy, OutputFormat
 from genai_perf.utils import split_and_strip_whitespace
 
+logger = logging.getLogger(__name__)
```
debermudez, line 26: 😄

genai-perf/genai_perf/config/input/config_endpoint.py
debermudez: It will be a source of a bug at _some_ point. 
I think we make the ticket.

genai-perf/genai_perf/inputs/input_constants.py
```diff
@@ -69,7 +69,7 @@ class OutputFormat(Enum):
     # Other output formats
     ################################################################
     IMAGE_RETRIEVAL = "IMAGE_RETRIEVAL"
-    DYANMIC_GRPC = "DYANMIC_GRPC"
+    DYNAMIC_GRPC = "DYANMIC_GRPC"
```
debermudez, line 72: why one and not the other?

genai-perf/genai_perf/config/input/config_endpoint.py
debermudez: Agree on your approach. Thanks for catching this.

debermudez (commented):
2 small things. This is great work!

debermudez (dismissed):

debermudez (approved):

genai-perf/genai_perf/config/input/config_endpoint.py
dyastremsky: Thank you both for calling this out! Created TPA-1070.

PR #356, opened 2025-04-10
Fix bug in tensorrtllm PA config generation

nv-braf:
David found some PA config generation dead code, and after I inspected it I realized I missed a case when converting from args to config. This escaped because we don't have coverage for this in the CI.

I've removed the dead code and added in unit testing to cover this missing case.

### DIFF
--- a/genai-perf/genai_perf/config/generate/perf_analyzer_config.py
+++ b/genai-perf/genai_perf/config/generate/perf_analyzer_config.py
@@ -293,6 +293,8 @@ class PerfAnalyzerConfig:
                 protocol_args += ["--shape", "max_tokens:1", "--shape", "text_input:1"]
         elif config.endpoint.service_kind == "openai":
             protocol_args += ["-i", "http"]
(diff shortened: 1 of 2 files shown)


dyastremsky (approved):
Great work!

genai-perf/genai_perf/config/generate/perf_analyzer_config.py
```diff
@@ -293,6 +293,8 @@ def _add_protocol_args(self, config: ConfigCommand) -> List[str]:
                 protocol_args += ["--shape", "max_tokens:1", "--shape", "text_input:1"]
         elif config.endpoint.service_kind == "openai":
             protocol_args += ["-i", "http"]
+        elif config.endpoint.service_kind == "tensorrtllm_engine":
+            protocol_args += ["--service-kind", "triton_c_api", "--streaming"]
```
matthewkotila, line 297: Thanks for fixing this. I noticed this bug a week or so ago but forgot to file a ticket.

PR #357, opened 2025-04-11
Clean up dead code

dyastremsky:
I ran a static analyzer to detect dead code and removed a lot of the true positives.

A lot of the code here is able to removed due to the migration to using custom config classes for setting defaults and providing inference. Instead of pointing to the old defaults, point to the new ones in the config directory.


genai-perf/genai_perf/inputs/converters/base_converter.py
```diff
@@ -78,7 +76,7 @@ def _select_model_name(self, index: int) -> str:
                 f"Model selection strategy '{self.config.endpoint.model_selection_strategy}' is unsupported"
             )
 
-    def _get_max_tokens(self, optional_data: Dict[Any, Any]) -> int:
+    def _get_max_tokens(self, optional_data: Dict[Any, Any]) -> Union[int, None]:
```
debermudez, line 79: This is a signature change. Why was None added?

genai-perf/tests/test_data_parser/test_llm_profile_data_parser.py
```diff
@@ -210,7 +210,7 @@ def test_triton_llm_profile_data(
             - experiment 2: [3, 4]
         """
         config = ConfigCommand({"model_name": "test_model"})
-        config.tokenizer.name = DEFAULT_TOKENIZER
+        config.tokenizer.name = "hf-internal-testing/llama-tokenizer"
```
debermudez, line 213: This feels like it opens us to subtle errors.
Why did we move away from using a variable for the default tokenizer?

genai-perf/genai_perf/inputs/converters/base_converter.py
dyastremsky: Previously, we had a default output token mean that was [an integer](https://github.com/triton-inference-server/perf_analyzer/pull/357/files/1c2a07257ede5c4b186822c003935b41160a12b8) (-1). Now the default is None. That default was added as part of the Config changes.

So if we're returning the default value, it is equal to None.

genai-perf/tests/test_retrievers/test_synthetic_prompt_generator.py
```diff
@@ -28,24 +28,25 @@
...
-from genai_perf.tokenizer import DEFAULT_TOKENIZER, get_tokenizer
+from genai_perf.tokenizer import get_tokenizer
 
 
 class TestSyntheticPromptGenerator:
 
     def test_synthetic_prompt_default(self):
         config = ConfigCommand({"model_name": "test_model"})
-        config.tokenizer.name = DEFAULT_TOKENIZER
+        config.tokenizer.name = "gpt2"
```
debermudez, line 43: does using gpt2 instead of llama change the numbers at all in the test?

debermudez (commented):
couple small questions

genai-perf/tests/test_retrievers/test_synthetic_prompt_generator.py
dyastremsky: Not for these. It did for others, so I set the old default tokenizer to minimize the amount of change.

genai-perf/tests/test_data_parser/test_llm_profile_data_parser.py
dyastremsky: Multiple users were upset because the default tokenizer was set when no tokenizer was provided, resulting in the wrong numbers. It made more sense to require a tokenizer to always be set, unless it could be inferred from the model name (usually can be). So we avoid silent errors and improve the user experience.

It's only a very niche case where a user does not care about the numbers and therefore does not want to set the tokenizer, e.g. for testing or validation. If we want to declare a DEFAULT_TESTING_TOKENIZER just so that our tests all have the same tokenizer, we can do so, though I'd rather do that in a separate PR. I think the marginal utility would be negligible.

debermudez: I forgot about that case as well.
In light of that, this change makes way more sense.

debermudez (approved):

nv-braf (approved):
Assuming all tests pass, this looks good to me. Nice job!

genai-perf/tests/test_converters/test_dynamic_grpc_converter.py
```diff
@@ -139,7 +139,7 @@ def test_check_config_valid(valid_config):
 
 
 def test_check_config_invalid_batch_size(valid_config):
-    valid_config.input.batch_size = [2]
+    valid_config.input.batch_size = InputDefaults.BATCH_SIZE + 1
```
nv-hwoo, line 142: How did it change from a list to an integer?

nv-hwoo (commented):
Very nice clean up! Just one question and looks good overall.

genai-perf/tests/test_converters/test_dynamic_grpc_converter.py
dyastremsky: Good catch, fixed! I don't know how this happened.

nv-hwoo (approved):
LGTM

PR #359, opened 2025-04-14
Speed up LLM metrics parsing by 60%

dyastremsky:
Speed up metrics parsing with a variety of small improvements. In the testing, this sped up metrics parsing by 60% (see below). This is still single-threaded to avoid unnecessary complexities like initializing multiple tokenizers, though we can move towards multi-threaded options in the future.

Updates:
- Use orjson for faster JSON parsing
- Implement caching for repeated input/outputs
- Cache repeated append calls and variable lookups
- Batch getting output counts
- Reduce TQDM progress updates to every 1%

The speedup will be significantly greater in certain cases:
- Larger number of requests due to the caching of repeated calls and variable lookups plus the TQDM bar being updated much less often.
- Repeated responses (e.g. due to K-V caching) due to the caching of encodes. In this testing, no responses were repeated, so the caching only benefitted the repeated inputs. In cases where responses are more deterministic, the speedup will be much greater.
- Larger inputs/outputs due to the orjson library being ~6x faster than the JSON library, especially for larger texts.

Before:
image

After:
image


nv-braf (dismissed):
This all looks fine to me. Nice job!

lkomali (dismissed):
LGTM

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
```diff
@@ -349,7 +368,11 @@ def _get_input_token_count(self, req_inputs: dict) -> int:
         else:
             raise ValueError(f"Unknown service kind: '{self._service_kind}'.")
 
-        return len(self._tokenizer.encode(input_text))
+        return self._cached_encode(input_text)
+
+    @lru_cache(maxsize=100_000)
```
nicolasnoble, line 373: Should we move this to a named constant somewhere instead?

genai-perf/genai_perf/utils.py
```diff
@@ -97,8 +97,8 @@ def load_json_str(json_str: str, func: Callable = lambda x: x) -> Dict[str, Any]
           run validation checks on the object. Defaults to identity function.
     """
     try:
-        return func(json.loads(json_str))
-    except json.JSONDecodeError:
+        return func(orjson.loads(json_str))
```
nicolasnoble, line 100: As far as I know, from a quick search, orjson doesn't strictly parses json the same exact way as the stock Python implementation, especially regarding utf-8 sequences. Maybe add a comment up to indicate this detail, and to refer the reader to orjson's documentation for the specificities of its parsing?

nicolasnoble (approved):

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
dyastremsky: Sure, done!

genai-perf/genai_perf/utils.py
dyastremsky: Good point. Added a comment!

genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
```diff
@@ -134,43 +141,56 @@ def _parse_requests(self, requests: dict) -> LLMMetrics:
             if not res_timestamps:
                 continue
 
+            # Cache length
+            num_responses = len(res_timestamps)
+
+            # Cache frequently used appends
+            request_latencies_append = request_latencies.append
```
debermudez, line 148: how does this cache the append?

dyastremsky: It caches the method lookup for .append. Every time the function gets called, the interpreter in Python looks up the append attribute for this list before calling it. By caching it, we do the lookup once, then it just directly calls it each time. It reduces some overhead.

I tested the performance after each of these micro-optimizations, and this one reduces parsing time by ~5%. That number would increase as the number of requests grows.

debermudez: well thats really handy. 
nice job with this PR.

debermudez (approved):

PR #363, opened 2025-04-18
Process export files subcommand

lkomali:
This is a PR for process-export-files feature.

It has all previous PRs for this work and I fixed any errors, unit tests with new changes to config. 

Added support for --input-directory/-d to make the option consistent with all other options.
1. Using process-export-files subcommand

image

2. Using config file

image


genai-perf/genai_perf/config/input/base_config.py
```diff
@@ -56,6 +57,9 @@ def any_field_set_by_user(self) -> bool:
 
     def check_required_fields_are_set(self) -> None:
         for name, field in self._fields.items():
+            # Skip validation model_names is set for process-export-files subcommand
```
nv-braf, line 60: This isn't the right level to have this check. The BaseConfig class should have no knowledge about decisions we are making when creating config files (think of it as a library).
I think the proper course of action here is to remove the requirement of model name being set from ConfigCommand if this subcomand is set in the check_required_fields_are_set() in that file.

```
def _check_required_fields_are_set(self) -> None:
        self.get_field("model_name").required = False
        super().check_required_fields_are_set()
```

genai-perf/genai_perf/config/input/create_config.py
```diff
@@ -95,11 +96,16 @@ def _add_cli_options_to_config(
 
         CreateConfig._check_that_override_is_set(args)
         CreateConfig._add_top_level_args_to_config(config, args)
+        CreateConfig._add_output_args_to_config(config, args)
```
nv-braf, line 99: This isn't a clean way to do this as you are relying on an implicit ordering of the function calls. A better way is to make this:
```
if config.subcommand == Subcommand.PROCESS:
   CreateConfig._add_process_args_to_config()
else:
   CreateConfig._add_profile_analyze_args_to_config()
```
And then call the exact args_to_config methods you need from these new methods

genai-perf/genai_perf/config/input/create_config.py
```diff
@@ -133,8 +139,9 @@ def _check_that_override_is_set(args: argparse.Namespace) -> None:
     def _add_top_level_args_to_config(
         config: ConfigCommand, args: argparse.Namespace
     ) -> ConfigCommand:
-        if args.model:
-            config.model_names = args.model
+        if args.subcommand != Subcommand.PROCESS.value:
```
nv-braf, line 142: I think it's cleaner to do:
```
if hasattr(args, "model") and args.model:
```

genai-perf/genai_perf/parser.py
```diff
@@ -154,6 +155,15 @@ def _create_sweep_list(args):
 
 
 ### Types ###
+
+
+def directory(value: str) -> Path:
```
nv-braf, line 160: We don't want type checking in parser anymore, this checking should be done in ConfigCommand (or it's children).

genai-perf/genai_perf/subcommand/subcommand.py
```diff
@@ -179,7 +194,8 @@ def _create_tokenizer(self) -> None:
         if self._tokenizer:
             return
 
-        logger.info(f"Creating tokenizer for: {self._config.tokenizer.name}")
+        if self._config.subcommand != ic.Subcommand.PROCESS:
```
nv-braf, line 197: Why don't you want to print the logging message if you are creating a tokenizer?

nv-braf (commented):
I know this is still a draft, but wanted to make some comments before I leave on vacation. Overall this look really good. Nice job!

lkomali:
@nv-braf I was going to put it into review this morning. Thank you for reviewing, I will make those changes and then mark for review.

genai-perf/genai_perf/subcommand/subcommand.py
lkomali: yes i can print this, will remove.

genai-perf/genai_perf/parser.py
lkomali: But we still do --input-file type checking in parser.py.

genai-perf/genai_perf/subcommand/process_export_files.py
```diff
@@ -0,0 +1,169 @@
...
+
+class ProcessExportFiles(Subcommand):
+    """
+    Contains all the methods needed to run the process-export-files subcommand
+    """
+
+    def __init__(self, config: ConfigCommand, extra_args: Optional[List[str]]) -> None:
+        super().__init__(config, extra_args)
+
+    def process_export_files(self) -> None:
```
debermudez, line 60: I like the way this is done. It reads like a story. Well done

genai-perf/genai_perf/subcommand/subcommand.py
```diff
@@ -75,6 +77,18 @@ def __init__(
         # and are used by multiple methods
         self._data_parser: Optional[ProfileDataParser] = None
 
+    def create_plots(self) -> None:
```
debermudez, line 80: I thought we removed plotting entirely.
Does this work?

genai-perf/genai_perf/parser.py
lkomali: Oh! I moved it ConfigCommand but probably missed removing it from parser.py. I will update.

genai-perf/genai_perf/config/input/config_command.py
```diff
@@ -230,6 +241,14 @@ def _check_profile_export_file(self) -> None:
                     "the profile_export_file."
                 )
 
+    def _check_input_path_is_valid(self) -> None:
```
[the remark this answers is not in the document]
lkomali: This is just for this particular subcommand. This is a refactor so that when somebody is using process-export-files subcommand through CLI or config, this type check happens.

genai-perf/genai_perf/config/input/config_process.py
```diff
@@ -0,0 +1,49 @@
...
+        self.input_path: Any = ConfigField(
+            default=ProcessDefaults.INPUT_PATH,
+            add_to_template=False,
+            template_comment=process_export_files_comment,
+        )
+
+    ###########################################################################
+    # Parsing Methods
+    ###########################################################################
+    def parse(self, process: Dict[str, Any]) -> None:
```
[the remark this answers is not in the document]
lkomali: No, I don't remember the specifics correctly because i added this check a while ago, but there's a case where the process.parse is called but throws an error because process object is None.
There's a similar check in config_analyze.py too.
I unfortunately don't exactly remember the use case when it happens.

genai-perf/genai_perf/subcommand/subcommand.py
lkomali: Did we ? I'm not aware. I did not check, this was there in profile subcommand, so i assumed plotting exists.

lkomali:
I do have a CI test but I delayed raising an MR for it, because I'm actively working on processing multiple files now and the CI test would soon fail once that part is done. So, I have the CI test locally, will update it to work for processing multiple files and raise an MR for it, when that work is done.

genai-perf/genai_perf/parser.py
lkomali: I removed it from parser.py.

genai-perf/genai_perf/config/input/config_process.py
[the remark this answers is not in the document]
debermudez: okay. I think we would want this refactored at some point, but not given the current lifetime of the tool.

genai-perf/genai_perf/subcommand/subcommand.py
debermudez: can you check this please? either remove it or verify this works.

debermudez (approved):

PR #365, opened 2025-04-23
Add tutorial for benchmarking Dynamo

AndyDai-nv:
Tutorial to build Dynamo and benchmark Dynamo with GenAI-Perf. Please try it and let me know how I can improve it.

### DIFF
--- a/genai-perf/docs/dynamo.md
+++ b/genai-perf/docs/dynamo.md
@@ -1,0 +1,165 @@
+<!--
+Copyright (c) 2024-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+ * Redistributions of source code must retain the above copyright
+   notice, this list of conditions and the following disclaimer.
+ * Redistributions in binary form must reproduce the above copyright
+   notice, this list of conditions and the following disclaimer in the
+   documentation and/or other materials provided with the distribution.
+ * Neither the name of NVIDIA CORPORATION nor the names of its
+   contributors may be used to endorse or promote products derived
+   from this software without specific prior written permission.
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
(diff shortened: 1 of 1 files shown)


genai-perf/docs/dynamo.md
matthewkotila, line 1: @AndyDai-nv this looks great on the whole--very readable!

A few questions.

1. Is there a simpler way to get Dynamo installed? Can it just be pip installed from pypi?
2. What's the difference between dynamo run and dynamo serve?
2a. What seems peculiar about dynamo serve is that the model name doesn't show up in the serve command--how does `deepseek-ai/DeepSeek-R1-Distill-Llama-8B` become loaded when running `dynamo serve graphs.agg:Frontend -f ./configs/agg.yaml`?

lkomali:
Overall, the doc looks good to me.

genai-perf/docs/dynamo.md
AndyDai-nv: 1. Yes, but I didn't test, refer to [these](https://github.com/ai-dynamo/dynamo/blob/main/README.md?plain=1#L41-L47). The current build can also be used for making code changes to dynamo, so I used this.

2. In my understanding, `dynamo run` is more like a quickstart, without setting `in=http` flags, it offers a chat interface for users to quickly try it out. While `dynamo serve` is more formal, especially when it involves more inference components and more complicated serving scenario which can be defined in the config. But for these two examples in tutorial, I think they are just similar and simple http servers.

matthewkotila: Why not use the simpler installation for the tutorial?

AndyDai-nv: I encountered this [issue](https://github.com/ai-dynamo/dynamo/issues/785) using that way, seems like the suggested way is to build the container from scratch.

matthewkotila: Oooh, that sounds bad.

So the easiest installation for dynamo is currently broken, it sounds like?

Ok, I guess we gotta use the more complicated installation, as you said. Bummer.

matthewkotila (approved):

PR #366, opened 2025-04-23
Add support for fixed schedule warmup

matthewkotila:
Previously, someone would run PA fixed schedule mode like this:

```bash
perf_analyzer \
  --fixed-schedule \
  --input-data=input_data.json \
  -m facebook/opt-125m \
  --service-kind=openai \
  --endpoint=v1/chat/completions \
  --async
```

with an `input_data.json` like this:

```json
{
  "data": [
    {
      "payload": [{
          "model": "facebook/opt-125m",
          "messages": [{"role": "user","content": "my_prompt_1"}],
          "max_completion_tokens": 1
        }],
      "timestamp": [1000]
    },
    {
      "payload": [{
          "model": "facebook/opt-125m",
          "messages": [{"role": "user","content": "my_prompt_2"}],
          "max_completion_tokens": 1
        }],
      "timestamp": [2000]
    }
  ]
}
```

But they wouldn't be able to use the warmup feature ([`--warmup-request-count`](https://github.com/triton-inference-server/perf_analyzer/blob/main/docs/cli.md#--warmup-request-countn)).

Now, with this PR, users can use the warmup feature with the fixed schedule feature:

```bash
perf_analyzer \
  --fixed-schedule \
  --warmup-request-count=1 \
  --input-data=input_data.json \
  -m facebook/opt-125m \
  --service-kind=openai \
  --endpoint=v1/chat/completions \
  --async
```

Basically, if `--warmup-request-count=N`, the first `N` payloads in `input_data.json` are sent as "warmup" requests (i.e. excluded from the final performance metric/statistic calculations and profile export JSON), and the rest are part of the standard benchmark.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -291,6 +291,17 @@ duration of execution exceeds the amount of time specified by the intervals.
 This option can not be used with `--request-rate-range` or
 `--concurrency-range`.
 
+#### `--fixed-schedule`
+
+Enables fixed schedule inference load mode. In this mode, Perf Analyzer runs
+through the `--input-data` JSON once. Each entry represents a request and must
+have a `timestamp` "input", which tells Perf Analyzer when to send that request.
+A timestamp of `N` means Perf Analyzer will send that request `N` milliseconds
+after the start of the benchmark.
+
+See [fixed schedule](inference_load_modes.md#fixed-schedule-mode) documentation
+for more info.
+
 #### `--max-threads=<n>`
 
 Specifies the maximum number of threads that will be created for providing
--- a/docs/inference_load_modes.md
+++ b/docs/inference_load_modes.md
@@ -1,5 +1,5 @@
 <!--
-Copyright (c) 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+Copyright (c) 2023-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions
@@ -98,3 +98,45 @@ where `my_intervals.txt` contains:
 
 Perf Analyzer will attempt to send requests at the following times: 0.1s, 0.3s,
 0.8s, 0.9s, 1.1s, 1.6s, and so on, during profiling.
+
+## Fixed Schedule Mode
+
+In fixed schedule inference load mode, Perf Analyzer runs through the
+[`--input-data`](cli.md#--input-datazerorandompath) JSON once. Each entry
+represents a request and must have a `timestamp` "input", which tells Perf
+Analyzer when to send that request. A timestamp of `N` means Perf Analyzer will
+send that request `N` milliseconds after the start of the benchmark. For
+example:
+
+```json
+# input_data.json
+{
+  "data": [
+    {
+      "payload": [{
+          "model": "facebook/opt-125m",
+          "messages": [{"role": "user","content": "What is the capital of France?"}],
+          "max_completion_tokens": 1
+        }],
+      "timestamp": [1000]
+    },
+    {
+      "payload": [{
+          "model": "facebook/opt-125m",
+          "messages": [{"role": "user","content": "What is 1+1?"}],
+          "max_completion_tokens": 1
+        }],
+      "timestamp": [2000]
+    }
+  ]
+}
+```
+
+Using this `input_data.json` with fixed schedule mode will run a benchmark that
+sends 2 requests total.
+
+The first request gets sent 1000ms after the start of the benchmark with the
+prompt `What is the capital of France?`.
+
+The second request gets sent 2000ms after the start of the benchmark with the
+prompt `What is 1+1?`.
--- a/genai-perf/genai_perf/config/generate/perf_analyzer_config.py
+++ b/genai-perf/genai_perf/config/generate/perf_analyzer_config.py
@@ -246,12 +246,16 @@ class PerfAnalyzerConfig:
     def _add_perf_analyzer_args(self, config: ConfigCommand) -> List[str]:
         perf_analyzer_args = []
 
+        if config.perf_analyzer.warmup_request_count > 0:
+            perf_analyzer_args += [
+                "--warmup-request-count",
+                f"{config.perf_analyzer.warmup_request_count}",
+            ]
+
         if config.input.prompt_source != PromptSource.PAYLOAD:
             perf_analyzer_args += [
                 f"--stability-percentage",
                 f"{config.perf_analyzer.stability_percentage}",
-                f"--warmup-request-count",
-                f"{config.perf_analyzer.warmup_request_count}",
             ]
 
             mode = config.perf_analyzer.measurement.mode
@@ -290,8 +294,6 @@ class PerfAnalyzerConfig:
                 protocol_args += ["--shape", "max_tokens:1", "--shape", "text_input:1"]
         elif config.endpoint.service_kind == "openai":
             protocol_args += ["-i", "http"]
-        elif config.endpoint.service_kind == "tensorrtllm_engine":
-            protocol_args += ["--service-kind", "triton_c_api", "--streaming"]
 
         return protocol_args
 
@@ -346,7 +348,10 @@ class PerfAnalyzerConfig:
 
     def _add_endpoint_args(self, config: ConfigCommand) -> List[str]:
         endpoint_args = []
-        endpoint_args += ["--service-kind", f"{config.endpoint.service_kind}"]
+        if config.endpoint.service_kind == "tensorrtllm_engine":
+            endpoint_args += ["--service-kind", "triton_c_api", "--streaming"]
+        else:
+            endpoint_args += ["--service-kind", f"{config.endpoint.service_kind}"]
 
         if config.endpoint.custom:
             endpoint_args += ["--endpoint", f"{config.endpoint.custom}"]
--- a/genai-perf/genai_perf/config/input/config_command.py
+++ b/genai-perf/genai_perf/config/input/config_command.py
@@ -207,10 +207,6 @@ class ConfigCommand(BaseConfig):
                         f"User Config: perf_analyzer.stimulus: {key} is not supported with the payload input source."
                     )
 
-        if self.perf_analyzer.get_field("warmup_request_count").is_set_by_user:
-            raise ValueError(
-                f"User Config: perf_analyzer.warmup_request_count is not supported with the payload input source."
-            )
         if (
             self.perf_analyzer.measurement.get_field("mode").is_set_by_user
             and self.perf_analyzer.measurement.mode
--- a/genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
+++ b/genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
@@ -30,7 +30,7 @@ from genai_perf.config.input.config_defaults import InputDefaults
 from genai_perf.exceptions import GenAIPerfException
 from genai_perf.inputs.converters.base_converter import BaseConverter
 from genai_perf.inputs.input_constants import DEFAULT_TENSORRTLLM_MAX_TOKENS
-from genai_perf.inputs.retrievers.generic_dataset import GenericDataset
+from genai_perf.inputs.retrievers.generic_dataset import DataRow, GenericDataset
 from genai_perf.utils import sample_bounded_normal
 
 
@@ -112,3 +112,10 @@ class TensorRTLLMEngineConverter(BaseConverter):
 
     def _construct_default_template(self, prompt: str) -> List[Dict[str, str]]:
         return [{"role": "user", "content": prompt}]
+
+    def _add_payload_optional_data(self, payload: Dict[Any, Any], row: DataRow) -> None:
+        for key, value in row.optional_data.items():
+            if key == "max_tokens":
+                payload["request_output_len"] = [value]
+            else:
+                payload[key] = value
--- a/genai-perf/tests/test_cli.py
+++ b/genai-perf/tests/test_cli.py
@@ -1115,10 +1115,6 @@ class TestCLIArguments:
                 ["--request-count", "3"],
                 "User Config: perf_analyzer.measurement.mode of request_count is not supported with the payload input source.",
             ),
-            (
-                ["--warmup-request-count", "7"],
-                "User Config: perf_analyzer.warmup_request_count is not supported with the payload input source.",
-            ),
         ],
     )
     def test_check_payload_input_args_invalid_args(
--- a/genai-perf/tests/test_perf_analyzer_config.py
+++ b/genai-perf/tests/test_perf_analyzer_config.py
@@ -110,8 +110,6 @@ class TestPerfAnalyzerConfig(unittest.TestCase):
             "999",
             "--request-count",
             "128",
-            "--warmup-request-count",
-            "0",
             "--input-data",
             "artifacts/test_model-triton-tensorrtllm-runtime_batch_size1-concurrency64/inputs.json",
             "--profile-export-file",
@@ -144,8 +142,6 @@ class TestPerfAnalyzerConfig(unittest.TestCase):
                 "999",
                 "--request-count",
                 "128",
-                "--warmup-request-count",
-                "0",
                 "--streaming",
                 "--shape",
                 "max_tokens:1",
@@ -300,10 +296,10 @@ class TestPerfAnalyzerConfig(unittest.TestCase):
         self._config.perf_analyzer.warmup_request_count = 30
 
         expected_args = [
-            "--stability-percentage",
-            "567",
             "--warmup-request-count",
             "30",
+            "--stability-percentage",
+            "567",
             "--measurement-interval",
             "5000",
         ]
@@ -350,21 +346,6 @@ class TestPerfAnalyzerConfig(unittest.TestCase):
         self._config.endpoint.service_kind = "openai"
 
         expected_args = ["-i", "http"]
-
-        actual_args = self._default_perf_analyzer_config._add_protocol_args(
-            self._config
-        )
-
-        self.assertEqual(expected_args, actual_args)
-
-    def test_add_protocol_args_with_tensorrtllm_engine(self):
-        """
-        Test that _add_protocol_args returns the correct arguments
-        when service_kind is 'tensorrtllm_engine'
-        """
-        self._config.endpoint.service_kind = "tensorrtllm_engine"
-
-        expected_args = ["--service-kind", "triton_c_api", "--streaming"]
 
         actual_args = self._default_perf_analyzer_config._add_protocol_args(
             self._config
@@ -528,6 +509,21 @@ class TestPerfAnalyzerConfig(unittest.TestCase):
         self._config.endpoint.custom = None
 
         expected_args = ["--service-kind", "triton"]
+
+        actual_args = self._default_perf_analyzer_config._add_endpoint_args(
+            self._config
+        )
+
+        self.assertEqual(expected_args, actual_args)
+
+    def test_add_endpoint_args_with_tensorrtllm_engine_endpoint(self):
+        """
+        Test that _add_endpoint_args returns the correct arguments
+        when tensorrtllm_engine endpoint is set
+        """
+        self._config.endpoint.service_kind = "tensorrtllm_engine"
+
+        expected_args = ["--service-kind", "triton_c_api", "--streaming"]
 
         actual_args = self._default_perf_analyzer_config._add_endpoint_args(
             self._config
--- a/src/command_line_parser.cc
+++ b/src/command_line_parser.cc
@@ -139,6 +139,7 @@ CLParser::Usage(const std::string& msg)
   std::cerr << "\t--request-count <number of requests>" << std::endl;
   std::cerr << "\t--warmup-request-count <number of warmup requests>"
             << std::endl;
+  std::cerr << "\t--fixed-schedule" << std::endl;
   std::cerr << "\tDEPRECATED OPTIONS" << std::endl;
   std::cerr << "\t-t <number of concurrent requests>" << std::endl;
   std::cerr << "\t-c <maximum concurrency>" << std::endl;
@@ -490,6 +491,16 @@ CLParser::Usage(const std::string& msg)
                    "which means that no warmup requests will be sent.",
                    18)
             << std::endl;
+  std::cerr
+      << FormatMessage(
+             " --fixed-schedule: Enables fixed schedule inference load mode. "
+             "In this mode, Perf Analyzer runs through the `--input-data` JSON "
+             "once. Each entry represents a request and must have a "
+             "`timestamp` \"input\", which tells Perf Analyzer when to send "
+             "that request. A timestamp of `N` means Perf Analyzer will send "
+             "that request `N` milliseconds after the start of the benchmark.",
+             18)
+      << std::endl;
   std::cerr << FormatMessage(
                    " --serial-sequences: Enables serial sequence mode "
                    "where a maximum of one request is outstanding at a time "
@@ -1707,7 +1718,7 @@ CLParser::ParseCommandLine(int argc, char** argv)
           break;
         }
         case long_option_idx_base + 62: {
-          if (std::stoi(optarg) < 0) {
+          if (std::stoi(optarg) <= 0) {
             Usage("Failed to parse --request-count. The value must be > 0.");
           }
           if (params_->request_count == 0) {
@@ -1912,8 +1923,21 @@ CLParser::ParseCommandLine(int argc, char** argv)
   }
 
   // In fixed schedule mode, the request count is based on the number of
-  // payloads in the user-provided dataset
+  // payloads in the user-provided dataset (minus warmup request count)
   if (params_->inference_load_mode == pa::InferenceLoadMode::FixedSchedule) {
+    if (params_->request_count != 0) {
+      throw std::runtime_error(
+          "Cannot specify `--request-count` when using `--fixed-schedule`. The "
+          "request count will be equal to the number of requests in the input "
+          "data minus the warmup request count.");
+    }
+
+    if (params_->user_data.empty()) {
+      throw std::runtime_error(
+          "When using `--fixed-schedule`, you must provide a dataset using "
+          "`--input-data`.");
+    }
+
     for (const auto& input_data_path : params_->user_data) {
       if (!std::filesystem::is_regular_file(input_data_path)) {
         throw std::runtime_error(
@@ -1923,7 +1947,22 @@ CLParser::ParseCommandLine(int argc, char** argv)
       }
     }
 
-    params_->request_count = pa::DataLoader::GetDatasetSize(params_->user_data);
+    const size_t dataset_size{
+        pa::DataLoader::GetDatasetSize(params_->user_data)};
+
+    if (dataset_size == 0) {
+      throw std::runtime_error(
+          "The dataset provided is empty. When using `--fixed-schedule`, the "
+          "dataset must contain at least one payload.");
+    }
+
+    if (params_->warmup_request_count >= dataset_size) {
+      throw std::runtime_error(
+          "The warmup request count must be less than the total number of "
+          "payloads in the dataset.");
+    }
+
+    params_->request_count = dataset_size - params_->warmup_request_count;
   }
 
   // When the request-count feature is enabled, override the measurement mode to
@@ -1984,11 +2023,6 @@ CLParser::VerifyOptions()
   }
   if (params_->async && params_->forced_sync) {
     Usage("Cannot specify --async and --sync simultaneously.");
-  }
-
-  if (params_->inference_load_mode == InferenceLoadMode::FixedSchedule &&
-      params_->warmup_request_count > 0) {
-    Usage("Cannot use warmup options with --fixed-schedule");
   }
 
   if (params_->inference_load_mode == InferenceLoadMode::PeriodicConcurrency &&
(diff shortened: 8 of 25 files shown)


docs/cli.md
```diff
@@ -291,6 +291,17 @@ duration of execution exceeds the amount of time specified by the intervals.
...
+
+Enables fixed schedule inference load mode. In this mode, Perf Analyzer runs
+through the `--input-data` JSON once. Each entry represents a request and must
+have a `timestamp` "input", which tells Perf Analyzer when to send that request.
+A timestamp of `N` means Perf Analyzer will send that request `N` milliseconds
+after the start of the benchmark.
+
+See [fixed schedule](inference_load_modes.md#fixed-schedule-mode) documentation
+for more info.
+
```
matthewkotila, line 304: Fixed schedule mode existed before this PR, but its docs were forgotten.

Just adding docs here.

docs/inference_load_modes.md
```diff
@@ -98,3 +98,45 @@ where `my_intervals.txt` contains:
 
 Perf Analyzer will attempt to send requests at the following times: 0.1s, 0.3s,
 0.8s, 0.9s, 1.1s, 1.6s, and so on, during profiling.
+
+## Fixed Schedule Mode
```
matthewkotila, line 102: Again, docs were missing for this feature, so I added them in this PR.

genai-perf/genai_perf/config/generate/perf_analyzer_config.py
```diff
@@ -246,12 +246,16 @@ def _add_required_args(self, config: ConfigCommand) -> List[str]:
...
+                "--warmup-request-count",
+                f"{config.perf_analyzer.warmup_request_count}",
+            ]
+
         if config.input.prompt_source != PromptSource.PAYLOAD:
             perf_analyzer_args += [
                 f"--stability-percentage",
                 f"{config.perf_analyzer.stability_percentage}",
-                f"--warmup-request-count",
-                f"{config.perf_analyzer.warmup_request_count}",
```
matthewkotila, line 254: Previously, the warmup feature wasn't supported in GAP "payload" mode, but now it is.

In the old code, warmup could only get appending to the PA command if "payload" mode _wasn't_ being used.

Now, valid warmup gets appending to PA command regardless of "payload" mode being used.

genai-perf/genai_perf/config/input/config_command.py
```diff
@@ -207,10 +207,6 @@ def _check_payload_input(self) -> None:
                         f"User Config: perf_analyzer.stimulus: {key} is not supported with the payload input source."
                     )
 
-        if self.perf_analyzer.get_field("warmup_request_count").is_set_by_user:
-            raise ValueError(
-                f"User Config: perf_analyzer.warmup_request_count is not supported with the payload input source."
-            )
```
matthewkotila, line 213: As mentioned earlier, GAP used to not support using fixed schedule mode with the warmup feature, hence this error message.

Now, it does, so the error message/check is removed.

genai-perf/genai_perf/inputs/converters/tensorrtllm_engine_converter.py
```diff
@@ -112,3 +112,10 @@ def _encode_with_chat_template(self, prompt: str) -> List[int]:
 
     def _construct_default_template(self, prompt: str) -> List[Dict[str, str]]:
         return [{"role": "user", "content": prompt}]
+
+    def _add_payload_optional_data(self, payload: Dict[Any, Any], row: DataRow) -> None:
+        for key, value in row.optional_data.items():
+            if key == "max_tokens":
+                payload["request_output_len"] = [value]
+            else:
+                payload[key] = value
```
matthewkotila, line 121: The payload input retriever is responsible for parsing/storing `output_length` in the users passed file via `--input-file payload:my_input_file.jsonl`.

It currently stores it under the `max_tokens` in the `optional_data` that is passed to the converter.

Previously, every converter (via base_converter) would automatically write every key value pair in `optional_data` exactly as it is to the outgoing payload (given to PA via input data JSON).

This caused issues for PA when OpenAI mode wasn't being used. In the case of tensorrtllm_engine mode, PA cannot have an input named `max_tokens` with a single scalar integer value. The "maximum tokens" needs to be written to an input named `request_output_len` as an array with one integer element.

This new code overrides the base_converter logic to handle this special case for tensorrtllm_engine mode.

genai-perf/tests/test_cli.py
```diff
@@ -1115,10 +1115,6 @@ def test_positive_audio_input_args(self, monkeypatch, args):
                 ["--request-count", "3"],
                 "User Config: perf_analyzer.measurement.mode of request_count is not supported with the payload input source.",
             ),
-            (
-                ["--warmup-request-count", "7"],
-                "User Config: perf_analyzer.warmup_request_count is not supported with the payload input source.",
-            ),
```
matthewkotila, line 1121: Again, GAP now supports payload mode with warmup feature.

No need for this test that checks for error message on using both features.

genai-perf/tests/test_perf_analyzer_config.py
```diff
@@ -110,8 +110,6 @@ def test_command_creation(self):
             "999",
             "--request-count",
             "128",
-            "--warmup-request-count",
-            "0",
```
matthewkotila, line 114: GAP no longer writes `--warmup-request-count` when its value is 0. It's unnecessary from PA perspective.

https://github.com/triton-inference-server/perf_analyzer/pull/366/files#diff-d1c5f6fe3374963d887eec07bbc9270be4be7ca81e1c6ccf2632c50027204eb3R249-R253

genai-perf/tests/test_perf_analyzer_config.py
```diff
@@ -300,10 +296,10 @@ def test_add_perf_analyzer_args_with_non_payload_prompt_source(self):
         self._config.perf_analyzer.warmup_request_count = 30
 
         expected_args = [
-            "--stability-percentage",
-            "567",
             "--warmup-request-count",
             "30",
+            "--stability-percentage",
+            "567",
```
matthewkotila, line 302: My logic change here:

https://github.com/triton-inference-server/perf_analyzer/pull/366/files#diff-d1c5f6fe3374963d887eec07bbc9270be4be7ca81e1c6ccf2632c50027204eb3R249-R253

Caused these options to be appended in a different order, nothing changes otherwise.

genai-perf/tests/test_perf_analyzer_config.py
```diff
@@ -535,6 +516,21 @@ def test_add_endpoint_args_without_custom_endpoint(self):
...
+        self._config.endpoint.service_kind = "tensorrtllm_engine"
+
+        expected_args = ["--service-kind", "triton_c_api", "--streaming"]
+
+        actual_args = self._default_perf_analyzer_config._add_endpoint_args(
+            self._config
+        )
+
+        self.assertEqual(expected_args, actual_args)
+
```
matthewkotila, line 533: As mentioned [here](https://github.com/triton-inference-server/perf_analyzer/pull/366/files#r2057039805), GAP now appends `tensorrtllm_engine`-related PA options in `_add_endpoint_args()` instead of the previous `_add_protocol_args()`, so this test was added.

src/command_line_parser.cc
```diff
@@ -139,6 +139,7 @@ CLParser::Usage(const std::string& msg)
   std::cerr << "\t--request-count <number of requests>" << std::endl;
   std::cerr << "\t--warmup-request-count <number of warmup requests>"
             << std::endl;
+  std::cerr << "\t--fixed-schedule" << std::endl;
```
matthewkotila, line 142: Similar to [this](https://github.com/triton-inference-server/perf_analyzer/pull/366/files#r2057033395), PA was accidentally missing help option documentation for fixed schedule mode.

Adding in this PR.

src/command_line_parser.cc
```diff
@@ -490,6 +491,16 @@ CLParser::Usage(const std::string& msg)
...
+  std::cerr
+      << FormatMessage(
+             " --fixed-schedule: Enables fixed schedule inference load mode. "
+             "In this mode, Perf Analyzer runs through the `--input-data` JSON "
+             "once. Each entry represents a request and must have a "
+             "`timestamp` \"input\", which tells Perf Analyzer when to send "
+             "that request. A timestamp of `N` means Perf Analyzer will send "
+             "that request `N` milliseconds after the start of the benchmark.",
+             18)
+      << std::endl;
```
matthewkotila, line 503: Again, PA was missing help option docs for fixed schedule.

Adding in this PR.

src/command_line_parser.cc
```diff
@@ -1707,7 +1718,7 @@ CLParser::ParseCommandLine(int argc, char** argv)
           break;
         }
         case long_option_idx_base + 62: {
-          if (std::stoi(optarg) < 0) {
+          if (std::stoi(optarg) <= 0) {
```
matthewkotila, line 1721: This was a random "bug" fix.

As the error message below says, user should not be able to specify request count of zero. It's means no benchmark gets run, which is useless.

src/command_line_parser.cc
```diff
@@ -1912,8 +1923,21 @@ CLParser::ParseCommandLine(int argc, char** argv)
...
+          "request count will be equal to the number of requests in the input "
+          "data minus the warmup request count.");
+    }
+
+    if (params_->user_data.empty()) {
+      throw std::runtime_error(
+          "When using `--fixed-schedule`, you must provide a dataset using "
+          "`--input-data`.");
+    }
+
```
matthewkotila, line 1940: As a bit of a wrinkle when adding warmup support for fixed schedule, the number of requests in the input data JSON given to PA needs to be known quite early in the application flow.

This code quickly reads the JSON to get the number of entries, which is then used for validation purposes and later for proper thread allocation.

This wasn't my ideal solution, but it was the best middle ground I could come up with without severely redesigning other parts of PA.

matthewkotila, line 1940: Just some UX cleanup/validation.

PA now forbids user from specifying `--request-count` in `--fixed-schedule` mode.

Fixed schedule mode deduces request count based on the dataset size minus the specified warmup request count.

src/command_line_parser.cc
```diff
@@ -1923,7 +1947,22 @@ CLParser::ParseCommandLine(int argc, char** argv)
...
+          "dataset must contain at least one payload.");
+    }
+
+    if (params_->warmup_request_count >= dataset_size) {
+      throw std::runtime_error(
+          "The warmup request count must be less than the total number of "
+          "payloads in the dataset.");
+    }
+
+    params_->request_count = dataset_size - params_->warmup_request_count;
```
matthewkotila, line 1965: More validation to catch dataset size/invalid option values earlier.

src/command_line_parser.cc
```diff
@@ -1986,11 +2025,6 @@ CLParser::VerifyOptions()
     Usage("Cannot specify --async and --sync simultaneously.");
   }
 
-  if (params_->inference_load_mode == InferenceLoadMode::FixedSchedule &&
-      params_->warmup_request_count > 0) {
-    Usage("Cannot use warmup options with --fixed-schedule");
-  }
-
```
matthewkotila, line 1993: Warmup with fixed schedule now works.

Removed this check.

src/command_line_parser.cc
```diff
@@ -2227,11 +2261,6 @@ CLParser::VerifyOptions()
     }
   }
 
-  if (params_->inference_load_mode == InferenceLoadMode::FixedSchedule &&
-      params_->kind != cb::BackendKind::OPENAI) {
-    Usage("Fixed schedule mode is only supported with OpenAI service kind.");
-  }
-
```
matthewkotila, line 2234: Additionally, fixed schedule mode is no longer coupled to OpenAI mode.

It never really needed to be, it was just over-restricted to stick to past requirements.

src/custom_request_schedule_manager.cc
```diff
@@ -43,6 +43,15 @@ CustomRequestScheduleManager::Create(
...
 
+void
+CustomRequestScheduleManager::Start()
+{
+  if (!warmup_schedule_.empty()) {
+    PerformWarmup();
+  }
+  StartBenchmark();
+}
+
```
matthewkotila, line 54: As mentioned [here](https://github.com/triton-inference-server/perf_analyzer/pull/366/files#r2057082514), CustomRequestScheduleManager can now perform warmup logic.

src/custom_request_schedule_manager.cc
```diff
@@ -52,62 +61,35 @@ CustomRequestScheduleManager::CustomRequestScheduleManager(
           params.batch_size, params.measurement_window_ms, params.max_trials,
           params.max_threads, params.num_of_sequences,
           params.shared_memory_type, params.output_shm_size,
-          params.serial_sequences, parser, factory, params.request_parameters)
+          params.serial_sequences, parser, factory, params.request_parameters),
+      warmup_request_count_(params.warmup_request_count),
+      benchmark_request_count_(params.request_count)
```
matthewkotila, line 66: These values are now passed to `CustomRequestScheduleManager` at its creation.

The `warmup_request_count_` value is needed [here](https://github.com/triton-inference-server/perf_analyzer/pull/366/files#diff-208ed87aa8e88c62da2501d92468c47876fd526bd164a71fd4d9d1b4536194dbR119) in order to know how many payloads to consider as part of the warmup phase.

The `benchmark_request_count_` value is loosely needed [here](https://github.com/triton-inference-server/perf_analyzer/pull/366/files#diff-208ed87aa8e88c62da2501d92468c47876fd526bd164a71fd4d9d1b4536194dbR113) for extra validation.

src/custom_request_schedule_manager.cc
```diff
@@ -52,62 +61,35 @@ CustomRequestScheduleManager::CustomRequestScheduleManager(
...
 {
-  max_threads_ = std::min(max_threads_, params.request_count);
```
matthewkotila, line 57: Since there are two phases of the fixed schedule mode (CustomRequestScheduleManager), `max_threads_` needs to be potentially altered before each begin.

This potential altering happens [here](https://github.com/triton-inference-server/perf_analyzer/pull/366/files#diff-208ed87aa8e88c62da2501d92468c47876fd526bd164a71fd4d9d1b4536194dbR162) for warmup and [here](https://github.com/triton-inference-server/perf_analyzer/pull/366/files#diff-208ed87aa8e88c62da2501d92468c47876fd526bd164a71fd4d9d1b4536194dbR209) for the following benchmark

src/custom_request_schedule_manager.cc
```diff
@@ -52,62 +61,35 @@ CustomRequestScheduleManager::CustomRequestScheduleManager(
...
-cb::Error
-CustomRequestScheduleManager::InitCustomSchedule(const size_t request_count)
-{
-  PauseWorkers();
-  ConfigureThreads(request_count);
-  GenerateSchedule();
-  ResumeWorkers();
-
-  return cb::Error::Success;
 }
```
matthewkotila, line 68: This was just moved lower.

src/custom_request_schedule_manager.cc
```diff
@@ -52,62 +61,35 @@ CustomRequestScheduleManager::CustomRequestScheduleManager(
...
 
 void
-CustomRequestScheduleManager::GenerateSchedule()
+CustomRequestScheduleManager::DistributeScheduleToWorkers(
+    const Schedule& schedule)
```
matthewkotila, line 72: Just gave this a more accurate name.

src/custom_request_schedule_manager.cc
```diff
@@ -52,62 +61,35 @@ CustomRequestScheduleManager::CustomRequestScheduleManager(
...
-  return worker_schedules;
-}
-
 void
 CustomRequestScheduleManager::InitManagerFinalize()
 {
-  schedule_ = GetScheduleFromDataset();
+  auto [warmup_schedule, benchmark_schedule]{GetSchedulesFromDataset()};
+  warmup_schedule_ = std::move(warmup_schedule);
+  benchmark_schedule_ = std::move(benchmark_schedule);
```
matthewkotila, line 83: As mentioned, there's now two phases: warmup and benchmark, so there's two schedules that get stored and used during the benchmark session.

src/custom_request_schedule_manager.cc
```diff
@@ -118,14 +100,35 @@ CustomRequestScheduleManager::GetScheduleFromDataset() const
...
+  if (warmup_request_count_ >= dataset_size) {
+    throw std::runtime_error(
+        "Expected warmup request count to be less than the dataset size.");
+  }
+
+  if (dataset_size != warmup_request_count_ + benchmark_request_count_) {
+    throw std::runtime_error(
+        "Expected input data JSON size to be equal to the warmup request count "
+        "plus the benchmark request count.");
+  }
```
matthewkotila, line 117: Just extra validation.

Some of this was already checked in command_line_parser.cc, but it's better to re-check it here in case developers change expectations before this code in future.

src/custom_request_schedule_manager.cc
```diff
@@ -118,14 +100,35 @@ CustomRequestScheduleManager::GetScheduleFromDataset() const
...
-    schedule.push_back(timestamp);
+    warmup_schedule.push_back(timestamp);
   }
 
-  std::sort(schedule.begin(), schedule.end());
+  for (size_t dataset_index{warmup_request_count_};
+       dataset_index < dataset_size; ++dataset_index) {
+    const auto timestamp{GetTimestamp(dataset_index)};
+    benchmark_schedule.push_back(timestamp);
+  }
```
matthewkotila, line 129: This is the main magic of the implementation.

Two schedules get created instead of one.

First one is the warmup schedule up to the `warmup_request_count_`-th payload.

Second one is benchmark schedule starting at `warmup_request_count_+1`-th payload up to the last.

src/custom_request_schedule_manager.cc
```diff
@@ -147,4 +150,67 @@ CustomRequestScheduleManager::GetTimestamp(size_t dataset_index) const
...
+
+  const size_t old_max_threads_{max_threads_};
+
+  max_threads_ = std::min(max_threads_, warmup_schedule_.size());
+
+  InitCustomSchedule(warmup_schedule_);
+  WaitForWarmupAndCleanup();
+
+  max_threads_ = old_max_threads_;
```
matthewkotila, line 167: This bit about temporarily changing `max_threads_` was not ideal but the middle-ground implementation that didn't explode a redesign.

In rare cases where there's less than 4 payloads in the warmup, we need to "truncate" `max_threads_` down to that otherwise an infinite loop happens if a thread has no requests to send.

We restore `max_threads_` to its original value in case the user specified something high and the benchmark schedule/dataset is large.

src/custom_request_schedule_manager.cc
```diff
@@ -147,4 +150,67 @@ CustomRequestScheduleManager::GetTimestamp(size_t dataset_index) const
...
+void
+CustomRequestScheduleManager::InitCustomSchedule(
+    const Schedule& schedule, size_t dataset_offset)
+{
+  PauseWorkers();
+  const size_t request_count{schedule.size()};
+  ConfigureThreads(request_count, dataset_offset);
+  DistributeScheduleToWorkers(schedule);
+  ResumeWorkers();
+}
```
matthewkotila, line 179: This was just moved lower.

src/custom_request_schedule_manager.cc
```diff
@@ -147,4 +150,67 @@ CustomRequestScheduleManager::GetTimestamp(size_t dataset_index) const
...
+    const auto timestamp_ns{
+        std::chrono::duration_cast<std::chrono::nanoseconds>(timestamp)};
+    worker_index = thread_ids[thread_id_index];
+    thread_id_index = ++thread_id_index % thread_ids.size();
+    worker_schedules[worker_index]->intervals.push_back(timestamp_ns);
+  }
+  SetScheduleDurations(worker_schedules);
+
+  return worker_schedules;
+}
```
matthewkotila, line 200: Not new--just moved lower.

src/custom_request_schedule_manager.cc
```diff
@@ -147,4 +150,67 @@ CustomRequestScheduleManager::GetTimestamp(size_t dataset_index) const
...
+{
+  if (benchmark_schedule_.empty()) {
+    throw std::runtime_error("Expected benchmark schedule to be non-empty.");
+  }
+
+  max_threads_ = std::min(max_threads_, benchmark_schedule_.size());
+
+  const size_t dataset_offset{warmup_schedule_.size()};
+
+  InitCustomSchedule(benchmark_schedule_, dataset_offset);
```
matthewkotila, line 213: This is also part of the new logic for fixed schedule warmup implementation.

Now, after warmup, the workers take a "dataset offset" value to know to skip that many payloads when starting benchmarking.

They skip however many payloads were sent during warmup.

src/infer_context.cc
```diff
@@ -103,6 +103,12 @@ InferContext::CompleteOngoingSequence(uint32_t seq_stat_index)
   }
 }
 
+void
+InferContext::ApplyDatasetOffset(size_t dataset_offset)
+{
+  data_step_id_ += dataset_offset;
+}
+
```
matthewkotila, line 111: Ultimately, the InferContext inside of the RequestRateWorker needs to know what the dataset offset from [here](https://github.com/triton-inference-server/perf_analyzer/pull/366/files#r2058824073) is in order to skip it during benchmarking.

src/inference_profiler.cc
```diff
@@ -618,12 +618,12 @@ InferenceProfiler::Profile(
...
   if (custom_manager) {
-    RETURN_IF_ERROR(custom_manager->InitCustomSchedule(request_count));
+    custom_manager->Start();
   } else {
     auto* request_rate_manager =
         dynamic_cast<RequestRateManager*>(manager_.get());
-    RETURN_IF_ERROR(
-        request_rate_manager->PerformWarmup(request_rate, request_count));
+    RETURN_IF_ERROR(request_rate_manager->PerformWarmup(
+        request_rate, warmup_request_count));
```
matthewkotila, line 626: This was actually an unrelated bug.

Previously, request rate mode was accidentally using the request count instead of the warmup request count for the number of warmup requests to send.

src/mock_request_rate_worker.h
```diff
@@ -46,12 +46,13 @@ class NaggyMockRequestRateWorker : public RequestRateWorker {
...
       const std::shared_ptr<IInferDataManager>& infer_data_manager,
-      std::shared_ptr<SequenceManager> sequence_manager)
+      std::shared_ptr<SequenceManager> sequence_manager, size_t dataset_offset)
       : RequestRateWorker(
             id, thread_stat, thread_config, parser, data_loader, factory,
             on_sequence_model, async, max_threads, using_json_data, streaming,
             batch_size, wake_signal, wake_mutex, execute, start_time,
-            serial_sequences, infer_data_manager, sequence_manager)
+            serial_sequences, infer_data_manager, sequence_manager,
+            dataset_offset)
```
matthewkotila, line 55: As mentioned [here](https://github.com/triton-inference-server/perf_analyzer/pull/366/files#r2058826068), we need to pass the dataset offset down from CustomRequestScheduleManager down to InferContext so they can pick up in the payload dataset where the warmup left off.

src/model_parser.cc
```diff
@@ -31,6 +31,36 @@
...
+    delay_input->second.shape_.push_back(1);
+    delay_input->second.is_optional_ = true;
+
+    auto session_id_input = inputs_->emplace("session_id", ModelTensor()).first;
+    session_id_input->second.name_ = "session_id";
+    session_id_input->second.datatype_ = "BYTES";
+    session_id_input->second.shape_.push_back(1);
+  }
+}
+
```
matthewkotila, line 63: Previously, these "auxiliary" inference-load-mode-related inputs would be added only in the OpenAI case [here](https://github.com/triton-inference-server/perf_analyzer/pull/366/files#diff-0003e7830a3a4e30facdc6c8b9f2c840ff30a9e11a2728d1547cec946039a8eaL322-L343).

But they don't really need to be limited just to OpenAI backend case.

So now they're added in the constructor.

This allows, for example, fixed schedule warmup to work with tensorrtllm_engine case.

src/model_parser.cc
```diff
@@ -299,7 +329,7 @@ ModelParser::InitDynamicGrpc(
 cb::Error
 ModelParser::InitOpenAI(
     const std::string& model_name, const std::string& model_version,
-    const int32_t batch_size, InferenceLoadMode inference_load_mode)
+    const int32_t batch_size)
```
matthewkotila, line 332: As mentioned [here](https://github.com/triton-inference-server/perf_analyzer/pull/366/files#r2058832731), using inference load mode to add extra "auxiliary" inputs is now done in the constructor to enable more backend kinds to support those inference load modes. Theoretically this is only relevant for fixed schedule. But not really any harm in moving SessionConcurrency stuff either, it already checks earlier that OpenAI mode is being used.

src/model_parser.h
```diff
@@ -68,15 +68,8 @@ class ModelParser {
...
-      : inputs_(std::make_shared<ModelTensorMap>()),
-        outputs_(std::make_shared<ModelTensorMap>()),
-        composing_models_map_(std::make_shared<ComposingModelMap>()),
-        scheduler_type_(NONE), max_batch_size_(0), is_decoupled_(false),
-        response_cache_enabled_(false),
-        top_level_response_caching_enabled_(false)
-  {
-  }
+  explicit ModelParser(
+      InferenceLoadMode inference_load_mode = InferenceLoadMode::None);
```
matthewkotila, line 72: Same comment as [here](https://github.com/triton-inference-server/perf_analyzer/pull/366/files#r2058835804). ModelParser constructor now handles creating "timestamp" input for fixed schedule mode.

src/request_rate_manager.cc
```diff
@@ -240,7 +240,8 @@ RequestRateManager::PauseWorkers()
 }
 
 void
-RequestRateManager::ConfigureThreads(const size_t request_count)
+RequestRateManager::ConfigureThreads(
+    const size_t request_count, size_t dataset_offset)
```
matthewkotila, line 244: Still related to [this](https://github.com/triton-inference-server/perf_analyzer/pull/366/files#r2058826068).

CustomRequestScheduleManager needs to pass the dataset offset down through RequestRateManager through RequestRateWorker to InferContext where it's used to start the benchmark after the payloads that were already sent during warmup.

src/test_command_line_parser.cc
```diff
@@ -1657,40 +1657,6 @@ TEST_CASE("Testing Command Line Parser")
...
-          "--measurement-request-count",
-          "50"};
-
-      REQUIRE_NOTHROW(act = parser.Parse(argc, argv));
-      CHECK(!parser.UsageCalled());
-
-      exp->request_count = 0;
-      exp->measurement_mode = MeasurementMode::COUNT_WINDOWS;
-      exp->measurement_request_count = 50;
-    }
```
matthewkotila, line 1693: This was testing behavior around user passing `--request-rate=0` but really that shouldn't have been allowed.

It errors now, as the command_line_parser [message](https://github.com/triton-inference-server/perf_analyzer/pull/366/files#r2057071264) said all along.

src/test_model_parser.cc
```diff
@@ -180,8 +180,7 @@ TEST_CASE("ModelParser: test Init* functions")
 
   SUBCASE("InitOpenAI")
   {
-    status = mmp.InitOpenAI(
-        model_name, model_version, batch_size, InferenceLoadMode::Concurrency);
+    status = mmp.InitOpenAI(model_name, model_version, batch_size);
```
matthewkotila, line 183: As mentioned [here](https://github.com/triton-inference-server/perf_analyzer/pull/366/files#r2058832731), `ModelParser::InitOpenAI` no longer takes the inference load mode, instead inference-load-mode-related input logic is handled in the `ModelParser` constructor.

But regardless, for some reason, this test didn't actually do anything differently based on the inference load mode.

src/custom_request_schedule_manager.h
```diff
@@ -80,8 +79,18 @@ class CustomRequestScheduleManager : public RequestRateManager {
...
+  std::pair<Schedule, Schedule> GetSchedulesFromDataset() const;
+
+  std::chrono::milliseconds GetTimestamp(size_t dataset_index) const;
+
+  void PerformWarmup();
+
+  void InitCustomSchedule(const Schedule& schedule, size_t dataset_offset = 0);
 
   /// Creates worker schedules based on the provided schedule
   /// \param duration The maximum duration for the schedule
```
nicolasnoble, line 96: I know you're not touching this, but... we should remove this line as it's not actually true.

src/custom_request_schedule_manager.cc
nicolasnoble: I mean, the check is a bit redundant with the one just above? The one on line 108 is checking if `dataset_size` is bigger or equal to `warmup_request_count_`, but then on 113, you ensure it's strictly equal. Not really a big deal, just maybe a tiny bit weird.

src/test_command_line_parser.cc
```diff
@@ -1657,40 +1657,6 @@ TEST_CASE("Testing Command Line Parser")
       exp->measurement_mode = MeasurementMode::COUNT_WINDOWS;
       exp->measurement_request_count = 2000;
     }
-    SUBCASE("zero value (no override to measurement mode)")
```
nicolasnoble, line 1660: Why was this one removed? I don't see it immediately.

nicolasnoble (dismissed):
Yeah looks good to me. The change is large due to propagation of some of the parameters, but that's mostly mechanical work. I've left only a few nits here and there, otherwise it's fine with me.

src/custom_request_schedule_manager.cc
matthewkotila: Yeah, you're spot on.

Removed this check and the `benchmark_request_count_` data member.

src/custom_request_schedule_manager.h
matthewkotila: Removed 👍

nicolasnoble (approved):

PR #371, opened 2025-05-02
fix: forward --header arguments to Perf Analyzer CLI

ajcasagrande:
Fixes issue where --header args to GAP are not forwarded to PA.


nv-braf (approved):

nv-braf (approved):

debermudez (approved):

xihajun:
thanks for updating! would it possible to update the version here as well? https://pypi.org/project/genai-perf/0.0.13/#description

PR #396, opened 2025-05-24
Get rid of --service-kind in genai-perf README

zhaohaidao:
When executing the genai-perf test with reference to the README, it will report an error that service-kind is an unrecognized argument. The error details is as follows:

> usage: genai-perf [-h] [--version] {config,profile,analyze,create-template,process-export-files} ...
genai-perf: error: unrecognized arguments: --service-kind openai

After analysis, it is known that the service-kind argument has been removed through (https://github.com/triton-inference-server/perf_analyzer/pull/355)


debermudez:
Good catch. I kicked off the workflows as a sanity check. 
The mirror wont work but the rest should. 
The rest of the testing already accounts for this change and the documentation was missed in this one place.

debermudez (approved):

PR #398, opened 2025-06-03
Add wheel support for Perf Analyzer

matthewkotila:
- In an effort to decouple the GenAI-Perf and Perf Analyzer release process from the Triton release process, we propose releasing Perf Analyzer as its own pypi package
- Previously, Perf Analyzer was included in the Triton `tritonclient` pypi package, definitionally coupling Triton with GAP/PA
- This PR adds some necessary support code for allowing a python wheel to be created for Perf Analyzer
- The wheel simply includes the `perf_analyzer` binary and a single wrapper python script that calls it and forwards all arguments

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,8 @@ artifacts/
 
 # Avoid squashing local people's file for now
 /.vscode
+
+# Python setuptools artifacts
+*.egg-info/
+dist/
+bin/
--- a/src/perf_analyzer/cli.py
+++ b/src/perf_analyzer/cli.py
@@ -1,4 +1,4 @@
-# Copyright 2023-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -6,3 +6,8 @@ artifacts/
 
 # Avoid squashing local people's file for now
 /.vscode
+
+# Python setuptools artifacts
+*.egg-info/
+dist/
+bin/
```
matthewkotila, line 13: This is just stuff that would get created when making the PA wheel. Don't want it accidentally getting added when doing wheel dev.

pyproject.toml
```diff
@@ -24,6 +24,28 @@
...
+
+[tool.setuptools.packages.find]
+where = ["src"]
+
+[tool.setuptools.package-data]
+"perf_analyzer" = ["bin/perf_analyzer"]
+
+[project.scripts]
+perf_analyzer = "perf_analyzer.cli:main"
+
```
matthewkotila, line 48: This is just necessary stuff for making a PA wheel that is effectively a wrapper for the pre-existing binary.

src/perf_analyzer/cli.py
```diff
@@ -0,0 +1,34 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import subprocess
+import sys
+from pathlib import Path
+
+
+def main():
+    bin_path = Path(__file__).parent / "bin/perf_analyzer"
+    subprocess.run([bin_path] + sys.argv[1:])
```
matthewkotila, line 34: As mentioned, the new `perf-analyzer` pip package needs a wrapper python script to forward to the binary also packaged with the wheel.

lkomali:
LGTM

I think we should update the [perf_analyzer/docs/install.md](https://github.com/triton-inference-server/perf_analyzer/blob/main/docs/install.md#pip) ?

.gitignore
matthewkotila, line 1: > @lkomali: LGTM
> 
> I think we should update the [perf_analyzer/docs/install.md](https://github.com/triton-inference-server/perf_analyzer/blob/main/docs/install.md#pip) ?

I know this sounds a little silly, but considering this product is being shifted into a lower level of support going forward, I'd prefer to do this later or never. Let's just get this PR merged for now if that's ok.

PR #417, opened 2025-07-09
build: TPRD-1622 - Update version to CMake 4.0 or 3.31 ( 3.31.8 ) which has support for CUDA new layout

mc-nv:
Update version to CMake 4.0 or 3.31 ( 3.31.8 ) which has support for CUDA new layout


src/client_backend/openai/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-cmake_minimum_required (VERSION 3.18)
+cmake_minimum_required (VERSION 3.31.8)
```
yinggeh, line 27: Update copyrights

yinggeh (approved):

matthewkotila (commented):
Ignore

CMakeLists.txt
matthewkotila, line 1: I've tried to update PA CMake to 4.0 before and ran into issues.

I think some dependencies set their minimum cmake version too low for cmake 4 to accept.

Did you see such issues? Is there a passing pipeline for this change?

mc-nv: You can pass environment variable to `CMAKE_POLICY_VERSION_MINIMUM=3.5` to bypass the issue. 
This is a shortest way to achieve it : https://cmake.org/cmake/help/latest/variable/CMAKE_POLICY_VERSION_MINIMUM.html

matthewkotila (approved):

PR #421, opened 2025-07-24
Fix: Resolve plots directory path inconsistency in GenAI-perf

Rythsman:
When using model names with special characters (e.g., paths with slashes), GenAI-perf creates sanitized artifact directory names but the plots generation fails due to path mismatches between profile() and create_plots() methods.

- Fixed create_plots() to use the same PerfAnalyzerConfig path logic as profile()
- Ensured plots directory is created under the correct sanitized artifact directory
- Added proper path resolution for profile_export_file in plots configuration

This resolves the "No such file or directory: 'artifacts/plots/config.yaml'" error fundamentally.

### DIFF
--- a/genai-perf/genai_perf/subcommand/profile.py
+++ b/genai-perf/genai_perf/subcommand/profile.py
@@ -58,44 +58,41 @@ class Profile(Subcommand):
 
     def __init__(self, config: ConfigCommand, extra_args: Optional[List[str]]) -> None:
         super().__init__(config, extra_args)
+        self._objectives = self._create_objectives_based_on_stimulus()
+        self._genai_perf_config = self._create_genai_perf_config(self._objectives)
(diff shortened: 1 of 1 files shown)


ExplorerRay:
I also noticed this issue. Current code only needs to create plot config one time, but your code will cause configs to be generated several times. I am not sure whether this is good.

Rythsman:
> I also noticed this issue. Current code only needs to create plot config one time, but your code will cause configs to be generated several times. I am not sure whether this is good.

i see, i have modified to make create plot config only one time

Rythsman:
> I also noticed this issue. Current code only needs to create plot config one time, but your code will cause configs to be generated several times. I am not sure whether this is good.

@ExplorerRay i see, i have modified to make create plot config only one time

PR #427, opened 2025-07-31
fix: set null text output to empty string

nv-hwoo:
Set null text content/outputs to empty strings to prevent errors when working with response outputs.


matthewkotila (approved):
🚀

PR #440, opened 2025-08-28
fix: gracefully handle erroneous SSE responses

nv-hwoo:
This PR detects any server side error events or parsing errors in SSE responses and drops them without exiting prematurely.

### DIFF
--- a/genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
+++ b/genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
@@ -384,7 +384,7 @@ class LLMProfileDataParser(ProfileDataParser):
                     elif self._is_empty_response(responses[0]):
                         res_outputs[i]["response"] = ""
                 except Exception as e:
-                    logger.debug(f"Error parsing a response: {str(e)}")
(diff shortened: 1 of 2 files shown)


genai-perf/genai_perf/profile_data_parser/llm_profile_data_parser.py
```diff
@@ -330,37 +339,54 @@ def _preprocess_response(
...
+                                data["choices"][0]["text"] = merged_text
+                            else:
+                                data["choices"][0]["delta"]["content"] = merged_text
+                        res_outputs[i] = {
+                            "response": orjson.dumps(data).decode("utf-8")
+                        }
+                    elif self._is_empty_response(responses[0]):
+                        res_outputs[i]["response"] = ""
+                except Exception as e:
+                    logger.debug(f"Error parsing a response: {str(e)}")
```
dyastremsky, line 387: Can you print any more information about the response here? Should this be logger.error? Also, will the response then be excluded from the metrics?

dyastremsky (dismissed):

dyastremsky (approved):

PR #460, opened 2026-03-07
fix: Fix memory overuse issues

matthewkotila:
a few small changes removed a memory doubling side-effect and converted an O(n*k) time operation to O(n)

### DIFF
--- a/src/inference_profiler.cc
+++ b/src/inference_profiler.cc
@@ -1,4 +1,4 @@
-// Copyright 2020-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2020-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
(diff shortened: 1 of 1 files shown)


matthewkotila:
The check failure is from GAP unit tests, and it is a known issue unrelated to this PR, it also exists in main.

src/inference_profiler.cc
```diff
@@ -1373,18 +1374,18 @@ InferenceProfiler::ValidLatencyMeasurement(
...
+  std::unordered_set<size_t> erase_set(
+      erase_indices.begin(), erase_indices.end());
+  for (size_t i = 0; i < all_request_records_.size(); i++) {
+    if (erase_set.count(i)) {
+      valid_requests.push_back(std::move(all_request_records_[i]));
+    }
+  }
+  auto new_end = std::remove_if(
+      all_request_records_.begin(), all_request_records_.end(),
+      [](const RequestRecord& r) { return r.response_timestamps_.empty(); });
```
debermudez: Potential correctness issue: this remove-if assumes moved-from `RequestRecord` objects always have `response_timestamps_` emptied after `std::move`, but moved-from container state is valid-yet-unspecified in C++. If a moved-from record retains a non-empty vector, it will remain in `all_request_records_` and can be reprocessed in later windows. Could we erase by index/flag directly (O(n)) instead of inferring from moved-from state?

src/inference_profiler.cc
```diff
@@ -1373,18 +1374,18 @@ InferenceProfiler::ValidLatencyMeasurement(
...
+  all_request_records_.erase(new_end, all_request_records_.end());
+  std::cout << "got here" << std::endl;
```
debermudez: Looks like temporary debug output left in the hot path. This will print on every call and can pollute benchmark output/logs; please remove before merge.

debermudez (approved):

src/inference_profiler.cc
matthewkotila: Good catch. I changed it to not rely on post-move state.

Manually re-tested with the bug that this fixes and it's still fixed.

src/inference_profiler.cc
matthewkotila: Good catch. Removed.

